PR #133, opened 2025-01-15
  Add Server and Client classes and UDN test for connectivity preservation after live migration

sbahar619:
##### Short description:
Add Server and Client classes and UDN test for connectivity preservation after live migration
##### More details:
Add UDN test for connectivity post live migration
    
    This commit introduces a test for UDN to ensure connectivity is preserved
    during live migration scenarios.
    
    Introducing server and client fixtures using the newly created
    classes to test TCP connectivity between two VMs. The fixtures
    sets up the connection between 2 VMs using iperf3.
    
    Additionally, a test named test_connectivity_is_preserved_after_live_migration
    is added to verify that TCP connectivity is preserved after live migration.
    This ensures the integrity of network connections during VM migration scenarios.
    
    This automation is based on the test case CNV-11427.

Add classes for iperf3 network testing
    
    The Server and Client classes have been added to
    represent the server and client components for iperf3-based
    network performance testing between two VMs.
    
    These changes provide a structured way to run network tests and
    validate connectivity between VMs over TCP protocol using iperf3.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-52875

### DIFF
--- a/libs/net/traffic_generator.py
+++ b/libs/net/traffic_generator.py
@@ -1,0 +1,104 @@
+import logging
+from typing import Final
+
+from ocp_utilities.exceptions import CommandExecFailed
+
+from libs.vm.vm import BaseVirtualMachine
+
+_DEFAULT_CMD_TIMEOUT_SEC: Final[int] = 10
+_IPERF_BIN: Final[str] = "iperf3"
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Server:
+    """
+    Represents a server running on a virtual machine for testing network performance.
+    Implemented with iperf3
+
+    Args:
+        vm (BaseVirtualMachine): The virtual machine where the server runs.
+        port (int): The port on which the server listens for client connections.
+    """
+
+    def __init__(
+        self,
+        vm: BaseVirtualMachine,
+        port: int,
+    ):
+        self._vm = vm
+        self._port = port
+        self._cmd = f"{_IPERF_BIN} --server --port {self._port} --one-off"
+
+    def __enter__(self) -> "Server":
+        self._vm.console(
+            commands=[f"{self._cmd} &"],
+            timeout=_DEFAULT_CMD_TIMEOUT_SEC,
+        )
+        return self
+
+    def __exit__(self, exc_type: BaseException, exc_value: BaseException, traceback: object) -> None:
+        _stop_process(vm=self._vm, cmd=self._cmd)
+
+    def is_running(self) -> bool:
+        return _is_process_running(vm=self._vm, cmd=self._cmd)
+
+
+class Client:
+    """
+    Represents a client that connects to a server to test network performance.
+    Implemented with iperf3
+
+    Args:
+        vm (BaseVirtualMachine): The virtual machine where the client runs.
+        server_ip (str): The destination IP address of the server the client connects to.
+        server_port (int): The port on which the server listens for connections.
+    """
+
+    def __init__(
+        self,
+        vm: BaseVirtualMachine,
+        server_ip: str,
+        server_port: int,
+    ):
+        self._vm = vm
+        self._server_ip = server_ip
+        self._server_port = server_port
+        self._cmd = f"{_IPERF_BIN} --client {self._server_ip} --time 0 --port {self._server_port}"
+
+    def __enter__(self) -> "Client":
+        self._vm.console(
+            commands=[f"{self._cmd} &"],
+            timeout=_DEFAULT_CMD_TIMEOUT_SEC,
+        )
+        return self
+
+    def __exit__(self, exc_type: BaseException, exc_value: BaseException, traceback: object) -> None:
+        _stop_process(vm=self._vm, cmd=self._cmd)
+
+    def is_running(self) -> bool:
+        return _is_process_running(vm=self._vm, cmd=self._cmd)
+
+
+def _stop_process(vm: BaseVirtualMachine, cmd: str) -> None:
+    try:
+        vm.console(commands=[f"pkill -f '{cmd}'"], timeout=_DEFAULT_CMD_TIMEOUT_SEC)
+    except CommandExecFailed as e:
+        LOGGER.warning(str(e))
+
+
+def _is_process_running(vm: BaseVirtualMachine, cmd: str) -> bool:
+    try:
+        vm.console(
+            commands=[f"pgrep -fAx '{cmd}'"],
+            timeout=_DEFAULT_CMD_TIMEOUT_SEC,
+        )
+        return True
+    except CommandExecFailed as e:
+        LOGGER.info(f"Process is not running on VM {vm.name}. Error: {str(e)}")
+        return False
+
+
+def is_tcp_connection(server: Server, client: Client) -> bool:
+    return server.is_running() and client.is_running()
--- a/tests/network/user_defined_network/test_user_defined_network.py
+++ b/tests/network/user_defined_network/test_user_defined_network.py
@@ -5,6 +5,7 @@ from ocp_resources.resource import Resource
 from ocp_resources.user_defined_network import Layer2UserDefinedNetwork
 from ocp_resources.utils.constants import TIMEOUT_1MINUTE
 
+from libs.net.traffic_generator import Client, Server, is_tcp_connection
 from libs.net.udn import udn_primary_network
 from libs.net.vmspec import lookup_iface_status, lookup_primary_network
 from libs.vm import affinity
@@ -18,6 +19,7 @@ from utilities.virt import migrate_vm_and_verify
 pytestmark = pytest.mark.udn
 
 IP_ADDRESS = "ipAddress"
+SERVER_PORT = 5201
 
 
 def udn_vm(namespace_name, name, template_labels=None):
@@ -69,6 +71,26 @@ def vmb_udn_non_migratable(namespace, namespaced_layer2_user_defined_network, ud
         yield vm
 
 
+@pytest.fixture()
+def server(vmb_udn_non_migratable):
+    with Server(vm=vmb_udn_non_migratable, port=SERVER_PORT) as server:
+        assert server.is_running()
+        yield server
+
+
+@pytest.fixture()
+def client(vma_udn, vmb_udn_non_migratable):
+    with Client(
+        vm=vma_udn,
+        server_ip=lookup_iface_status(
+            vm=vmb_udn_non_migratable, iface_name=lookup_primary_network(vm=vmb_udn_non_migratable).name
+        )[IP_ADDRESS],
+        server_port=SERVER_PORT,
+    ) as client:
+        assert client.is_running()
+        yield client
+
+
 @pytest.mark.ipv4
 class TestPrimaryUdn:
     @pytest.mark.polarion("CNV-11624")
@@ -107,3 +129,8 @@ class TestPrimaryUdn:
             vm=vmb_udn_non_migratable, iface_name=lookup_primary_network(vm=vmb_udn_non_migratable).name
         )[IP_ADDRESS]
         vma_udn.console(commands=[f"ping -c 3 {target_vm_ip}"], timeout=TIMEOUT_1MIN)
+
+    @pytest.mark.polarion("CNV-11427")
+    def test_connectivity_is_preserved_after_live_migration(self, vma_udn, server, client):
+        migrate_vm_and_verify(vm=vma_udn)
+        assert is_tcp_connection(server=server, client=client)


sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/07bfb39129144589e675dd32c482a50b83dae23c..0af99cc69fc1c79962f22bce0466420ef616ca47
Address @EdDev comments.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/de8e270624758af21e690f9c1918c90664cf4b82..2916d9ae7e5f35cf3a5fa1a78aee181944d70f5d
Fix pre commit.

sbahar619:
Change: rebase.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/a47b11990d3f49fb501458bae83b4b0606279016..4f44750860b5bdc0c3c767df3911f15f13ccb4d5
Refactor.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-133 published

sbahar619:
Change: refactor.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+    def vm(self) -> BaseVirtualMachine:
+        return self._vm
+
+    @property
+    def port(self) -> str:
+        return self._port
+
+    def start(self) -> None:
+        self._vm.console(
+            commands=[f"{self._cmd} &", f"pgrep -ofA '{self._cmd}'"],
```
sbahar619, line 79: The reason I used the additional command of pgrep is because the first command uses & to run the iperf3 process in the background, and therefore the exit code of the command will always be 0 regardless of whether the iperf3 command succeeds or fails. By adding the pgrep -ofA '{self._cmd}' command, we can check for the running process and ensure the iperf3 command is actively running. If it fails to start, pgrep will not find the process and return a non-zero exit code, allowing the console method to raise the expected exception. This ensures that we handle failures in the background process correctly.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/f345a4bdfaf661a7bc64f12f34798be395686bb0..43a656418017799cff2393dc671fb789dc2de2ee
refactor

yossisegev (approved):
I added a suggested comment, but it's not something that is required (unless someone thinks otherwise).

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+        self,
+        src_vm: BaseVirtualMachine,
+        dst_vm: BaseVirtualMachine,
+        dst_ip: str,
+        port: str,
+    ):
+        self._server = TrafficGeneratorServer(vm=src_vm, port=port)
+        self._client = TrafficGeneratorClient(vm=dst_vm, dst_ip=dst_ip, port=port)
+
+    def __enter__(self) -> TrafficGenerator:
```
EdDev, line 119: Self referencing the class needs to be done using parenthesis:
`def __enter__(self) -> 'TrafficGenerator':`

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
EdDev: That means you should "fix" it.
I think it is about self-referencing the class name, I commented at that location. I do not think it is worth adding the future annotation for it.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class TrafficGeneratorServer:
+    """
+    Represents a server running on a virtual machine that uses iperf3 for testing network performance.
+
+    Attributes:
```
EdDev, line 22: `Args`

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+_IPERF_BIN: Final[str] = "iperf3"
+_PROTOCOL_TCP = "TCP"
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class TrafficGeneratorServer:
+    """
+    Represents a server running on a virtual machine that uses iperf3 for testing network performance.
```
EdDev, line 20: iperf3 is an implementation detail, we are not interested in focusing on the tool used.
As I mentioned in the past, we want the freedom to change the implementation if we choose, so such details are not useful for the user (or we want it to be like that).

Said that, I think it is fine to add small details like this at the end. Something like `Implemented with iperf3` in a new sentence.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+    ):
+        self._vm = vm
+        self._port = port
+        self._cmd = f"{_IPERF_BIN} -s -D -p {self._port} -1"
+
+    @property
+    def vm(self) -> BaseVirtualMachine:
+        return self._vm
+
+    def start(self) -> None:
```
EdDev, line 40: Where is the `stop`?

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+            timeout=_DEFAULT_CMD_TIMEOUT_SEC,
+        )
+        LOGGER.info(f"Server started successfully on VM {self._vm.name}.")
+
+
+class TrafficGeneratorClient:
+    """
+    Represents a client that connects to a server using iperf3 to test network performance.
+
+    Attributes:
```
EdDev, line 52: `Args`

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+        self._vm.console(
+            commands=[self._cmd],
+            timeout=_DEFAULT_CMD_TIMEOUT_SEC,
+        )
+        LOGGER.info(f"Server started successfully on VM {self._vm.name}.")
+
+
+class TrafficGeneratorClient:
+    """
+    Represents a client that connects to a server using iperf3 to test network performance.
```
EdDev, line 50: Same

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+
+_DEFAULT_CMD_TIMEOUT_SEC: Final[int] = 10
+_IPERF_BIN: Final[str] = "iperf3"
+_PROTOCOL_TCP = "TCP"
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class TrafficGeneratorServer:
```
EdDev, line 18: As the module name is identical to the prefix here, it can be dropped. The context exists.
Same for others.

libs/net/traffic_generator.py
EdDev: This is raceful so we do not want to have something like this at all.
You call something in an async manner and you only know it is still running at the time you check it.
It may still not have been executed by bash, so you will not see it using `pgrep`, or you will not catch it failed because it happened 1ns after the check.

Therefore, you should add a method for the client and server to check that each is running (e.g. `is_running()`). The one who uses these will check the needed when relevant.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+    For more details about iperf3, its features, and configuration options, refer to the
+    official iperf3 documentation: https://iperf.fr/iperf-doc.php
+    """
+
+    def __init__(
+        self,
+        src_vm: BaseVirtualMachine,
+        dst_vm: BaseVirtualMachine,
+        dst_ip: str,
+        port: str,
```
EdDev, line 114: The decision of splitting into 3 classes implies that this should just get two inputs: The server and client.
If it get these inputs, it is hard for you to reason for the need of the server and client as independent objects. At least it is weak.

It is not a must, but it will be a better structure for sure (e.g. future different clients/servers could be combined).
If you want to encapsulate data and "hide" them from the testers, then you can use a very simple factory that will return a combination of the 3 (one-liner probably).

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+
+    def __init__(
+        self,
+        vm: BaseVirtualMachine,
+        port: str,
+    ):
+        self._vm = vm
+        self._port = port
+        self._cmd = f"{_IPERF_BIN} -s -D -p {self._port} -1"
```
EdDev, line 34: Please use the long-names, not shortcuts. I prefer not to go to the man if possible.
It also reads nicer.

Same for the client.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+        dst_vm: BaseVirtualMachine,
+        dst_ip: str,
+        port: str,
+    ):
+        self._server = TrafficGeneratorServer(vm=src_vm, port=port)
+        self._client = TrafficGeneratorClient(vm=dst_vm, dst_ip=dst_ip, port=port)
+
+    def __enter__(self) -> TrafficGenerator:
+        self._server.start()
+        if not self.is_listening():
```
EdDev, line 121: This seems unsafe: If this line raises or it proceeds to the body to raise, then the server will not be stopped.

Same for all the steps beyond here.

Failure to enter a context is not calling the exit.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,171 @@
...
+        return _is_socket(session=self, state=State.LISTENING)
+
+    def is_established(self) -> bool:
+        return _is_socket(session=self, state=State.ESTABLISHED)
+
+    def is_active(self) -> bool:
+        return self.is_listening() and self.is_established()
+
+
+def _is_socket(
```
EdDev, line 154: Please convince us why this complication is needed. I do not see an explanation on the commit and cannot figure out myself.

At the end, you are using iperf3 as a tool to make all this work. So you either trust it to do what you intend for it to do or you do not. If there is a need to dig so deep into the kernel sockets, we loose the benefit of using the tool.

But maybe I am missing something that makes this a hard requirement.

EdDev (changes requested):
Thank you, nice work.
I think it can be further simplified, so I asked for some clarifications in the review regarding some aspects of the implementation.

libs/net/traffic_generator.py
sbahar619: I implemented this function because it's part of the test case requirements. However, I agree that it's not strictly necessary, as we can trust the tool to create the TCP connections, and simply verifying the process is running would enough.

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: I explained it here:
https://github.com/RedHatQE/openshift-virtualization-tests/blob/78f80e21030ee13074c51651ead6281e0bfb109f/libs/net/traffic_generator.py#L128
Because we use --one-off flag, the server stops after the client finishes, so we don't need to stop the server, and therefore it was omitted.
If I will add the stop method, it will not be used.

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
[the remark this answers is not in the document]
sbahar619: ack

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/43a656418017799cff2393dc671fb789dc2de2ee..755832bc5bad4808980a47d32f3279028df2b934
Resolve @EdDev comments.

sbahar619:
Change: update commit messages.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-133 published

sbahar619:
Change: rebase.

tests/network/user_defined_network/test_user_defined_network.py
```diff
@@ -69,6 +70,19 @@ def vmb_udn_non_migratable(namespace, namespaced_layer2_user_defined_network, ud
         yield vm
 
 
+@pytest.fixture
+def tcp_connection(vma_udn, vmb_udn_non_migratable):
+    with connection_factory(
+        src_vm=vmb_udn_non_migratable,
+        dst_vm=vma_udn,
+        dst_ip=lookup_iface_status(
```
yossisegev, line 78: vmb is the source (`src_vm=vmb_udn_non_migratable`), but you search for the destination IP in it.

yossisegev (changes requested):

tests/network/user_defined_network/test_user_defined_network.py
sbahar619: Right, let me fix the roles here.

sbahar619: Done.
https://github.com/RedHatQE/openshift-virtualization-tests/compare/42d02cf3f572ce54feca8764cb77f983abe1342f..46ff2d07edd4b54419dc938c88ea9b601b93cdef

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/42d02cf3f572ce54feca8764cb77f983abe1342f..46ff2d07edd4b54419dc938c88ea9b601b93cdef
Address @yossisegev comment.

tests/network/user_defined_network/test_user_defined_network.py
yossisegev: Thank you

yossisegev (approved):

EdDev (changes requested):

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/46ff2d07edd4b54419dc938c88ea9b601b93cdef..8e2215385813f722b505c39a57caa27380cc650d
Resolve @EdDev comments.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-133 published

libs/net/traffic_generator.py
sbahar619: Thanks, good point, done.

tests/network/user_defined_network/test_user_defined_network.py
```diff
@@ -107,3 +121,8 @@ def test_basic_connectivity_between_udn_vms(self, vma_udn, vmb_udn_non_migratabl
             vm=vmb_udn_non_migratable, iface_name=lookup_primary_network(vm=vmb_udn_non_migratable).name
         )[IP_ADDRESS]
         vma_udn.console(commands=[f"ping -c 3 {target_vm_ip}"], timeout=TIMEOUT_1MIN)
+
+    @pytest.mark.polarion("CNV-11427")
+    def test_connectivity_is_preserved_after_live_migration(self, vma_udn, tcp_connection):
+        migrate_vm_and_verify(vm=vma_udn)
```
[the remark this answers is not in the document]
sbahar619: The connectivity is being validated as part of the setup:
https://github.com/RedHatQE/openshift-virtualization-tests/blob/e996e362db9bfd36da52cf0be7f3e63d52021216/libs/net/traffic_generator.py#L125

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,152 @@
...
+        vm: BaseVirtualMachine,
+        port: str,
+    ):
+        self._vm = vm
+        self._port = port
+        self._cmd = f"{_IPERF_BIN} --server --daemon --port {self._port} --one-off"
+
+    @property
+    def vm(self) -> BaseVirtualMachine:
+        return self._vm
```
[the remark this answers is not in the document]
sbahar619: Thanks, done.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,152 @@
...
+    def is_running(self) -> bool:
+        return _is_process_running(vm=self._vm, cmd=self._cmd)
+
+
+class Connection:
+    """
+    Orchestrates network testing between a server and a client.
+
+    This class sets up a server on the source virtual machine and a client on the
+    destination virtual machine to perform network performance testing. It manages the
```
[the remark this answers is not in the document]
sbahar619: ack, done.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,152 @@
...
+            commands=[f"pgrep -ofA '{cmd}'"],
+            timeout=_DEFAULT_CMD_TIMEOUT_SEC,
+        )
+        return True
+    except CommandExecFailed as e:
+        LOGGER.error(f"Process is not running on VM {vm.name}. Error: {str(e)}")
+        return False
+
+
+def connection_factory(dst_vm: BaseVirtualMachine, src_vm: BaseVirtualMachine, dst_ip: str, port: str) -> Connection:
```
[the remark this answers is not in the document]
sbahar619: Thanks, done.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,152 @@
...
+        self._port = port
+        self._cmd = f"{_IPERF_BIN} --client {self._dst_ip} --time 0 --port {self._port}"
+
+    @property
+    def vm(self) -> BaseVirtualMachine:
+        return self._vm
+
+    @property
+    def port(self) -> str:
+        return self._port
```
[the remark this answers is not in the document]
sbahar619: Thanks, done.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,152 @@
...
+        # Server stops after the client finishes, as only one connection (--one-off) is allowed.
+        self._client.stop()
+
+    @property
+    def server(self) -> Server:
+        return self._server
+
+    @property
+    def client(self) -> Client:
+        return self._client
```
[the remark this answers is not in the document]
sbahar619: Thanks, done.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    Orchestrates network testing between a server and a client.
+
+    This class sets up a server on the source virtual machine and a client on the
+    destination virtual machine to perform network performance testing. It manages the
+    lifecycle of the server and client, ensuring proper startup and cleanup.
+
+    Args:
+        server (Server): The server instance to run on the source virtual machine.
+        client (Client): The client instance to run on the destination virtual machine.
```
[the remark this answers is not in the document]
sbahar619: ack, done.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,152 @@
...
+
+class Client:
+    """
+    Represents a client that connects to a server to test network performance.
+    Implemented with iperf3
+
+    Args:
+        vm (BaseVirtualMachine): The virtual machine where the client runs.
+        dst_ip (str): The destination IP address of the server the client connects to.
+        port (str): The port on which the server listens for connections.
```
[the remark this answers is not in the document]
sbahar619: Thanks, done.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class Client:
+    """
+    Represents a client that connects to a server to test network performance.
+    Implemented with iperf3
+
+    Args:
+        vm (BaseVirtualMachine): The virtual machine where the client runs.
+        dst_ip (str): The destination IP address of the server the client connects to.
```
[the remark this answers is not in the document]
sbahar619: Thanks, done.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/8e2215385813f722b505c39a57caa27380cc650d..21c93883a445d1b91cbd928c4ff823bf67efb8c3
Refactor.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,163 @@
...
+        self._vm.console(
+            commands=[f"{self._cmd} &"],
+            timeout=_DEFAULT_CMD_TIMEOUT_SEC,
+        )
+
+    def stop(self) -> None:
+        self._vm.console(commands=[f"pkill {_IPERF_BIN}"], timeout=_DEFAULT_CMD_TIMEOUT_SEC)
+
+    def is_running(self) -> bool:
+        return _is_process_running(vm=self._vm, cmd=self._cmd)
```
Anatw, line 93: Feels like we should have a base class that both server and client will inherit from.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class ConnectionSetupError(Exception):
+    """Custom exception raised when connection setup fails."""
+
```
Anatw, line 98: You are missing "pass" (or it should raise SyntaxError).

libs/net/traffic_generator.py
sbahar619: ack

libs/net/traffic_generator.py
sbahar619: It's valid in python, no syntax error.

libs/net/traffic_generator.py
sbahar619: ack

sbahar619: ack, let me check it out.

sbahar619: @EdDev 
@Anatw 
I already implemented it as dev commit, please review, if it's a good fit I will add it:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/133/commits/da55efbf25e278b690d91645e589291ee77bb198

tests/network/user_defined_network/test_user_defined_network.py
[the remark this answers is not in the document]
EdDev: Ok, makes sense.

tests/network/user_defined_network/test_user_defined_network.py
```diff
@@ -69,6 +70,19 @@ def vmb_udn_non_migratable(namespace, namespaced_layer2_user_defined_network, ud
         yield vm
 
 
+@pytest.fixture
+def tcp_connection(vma_udn, vmb_udn_non_migratable):
```
EdDev, line 74: This feels like a border line between being part of the setup and the test itself.
Personally, I prefer it as part of the test, so it will be clear which are the two VMs involved (without looking outside the test).
It will be also nice if the connection failure between the two VMs is also a test failure and not a setup error.
I do acknowledge it could be a setup, so this is just to raise awareness and think about it.

libs/net/traffic_generator.py
EdDev: I do not think `TrafficGenerator` represents a traffic generator.
I feel it complicates things, but feel free to propose a good name that represents what it actually does.
Then I will check is a "server" is a "the new name".

sbahar619: ack

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,163 @@
...
+        if not self.is_active():
+            raise ConnectionSetupError("Connection setup failed")
+
+        return self
+
+    def __exit__(self, exc_type: BaseException, exc_value: BaseException, traceback: object) -> None:
+        if self._client.is_running():
+            self._client.stop()
+        if self._server.is_running():
+            self._server.stop()
```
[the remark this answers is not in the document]
sbahar619: That means I need to handle the CommandExecFailed exception at stop() method, just to realize there is no pid to kill.
We already have a method for checking if pid exists or not, so why not use it?

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,146 @@
...
+    ):
+        self._server = server
+        self._client = client
+
+    def __enter__(self) -> "Connection":
+        self._server.start()
+        self._client.start()
+
+        if not self.is_active():
+            raise ConnectionSetupError("Connection setup failed")
```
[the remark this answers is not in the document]
sbahar619: Why not?
Any exception during the __enter__ will cause an execution of __exit__, which means cleanup will be handled.
If server or client start raise an exception, the cleanup will be handled.
If it somehow got successful execution but terminated shortly after we raise the  ConnectionSetupError and again cleanup is handled.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,163 @@
...
+        name (str): The name of the server instance, used for identification.
+        vm (BaseVirtualMachine): The virtual machine where the server runs.
+        port (str): The port on which the server listens for client connections.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        vm: BaseVirtualMachine,
+        port: str,
```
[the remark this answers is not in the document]
sbahar619: ack

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,163 @@
...
+        server_ip (str): The destination IP address of the server the client connects to.
+        server_port (str): The port on which the server listens for connections.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        vm: BaseVirtualMachine,
+        server_ip: str,
+        server_port: str,
```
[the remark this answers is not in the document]
sbahar619: ack

tests/network/user_defined_network/test_user_defined_network.py
sbahar619: done.

libs/net/traffic_generator.py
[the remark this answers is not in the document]
sbahar619: done.

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,163 @@
...
+        port (str): The port on which the server listens for client connections.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        vm: BaseVirtualMachine,
+        port: str,
+    ):
+        self._name = name
```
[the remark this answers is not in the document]
sbahar619: done.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-133 published

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-133 published

libs/net/traffic_generator.py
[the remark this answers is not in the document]
EdDev: > Any exception during the enter will cause an execution of exit

Wrong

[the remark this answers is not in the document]
sbahar619: Right, I will fix it soon.

libs/net/traffic_generator.py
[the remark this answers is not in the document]
EdDev: It complies with the [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) style.
In this case, there is no need to even raise something, although it could be useful for troubleshooting (e.g. logging).

As a prinicpal rule, when you ask something and then act on it, you are in a classic race: Between the time you asked the question and got an answer and the time you acted, the actual state may change.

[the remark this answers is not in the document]
sbahar619: ack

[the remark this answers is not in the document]
sbahar619: Done.

libs/net/traffic_generator.py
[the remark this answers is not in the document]
sbahar619: Done.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/4647b47a766e748dd1eabe7f37f3da8d6a80a2a6..1d7e0d677e564710ce66f0727bf18d02e4cca206
Resolve @EdDev comments.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-133 published

EdDev:
> Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/4647b47a766e748dd1eabe7f37f3da8d6a80a2a6..1d7e0d677e564710ce66f0727bf18d02e4cca206 Resolve @EdDev comments.

Please do not mix rebase with the changes, it makes it impossible to review the diff.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/1d7e0d677e564710ce66f0727bf18d02e4cca206..be167d08db2ed312340fedf6906fafea4000aaf3
Resolve @EdDev comments.

libs/net/traffic_generator.py
sbahar619: I already did, we have stop method for both server and client.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-133 published

sbahar619:
Change: fix typo in commit message.

libs/net/traffic_generator.py
EdDev: I mean to resolve the github thread comments here. Only the owner of the PR or someone who has write permission on the upstream repo can do it physically.

tests/network/user_defined_network/test_user_defined_network.py
```diff
@@ -69,6 +70,26 @@ def vmb_udn_non_migratable(namespace, namespaced_layer2_user_defined_network, ud
...
+def client(vma_udn, vmb_udn_non_migratable):
+    with Client(
+        vm=vma_udn,
+        server_ip=lookup_iface_status(
+            vm=vmb_udn_non_migratable, iface_name=lookup_primary_network(vm=vmb_udn_non_migratable).name
+        )[IP_ADDRESS],
+        server_port=5201,
+    ) as client:
+        assert client.is_running()
+        yield client
```
EdDev, line 90: nit: You could do one:
```python
@pytest.fixture
def client_server_connection(...):
  with Server(...) as server:
    assert server.is_running()
    with Client(...) as client:
      assert client.is_running()
      yield server, client
```
It can even done flat IMO (`with Server(...), Client(...) as server, client`).

But this whole thing is not a must, up to you.

tests/network/user_defined_network/test_user_defined_network.py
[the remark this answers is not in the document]
EdDev: Please resolve the github thread here as well.

EdDev (approved):
Thank you!

The commit messages lines are too long, please see if you can fix them.
We need to add a linter for it to avoid these annoying comments :)

tests/network/user_defined_network/test_user_defined_network.py
sbahar619: Too complicated, nested context manager is not pretty.
Sure it can be flat and return the 2 objects, but I don't see a good justification.
I already self developed this kind of fixture, and its less readable, so I prefered to split it.

sbahar619:
> Thank you!
> 
> The commit messages lines are too long, please see if you can fix them. We need to add a linter for it to avoid these annoying comments :)

What?
I have pre-commit hook that run gitlint locally.
It passed.

sbahar619:
> > Thank you!
> > The commit messages lines are too long, please see if you can fix them. We need to add a linter for it to avoid these annoying comments :)
> 
> What? I have pre-commit hook that run gitlint locally. It passed.

@EdDev 
Thank you for your awesome review!

EdDev:
> > Thank you!
> > The commit messages lines are too long, please see if you can fix them. We need to add a linter for it to avoid these annoying comments :)
> 
> What? I have pre-commit hook that run gitlint locally. It passed.

This is why I said we need one to enforce line length limitation.
See https://cbea.ms/git-commit/

sbahar619:
Change: shorten commit messages for up to 50 characters.
@EdDev

sbahar619:
> > > Thank you!
> > > The commit messages lines are too long, please see if you can fix them. We need to add a linter for it to avoid these annoying comments :)
> > 
> > 
> > What? I have pre-commit hook that run gitlint locally. It passed.
> 
> This is why I said we need one to enforce line length limitation. See https://cbea.ms/git-commit/

Done.

EdDev (approved):

tests/network/user_defined_network/test_user_defined_network.py
```diff
@@ -69,6 +70,26 @@ def vmb_udn_non_migratable(namespace, namespaced_layer2_user_defined_network, ud
...
+
+
+@pytest.fixture
+def client(vma_udn, vmb_udn_non_migratable):
+    with Client(
+        vm=vma_udn,
+        server_ip=lookup_iface_status(
+            vm=vmb_udn_non_migratable, iface_name=lookup_primary_network(vm=vmb_udn_non_migratable).name
+        )[IP_ADDRESS],
+        server_port=5201,
```
yossisegev, line 87: To ensure consistency - don't use the hard-coded port number again for the client, but rather get it from the server object.

yossisegev (changes requested):

tests/network/user_defined_network/test_user_defined_network.py
EdDev: Well, exposing the port number just for this reason is a bit odd. I would just use a constant.
But I have no strong opinion on this one.

Funny, but keeping it a single fixture would have not presented this issue.

sbahar619: Also with single fixture you will need to pass the port twice, so instead of global constant we will have local variable.

sbahar619: We can add this constant:
```
SERVER_PORT: Final[int] = 5201
```

EdDev: > Also with single fixture you will need to pass the port twice, so instead of global constant we will have local variable.

The point is that you define it once.
Same as the constant, but with lower scope.

yossisegev: > Well, exposing the port number just for this reason is a bit odd.

I don't agree that this is odd.
I think it's fine to expose a class member if you need to use it in another class. It totally correlates with the fact that there are 2 members of the **Client** class that are called `_server_ip` and `_server_port`.
However, I'm fine with the compromise of defining a `SERVER_PORT` constant and using it in both classes.

sbahar619: Thanks, done.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/3f5f37c48d43ede37feddd7cbd10158eb4f807c4..5ed3ced03be322dfa8da839ed7ca62569de9a22a
Resolve @yossisegev comment.

yossisegev (approved):
/lgtm

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-133 published

EdDev (approved):

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,116 @@
...
+    def stop(self) -> None:
+        _stop_process(vm=self._vm, cmd=self._cmd)
+
+    def is_running(self) -> bool:
+        return _is_process_running(vm=self._vm, cmd=self._cmd)
+
+
+def _stop_process(vm: BaseVirtualMachine, cmd: str) -> None:
+    try:
+        vm.console(commands=[f"pkill -f '{cmd}'"], timeout=_DEFAULT_CMD_TIMEOUT_SEC)
```
dbasunag, line 98: Don't you need to validate the process was actually killed?

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,116 @@
...
+    def __exit__(self, exc_type: BaseException, exc_value: BaseException, traceback: object) -> None:
+        self.stop()
+
+    def start(self) -> None:
+        self._vm.console(
+            commands=[f"{self._cmd} &"],
+            timeout=_DEFAULT_CMD_TIMEOUT_SEC,
+        )
+
+    def stop(self) -> None:
```
dbasunag, line 47: I don't see the point of having a separate start() and stop(). You can directly call the right function in enter and exit

libs/net/traffic_generator.py
```diff
@@ -0,0 +1,116 @@
...
+        _stop_process(vm=self._vm, cmd=self._cmd)
```
dbasunag, line 90: Same comment here.

dbasunag (changes requested):

libs/net/traffic_generator.py
sbahar619: The _stop_process method is responsible only for attempting to terminate the process. Validation of whether the process was successfully stopped is delegated to the is_running method. This design keeps the responsibilities distinct, maintaining modularity and allowing the caller to decide how to handle the result.

libs/net/traffic_generator.py
sbahar619: ack

sbahar619: Done.

libs/net/traffic_generator.py
sbahar619: Done.

sbahar619:
Change: 
https://github.com/RedHatQE/openshift-virtualization-tests/compare/00d8c2295da0058f6190acc764fe9485a54289d4..e7b3afac5df621592301d3b16208cc64d9497507
Resolve @dbasunag comments.

tests/network/user_defined_network/test_user_defined_network.py
```diff
@@ -69,6 +71,26 @@ def vmb_udn_non_migratable(namespace, namespaced_layer2_user_defined_network, ud
         yield vm
 
 
+@pytest.fixture
```
dbasunag, line 74: Missing `()`

tests/network/user_defined_network/test_user_defined_network.py
```diff
@@ -69,6 +71,26 @@ def vmb_udn_non_migratable(namespace, namespaced_layer2_user_defined_network, ud
...
+def server(vmb_udn_non_migratable):
+    with Server(vm=vmb_udn_non_migratable, port=SERVER_PORT) as server:
+        assert server.is_running()
+        yield server
+
+
+@pytest.fixture
```
dbasunag, line 81: ditto.

libs/net/traffic_generator.py
dbasunag: Should not you then call is_running() in __exit__?

dbasunag (changes requested):

libs/net/traffic_generator.py
EdDev: Why is it needed?
If it fails, what will we do?

I do not expect this to ever fail to kill, it may just be already killed/missing, which is also good.

tests/network/user_defined_network/test_user_defined_network.py
EdDev: @dbasunag is it a convention you want to enforce?
I would use linters for this to avoid the need to waste time on these.

EdDev (approved):

libs/net/traffic_generator.py
sbahar619: Since _stop_process handles non-existing processes gracefully, adding is_running() would be redundant and unnecessary.

tests/network/user_defined_network/test_user_defined_network.py
sbahar619: ack

sbahar619: Done.

tests/network/user_defined_network/test_user_defined_network.py
sbahar619: Done.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/da3084fa304c40380d20cda433c70cbb61183b52..7d13370a86252d417cdc4253f9da87e270a27663
Resolve @dbasunag comments.

EdDev (approved):

libs/net/traffic_generator.py
dbasunag: vm.console() already uses verify_commands_output=True. So this comment is not relevant.

dbasunag (approved):

redhat-cnv-qe-bot1:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-133.

redhat-cnv-qe-bot1:

PR #138, opened 2025-01-16
Remove ErrorUnschedulable from VM_ERROR_STATUSES

hmeir:
##### Short description:
Currenly, VMS that reaches this status - fails the test immidietly.
But this is a recoverable one for a VM.
So we should give a chance for those VM's to recover.

##### More details:
for example, the printable status in this case is `ErrorUnschedulable.` at first, but after less that one min its recovering:

```
{'lastProbeTime': None, 'lastTransitionTime': '2025-01-09T05:00:18Z', 'message': '0/3 nodes are available: 3 Insufficient bridge.network.kubevirt.io/mg-br1. preemption: 0/3 nodes are available: 3 No preemption victims found for incoming pod.', 'reason': 'Unschedulable', 'status': 'False', 'type': 'PodScheduled 
```

Some scheduled t2 + upgrade + post uprade jobs are unstable because of it.

Maybe its a specific issue with compact cluster.
But anyway, since this error is recoverable it shouldn't fail tests right away in general.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-54781

### DIFF
--- a/utilities/virt.py
+++ b/utilities/virt.py
@@ -88,7 +88,6 @@ CIRROS_IMAGE = "kubevirt/cirros-container-disk-demo:latest"
 FLAVORS_EXCLUDED_FROM_CLOUD_INIT = (OS_FLAVOR_WINDOWS, OS_FLAVOR_CIRROS)
 VM_ERROR_STATUSES = [
     VirtualMachine.Status.CRASH_LOOPBACK_OFF,
-    VirtualMachine.Status.ERROR_UNSCHEDULABLE,
     VirtualMachine.Status.ERROR_PVC_NOT_FOUND,
     VirtualMachine.Status.IMAGE_PULL_BACK_OFF,
     VirtualMachine.Status.ERR_IMAGE_PULL,
@@ -1575,11 +1574,7 @@ def get_rhel_os_dict(rhel_version):
 
 def assert_vm_not_error_status(vm):
     vm_status = vm.printable_status
-    error_list = VM_ERROR_STATUSES.copy()
-    vm_devices = vm.instance.spec.template.spec.domain.devices
-    if vm_devices.gpus:
-        error_list.remove(VirtualMachine.Status.ERROR_UNSCHEDULABLE)
-    assert vm_status not in error_list, f"VM {vm.name} error status: {vm_status}"
+    assert vm_status not in VM_ERROR_STATUSES, f"VM {vm.name} error status: {vm_status}"
 
 
 def wait_for_running_vm(


redhat-cnv-qe-bot1:

The following are automatically added:
 * Add reviewers from OWNER file (in the root of the repository) under reviewers section.
 * Set PR size label.
 * New issue is created for the PR. (Closed when PR is merged/closed)
 * Run [pre-commit](https://pre-commit.ci/) if `.pre-commit-config.yaml` exists in the repo.

Available user actions:
 * To mark PR as WIP comment `/wip` to the PR, To remove it from the PR comment `/wip cancel` to the PR.
 * To block merging of PR comment `/hold`, To un-block merging of PR comment `/hold cancel`.
 * To mark PR as verified comment `/verified` to the PR, to un-verify comment `/verified cancel` to the PR.
        verified label removed on each new commit push.
 * To cherry pick a merged PR comment `/cherry-pick <target branch to cherry-pick to>` in the PR.
    * Multiple target branches can be cherry-picked, separated by spaces. (`/cherry-pick branch1 branch2`)
    * Cherry-pick will be started when PR is merged
 * To build and push container image command `/build-and-push-container` in the PR (tag will be the PR number).
    * You can add extra args to the Podman build command
        * Example: `/build-and-push-container --build-arg OPENSHIFT_PYTHON_WRAPPER_COMMIT=<commit_hash>`
 * To add a label by comment use `/<label name>`, to remove, use `/<label name> cancel`
 * To assign reviewers based on OWNERS file use `/assign-reviewers`
 * To check if PR can be merged use `/check-can-merge`
 * to assign reviewer to PR use `/assign-reviewer @<reviewer>`

<details>
<summary>Supported /retest check runs</summary>

 * `/retest tox`: Retest tox
 * `/retest build-container`: Retest build-container
 * `/retest all`: Retest all

</details>

<details>
<summary>Supported labels</summary>

 * hold
 * verified
 * wip
 * lgtm

</details>

EdDev (approved):
Thanks!

The commit message has some typos and problematic sentences, leaving it to you to fix or not.

dbasunag (approved):

vsibirsk (changes requested):
I don't agree this should go away completely. 
ANY error status is a reason for concern.
For e.g this gpu check is a legit bug - https://issues.redhat.com/browse/CNV-33982 and it was caught only with this check.
Another example - https://issues.redhat.com/browse/CNV-37875. this would've been missed again without the check.

I believe we need to improve this logic and not simply remove the error status from list. For e.g. we catch specific error status (like ERROR_UNSCHEDULABLE and give VM 1-2 attempts or 1-2 min to recover, but still fail if status persists)

EdDev:
> Agree with @vsibirsk, if VM is `ErrorUnschedulable` for a while, but the whole flow is correct and VM will come up - then the VM status is wrong, or there's an underlying bug. Like in this bug: https://issues.redhat.com/browse/CNV-51617 - VM is `ErrorUnschedulable` for 7 minutes, but then comes up with no intervention.

Let me disagree.. :)
If you think the VM should come up in 2min, you should fail on that timeout.
If it fails because you are not accepting acceptable intermediate states, it is a problem of interpreting wrongly the data and basic concepts of a distributed async system.

The unschedulable state is expected when the resources are not yet available.
The state is changes once everything is stabilizing and resources are available.
If you are not willing to wait 1min, 5min or 7min for that stabilization, great... do that from the test, but do not put it in the infra tooling and cause a flake for all tests.
Perhaps that was not emphasized enough in the PR description, but waiting for a intermediate state is raceful and incorrect. It is just a matter of chance to catch it when it passes quickly, which is probably most of the cases for flakes.

vsibirsk:
> it is a problem of interpreting wrongly the data and basic concepts of a distributed async system.

Yes, indeed VM can recover from `ErrorUnschedulable` (but I believe the bugs i shared above show that it's still an issue we need to check at least). There are just to many things `ErrorUnschedulable` can mean to just simply ignore it (at least leave it as a warning??)

> but waiting for a intermediate state is raceful and incorrect. It is just a matter of chance to catch it when it passes quickly, which is probably most of the cases for flakes.

I totally agree that current checks implementation is bad (but it's better then nothing at all!) and should be re-worked

dshchedr:
My 5 cents:

We should not completely ignore this error message. By default we should keep it as is (without any re-try).
the name `ErrorUnschedulable` is clearly say that there is an error.. it does not say "wait, I'm doing something" 

This picture can be very confusing for our customers (regardless how much time it takes between steps):

```
vm-fedora   0s     Starting                             False
vm-fedora   1s     ErrorUnschedulable        False
vm-fedora   20s   Starting                             False
vm-fedora   25s   Running                            True
```

as for me - it looks as a bug. If we ignore it - we can expect our customers opening this bugs instead of us...



> If you think the VM should come up in 2min, you should fail on that timeout.

@EdDev we can't exactly predict how much time it may take for VMs to start - it depends on many things: cluster type, loadage, VM type, VM image size (need to download or not), so we always add some extra time. With big timeout we can miss that the VM is staying in  ErrorUnschedulable state for several minutes (which is definitely sounds like a bug)


I would keep the default verification as is, but if some team ready to take a risk and ignore this error - they can pass extra argument to that function which will remove necessary error from the VM_ERROR_STATUSES list

EdDev:
> > it is a problem of interpreting wrongly the data and basic concepts of a distributed async system.
> 
> Yes, indeed VM can recover from `ErrorUnschedulable` (but I believe the bugs i shared above show that it's still an issue we need to check at least). There are just to many things `ErrorUnschedulable` can mean to just simply ignore it (at least leave it as a warning??)

You ref issues that when looking at the end result shows this error which persists after a timeout.
They are surely not proving that the cause of the problem is the mentioned condition.

As I see it, you are wrongly using a condition as the problem instead of asserting a (more or less) stable state.
If this specific condition is there for enough time (i.e. a timeout) it is a problem and a failure of the test makes sense.
On the other hand, if this condition just popped up, it says nothing about the system being in a problematic state.

Therefore, I stand with my previous statement.
It is simply incorrect to interpret reaching this condition as a problem. It is a problem only if it persists for a defined timeout.
 
> > but waiting for a intermediate state is raceful and incorrect. It is just a matter of chance to catch it when it passes quickly, which is probably most of the cases for flakes.
> 
> I totally agree that current checks implementation is bad (but it's better then nothing at all!) and should be re-worked

Sorry, "better then nothing" in this case is just plain wrong.
As proof, flakes have been identified due to this logic.
I have also tried to explain why it is wrong to interpret it as a problem without waiting enough time.

> @EdDev we can't exactly predict how much time it may take for VMs to start - it depends on many things: cluster type, loadage, VM type, VM image size (need to download or not), so we always add some extra time. With big timeout we can miss that the VM is staying in ErrorUnschedulable state for several minutes (which is definitely sounds like a bug)

Yes, the timeout for a VM may changed per its configuration and setup. In practice, you either choose the most appropriate time for the timeout (e.g. waiting for the "ready" condition) as the highest value observed or per an individual scenario. No matter which one you use, it has nothing to do with the condition mentioned here, which wrongly is assumed to be an unrecoverable error or a problem in general when observed.
I tried to explain why it is an expected condition to appear.

> I would keep the default verification as is, but if some team ready to take a risk and ignore this error - they can pass extra argument to that function which will remove necessary error from the VM_ERROR_STATUSES list

Well, I tried to reduce your flakes and the incorrect assumption that this condition should be a sign of a problem during VM initial start. If you check this condition after the timeout, fine.. it will give some context to the possible problem.
However, it is incorrect to monitor its random appearance and declare it is a problem when seen.

My recommendation to the sig-network is to make sure this is not in use for the network tests.
FYI @yossisegev , @sbahar619 , @Anatw

utilities/virt.py
```diff
@@ -88,7 +88,6 @@
 FLAVORS_EXCLUDED_FROM_CLOUD_INIT = (OS_FLAVOR_WINDOWS, OS_FLAVOR_CIRROS)
 VM_ERROR_STATUSES = [
     VirtualMachine.Status.CRASH_LOOPBACK_OFF,
-    VirtualMachine.Status.ERROR_UNSCHEDULABLE,
```
dbasunag, line 91: This change was added more than a year back if I recall correctly. Has anyone triaged to see why non-virt tests suddenly started failing with ErrorUnschedulable vms?

dbasunag (commented):
May be consider not using running_vms everywhere and have a simplified version of running_vms for scenarios that does not need these many checks?

utilities/virt.py
hmeir: For network, they triaged and its because of nmstate TLS cert renwal(currently every 15 min +-), and appernatly its the intended design 
So its an automation issue, they have a Jira card to fix it

hmeir: I think that we can close this PR - we can open a followup Jira to optimize the code to a retry mechanisem, or add args/another functions etc

Just deleting it seems wrong according to the feedbacks here

EdDev: Sad

EdDev:
I find it very strange that all feedback has not even tried to argue against the facts given on why this behavior is wrong.
Reasoning that it found bugs or that it was here for ages is not a technical reason that explains why it is correct or incorrect.

I can assure you that if I pick a random execution and cause an intentional failure, there is a chance that while you investigate it, you will find some problem.
That by itself is not a good reason to leave things in a bad shape with a known potential flake.

This behavior you are following here is a bug, you should remove it and find the real reason behind failures/bugs.
I cannot force you to do it, but I strongly recommend you rethink this over. Alternatively, provide a technical explanation on why this is correct.

hmeir:
> I find it very strange that all feedback has not even tried to argue against the facts given on why this behavior is wrong. Reasoning that it found bugs or that it was here for ages is not a technical reason that explains why it is correct or incorrect.
> 
> I can assure you that if I pick a random execution and cause an intentional failure, there is a chance that while you investigate it, you will find some problem. That by itself is not a good reason to leave things in a bad shape with a known potential flake.
> 
> This behavior you are following here is a bug, you should remove it and find the real reason behind failures/bugs. I cannot force you to do it, but I strongly recommend you rethink this over. Alternatively, provide a technical explanation on why this is correct.

I agree that the current implementation is bad and should change, And I know that Yossi is working on a re-try machanisem, So this PR is kind of a duplicate

dbasunag:
> I find it very strange that all feedback has not even tried to argue against the facts given on why this behavior is wrong. Reasoning that it found bugs or that it was here for ages is not a technical reason that explains why it is correct or incorrect.
> 
> I can assure you that if I pick a random execution and cause an intentional failure, there is a chance that while you investigate it, you will find some problem. That by itself is not a good reason to leave things in a bad shape with a known potential flake.
> 
> This behavior you are following here is a bug, you should remove it and find the real reason behind failures/bugs. I cannot force you to do it, but I strongly recommend you rethink this over. Alternatively, provide a technical explanation on why this is correct.

@EdDev the code was designed this way to ensure we don't take too long to fail. It is true that is it mostly focused towards virt and storage workflow. From network side, if we need to wait for something to complete, we should add it here https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/utilities/virt.py#L1670. After this point, if the vm is started, dvs are in good shape, why would we expect the vm in error state? Can you please explain why you think it is a bug? In the failures I see for network tests with vms in ErrorUnschedulable, NNCP is already reporting SuccessfullyConfigured. I don't understand why at that point a vm would failed to schedule with 
```0/3 nodes are available: 1 Insufficient bridge.network.kubevirt.io/<bridge name>, 2 node(s) didn't match Pod's node affinity/selector. preemption: 0/3 nodes are available: 1 No preemption victims found for incoming pod, 2 Preemption is not helpful for scheduling.```. 
Do we know why we are seeing that?

yossisegev:
> In the failures I see for network tests with vms in ErrorUnschedulable, NNCP is already reporting SuccessfullyConfigured. I don't understand why at that point a vm would failed to schedule with `0/3 nodes are available: 1 Insufficient bridge.network.kubevirt.io/<bridge name>, 2 node(s) didn't match Pod's node affinity/selector. preemption: 0/3 nodes are available: 1 No preemption victims found for incoming pod, 2 Preemption is not helpful for scheduling.`. Do we know why we are seeing that?

@dbasunag This happens if the NNCP was indeed configured successfully as reported in its status, which means that its configuration is applied on the node (for example - the bridge interface that is requested in the NNCP was created on the node), but the `Node` resource was not yet updated with the bridge marker `bridge.network.kubevirt.io/<bridge name>`, which virt-launcher. presents as a requirement from its designated host.
This is one example of a recoverable state, which justifies waiting when the VM gets to `ErrorUnschedulable`. Once this marker is updated on the node, the VMI will get out of `ErrorUnsched` and continue its spinning up.

EdDev:
> @EdDev the code was designed this way to ensure we don't take too long to fail. It is true that is it mostly focused towards virt and storage workflow. From network side, if we need to wait for something to complete, we should add it here https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/utilities/virt.py#L1670. After this point, if the vm is started, dvs are in good shape, why would we expect the vm in error state? Can you please explain why you think it is a bug?

I basically repeat myself, so I am unsure if I add here any new insight.
It is a bug because the condition you decided to exit early, expresses that the scheduler has not found an available node for the pod to be scheduled on. One nsec after it may find it (because some resource appeared, because memory got freed, because some other reason).
It is a normal operation for having this condition and your interpretation for it to be a problem is wrong IMO.

> In the failures I see for network tests with vms in ErrorUnschedulable, NNCP is already reporting SuccessfullyConfigured. I don't understand why at that point a vm would failed to schedule with
> `0/3 nodes are available: 1 Insufficient bridge.network.kubevirt.io/<bridge name>, 2 node(s) didn't match Pod's node affinity/selector. preemption: 0/3 nodes are available: 1 No preemption victims found for incoming pod, 2 Preemption is not helpful for scheduling.`.
> Do we know why we are seeing that?

Yes, the condition we wait for represents the successful configuration of the node with a linux bridge.
However, a different component that runs in the background (the bridge marker) needs to detect it and the configure the node with it as a resource. Only then the pod in question will have a node to be scheduled on.

But if we look at the bigger picture, waiting serially for things to happen is just not playing well with the Kubernetes/OpenShift ecosystem. The system is distributed, async and the only thing that can be sure about is that it will try to reach a stable state. You can exit early **only** when the condition is final (e.g. for pods it will be the complete and fail states).
If the condition is just expressing an intemediate state, you are not suppose to fail on it.
Moreover, it is probably a bad thing to wait for things serially, it introduces slowness and does not scale.
E.g. when you have 2 VMs, you should start both in parallel and only then wait for a condition like "ready", because then you allow both to start in parallel and reduce the time by half.

dbasunag:
> > > ere https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/utilities/virt.py#L1670. After this point, if the vm is started, dvs are in good shape, why would we expect the vm in error state? Can you please explain why you think it is a bug?
> > 
> > 
> > I basically repeat myself, so I am unsure if I add here any new insight. It is a bug because the condition you decided to exit early, expresses that the scheduler has not found an available node for the pod to be scheduled on. One nsec after it may find it (because some resource appeared, because memory got freed, because some other reason). It is a normal operation for having this condition and your interpretation for it to be a problem is wrong IMO.
> > > In the failures I see for network tests with vms in ErrorUnschedulable, NNCP is already reporting SuccessfullyConfigured. I don't understand why at that point a vm would failed to schedule with
> > > `0/3 nodes are available: 1 Insufficient bridge.network.kubevirt.io/<bridge name>, 2 node(s) didn't match Pod's node affinity/selector. preemption: 0/3 nodes are available: 1 No preemption victims found for incoming pod, 2 Preemption is not helpful for scheduling.`.
> > > Do we know why we are seeing that?
> > 
> > 
> > Yes, the condition we wait for represents the successful configuration of the node with a linux bridge. However, a different component that runs in the background (the bridge marker) needs to detect it and the configure the node with it as a resource. Only then the pod in question will have a node to be scheduled on.
> > But if we look at the bigger picture, waiting serially for things to happen is just not playing well with the Kubernetes/OpenShift ecosystem. The system is distributed, async and the only thing that can be sure about is that it will try to reach a stable state. You can exit early **only** when the condition is final (e.g. for pods it will be the complete and fail states). If the condition is just expressing an intemediate state, you are not suppose to fail on it. Moreover, it is probably a bad thing to wait for things serially, it introduces slowness and does not scale. E.g. when you have 2 VMs, you should start both in parallel and only then wait for a condition like "ready", because then you allow both to start in parallel and reduce the time by half.
> 
> We have 4 options as I see it:
> 
> 1. Get the agreement of storage & virt team to remove this part entirely, and just wait for the VMI running state.
> 2. Apply a retry machinisem, something like a timeout threshold agument(1,2,3 min whatever, depends). that will wait for the vmi running state, but will log a warning every X seconds to track the status for additional info.
> 3. Perform some checks before calling running_vm. for example for bridge, check something like NNCP succeess & node/s has been updated.
>    Something similar to the existing code:
> 
> ```
> def assert_node_is_marked_by_bridge(bridge_nad, vm):
>     for bridge_annotation in bridge_nad.instance.metadata.annotations.values():
>         assert bridge_annotation in vm.vmi.node.instance.status.capacity.keys()
>         assert bridge_annotation in vm.vmi.node.instance.status.allocatable.keys()
> 
> 
> def assert_nmstate_bridge_creation(bridge):
>     bridge_name = bridge.bridge_name
> 
>     # Although the bridge interface was already created, the NodeNetworkState resource takes some time to be updated.
>     sampler = TimeoutSampler(
>         wait_timeout=TIMEOUT_1MIN,
>         sleep=TIMEOUT_5SEC,
>         func=lambda: NodeNetworkState(name=get_node_selector_name(node_selector=bridge.node_selector)).get_interface(
>             name=bridge_name
>         ),
>     )
>     try:
>         for sample in sampler:
>             if sample:
>                 break
>     except TimeoutExpiredError:
>         LOGGER.error(f"Bridge {bridge_name} not found in NNS.")
>         raise
> ```
> 
> 4. Create different version of running_vm function.
> 
> Option 1 will do the job, but again requires everyone agreement. in my opinion each option, or even a combination of them, is cool. Personally I think the 1/2 + 3/4 is the best. wdyt?
> 
> @EdDev @dbasunag @jpeimer @vsibirsk @dshchedr

@hmeir please schedule a meeting on this topic for next week.

dshchedr:
> We have 4 options as I see it:
> 
> 1. Get the agreement of storage & virt team to remove this part entirely, and just wait for the VMI running state.

Lack of the bridge is only one of the reason for getting VM in `ErrorUnschedulable` state
There are multiple other reasons when VM can reach this state and if we skip this error from checking list - we will be just waisting time waiting when VM is active. For example: "Insufficient memory / CPU", lack of specific label on the node (vm with NodeSelector), and more other possible scenarios when we can stop the test run immideately if VM reached ErrorUnschedulable state.

As a virt_node side I would prefer to keep it in the list by default. And if some product behavior is changed - I would like to know about it (even if it is expected) - I want to double check why the VM hangs with this status even if it is only for 20 seconds (quite possible that there could be a good reason to add a note to the documentation)

> 2. Apply a retry machinisem, something like a timeout threshold agument(1,2,3 min whatever, depends). that will wait for the vmi running state, but will log a warning every X seconds to track the status for additional info.

Nobody is going to read Warning messages unless the test is failed, so I would say - it can be helpful only in triaging the test run

> 3. Perform some checks before calling running_vm. for example for bridge, check something like NNCP succeess & node/s has been updated.

+1, that's we usually do - check everything completed before going to next step

> 4. Create different version of running_vm function.

option 5. update `running_vm` function: pass the argument with the state that can be ignored

vsibirsk:
> 1. Get the agreement of storage & virt team to remove this part entirely, and just wait for the VMI running state.

I'm still against this approach. Following this path, we'll just eventually remove all error checks and fail-fast and will be back with 30min of waiting for an unschedulable VM too boot

> 3. Perform some checks before calling running_vm. for example for bridge, check something like NNCP succeess & node/s has been updated.

I believe that is what should be done in the first place. When we create/update some resource and that causes some instability on cluster (for e.g. node update, pods recreation, etc.), the wait for stabilizing should happen in that same fixture/function

> 4. Create different version of running_vm function.

That is also a good point. I think at this point `running_vm` should be split into smaller more specialized functions instead one-for-all-cases massive function

kbidarkar:
On a Production Environment: Error is recoverable, when the needed Infra is updated.

On a Test Environment: Error is not recoverable, as there is no plan to update the Infra, once the Job is triggered. While testing, we need to ensure all the resources are available.

EdDev:
Well, I disagree with all the point raised so far.
However, if you can assure a clean lab environment, keep doing what you do.

Maybe it is worth raising that the test suite is slow because 1) you have flakes and 2) you are not running the tests in parallel like tier-1.

You basically focus on handling failures quickly. I would prefer to see the focus on success oriented execution, i.e. making the tests that pass run quicker.

I would even claim, per your strong dependency on implementation details like the one mentioned here, that the tests are fragile. They may break when the implementation changes or when one will try to start optimizing for async execution.

PR #174, opened 2025-01-23
Refactor VM console proxy test

geetikakay:
- Updated vm_console_proxy_service_account_role_binding fixture to reference ClusterRole instead of Role
- Introduced a new helper function assert_resource_status_and_wait_for_availability to check resource availability status in tests.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-34769

### DIFF
--- a/tests/infrastructure/vm_console_proxy/conftest.py
+++ b/tests/infrastructure/vm_console_proxy/conftest.py
@@ -5,7 +5,6 @@ from subprocess import check_output
 import pytest
 from ocp_resources.cluster_role import ClusterRole
 from ocp_resources.resource import Resource
-from ocp_resources.role import Role
 from ocp_resources.role_binding import RoleBinding
 from ocp_resources.service_account import ServiceAccount
 from ocp_resources.ssp import SSP
@@ -24,6 +23,7 @@ from tests.infrastructure.vm_console_proxy.constants import (
     TOKEN_API_VERSION,
     TOKEN_ENDPOINT,
     VM_CONSOLE_PROXY,
+    VM_CONSOLE_PROXY_CLUSTER_ROLE,
     VM_CONSOLE_PROXY_USER,
 )
 from tests.infrastructure.vm_console_proxy.utils import (
@@ -104,22 +104,6 @@ def vm_for_console_proxy(namespace, unprivileged_client):
 
 
 @pytest.fixture(scope="class")
-def vm_console_proxy_role(namespace):
-    with Role(
-        name="vnc-token-access",
-        namespace=namespace.name,
-        rules=[
-            {
-                "apiGroups": [TOKEN_ENDPOINT],
-                "resources": ["virtualmachines/vnc"],
-                "verbs": ["get"],
-            }
-        ],
-    ) as vm_console_proxy_role:
-        yield vm_console_proxy_role
-
-
-@pytest.fixture(scope="class")
 def vm_console_proxy_service_account(namespace):
     with ServiceAccount(name=f"{VM_CONSOLE_PROXY_USER}1", namespace=namespace.name) as sa:
         yield sa
@@ -140,15 +124,24 @@ def vm_service_account_role_binding(namespace, vm_console_proxy_service_account)
 
 
 @pytest.fixture(scope="class")
-def vm_console_proxy_service_account_role_binding(namespace, vm_console_proxy_service_account, vm_console_proxy_role):
+def vm_console_proxy_cluster_role_exists():
+    assert ClusterRole(name=VM_CONSOLE_PROXY_CLUSTER_ROLE).exists, (
+        f"ClusterRole {VM_CONSOLE_PROXY_CLUSTER_ROLE} not found"
+    )
+
+
+@pytest.fixture(scope="class")
+def vm_console_proxy_service_account_role_binding(
+    namespace, vm_console_proxy_cluster_role_exists, vm_console_proxy_service_account
+):
     with RoleBinding(
         name=f"{VM_CONSOLE_PROXY_USER}-token-access",
         namespace=namespace.name,
         subjects_kind=vm_console_proxy_service_account.kind,
         subjects_name=vm_console_proxy_service_account.name,
         subjects_namespace=namespace.name,
-        role_ref_kind=Role.kind,
-        role_ref_name=vm_console_proxy_role.name,
+        role_ref_kind=ClusterRole.kind,
+        role_ref_name=VM_CONSOLE_PROXY_CLUSTER_ROLE,
     ) as vm_role_service_account_role_binding:
         yield vm_role_service_account_role_binding
 
--- a/tests/infrastructure/vm_console_proxy/utils.py
+++ b/tests/infrastructure/vm_console_proxy/utils.py
@@ -1,7 +1,11 @@
+from __future__ import annotations
+
 import logging
+from typing import Any, Type
 
 import requests
 from ocp_resources.api_service import APIService
+from ocp_resources.resource import Resource
 
 from tests.infrastructure.vm_console_proxy.constants import (
     TOKEN_API_VERSION,
@@ -13,20 +17,20 @@ LOGGER = logging.getLogger(__name__)
 
 
 def create_vnc_console_token(
-    url,
-    endpoint,
-    api_version,
-    namespace,
-    virtual_machine,
-    duration,
-    runtime_headers=None,
-):
+    url: str,
+    endpoint: str,
+    api_version: str,
+    namespace: str,
+    virtual_machine: str,
+    duration: int,
+    runtime_headers: dict[str, str] | None,
+) -> str:
     """
     Requests a VNC console token for a virtual machine
 
     Args:
         url (str): The base URL of API server.
-        endpoint (str): The API endpoint for  VM console proxy.
+        endpoint (str): The API endpoint for VM console proxy.
         api_version (str): The API version for the VM console proxy.
         namespace (str): The namespace of the virtual machine.
         virtual_machine (str): The name of the virtual machine.
@@ -55,7 +59,7 @@ def create_vnc_console_token(
         raise
 
 
-def get_vm_console_proxy_resource(resource_kind, namespace=None):
+def get_vm_console_proxy_resource(resource_kind: Type, namespace: str | None = None) -> Type:
     if namespace:
         vm_console_proxy_resource_object = resource_kind(
             name=VM_CONSOLE_PROXY,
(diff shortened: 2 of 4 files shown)


tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -13,20 +15,20 @@
...
-    duration,
-    runtime_headers=None,
-):
+    url: str,
+    endpoint: str,
+    api_version: str,
+    namespace: str,
+    virtual_machine: str,
+    duration: int,
+    runtime_headers: Optional[Dict[str, str]] = None,
```
dbasunag, line 24: please use `dict` and `dict[str, str] | None` should work

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -55,7 +57,7 @@ def create_vnc_console_token(
         raise
 
 
-def get_vm_console_proxy_resource(resource_kind, namespace=None):
+def get_vm_console_proxy_resource(resource_kind: Type, namespace: Optional[str] = None) -> Type:
```
dbasunag, line 60: Same comment here. No need to have Optional. For resource_kind, you can only have two options from what I see. So please specify and use `|`

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -66,3 +68,21 @@ def get_vm_console_proxy_resource(resource_kind, namespace=None):
...
+
+def assert_resource_status_and_wait_for_availability(resource: Any, timeout: int) -> None:
+    assert resource.exists, (
+        f"Missing: {resource.kind}/{resource.name} under {getattr(resource, 'namespace', 'cluster')}"
+    )
+
+    if not hasattr(resource, "status"):
+        LOGGER.info(f"Resource {resource.kind} has no status. Skipping availability check.")
+        return
+
```
dbasunag, line 81: Should not it be a bug?

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -66,3 +68,21 @@ def get_vm_console_proxy_resource(resource_kind, namespace=None):
...
+    if getattr(resource.instance.status, "conditions", None):
+        LOGGER.info(f"Checking availability of {resource.kind}")
+        resource.wait_for_condition(
+            condition=Resource.Condition.AVAILABLE,
+            status=Resource.Condition.Status.TRUE,
+            timeout=timeout,
+        )
```
dbasunag, line 88: If you are making this call, I am assuming status.conditions are supposed to be present? This won't fail if it is not present. That seems like an automation bug to me.

dbasunag (changes requested):

tests/infrastructure/vm_console_proxy/utils.py
geetikakay: example : clusterroles  they don't have status

tests/infrastructure/vm_console_proxy/utils.py
geetikakay: resource_kind: is list of 3 cluster and  list of 5 namespaced resources

tests/infrastructure/vm_console_proxy/conftest.py
```diff
@@ -140,15 +124,18 @@ def vm_service_account_role_binding(namespace, vm_console_proxy_service_account)
 
 
 @pytest.fixture(scope="class")
-def vm_console_proxy_service_account_role_binding(namespace, vm_console_proxy_service_account, vm_console_proxy_role):
+def vm_console_proxy_service_account_role_binding(namespace, vm_console_proxy_service_account):
+    assert ClusterRole(name=VM_CONSOLE_PROXY_CLUSTER_ROLE).exists, (
+        f"ClusterRole {VM_CONSOLE_PROXY_CLUSTER_ROLE} not found"
+    )
```
dbasunag, line 130: Should be a fixture

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -55,7 +59,7 @@ def create_vnc_console_token(
         raise
 
 
-def get_vm_console_proxy_resource(resource_kind, namespace=None):
+def get_vm_console_proxy_resource(resource_kind: Type, namespace: str | None) -> Type:
```
dbasunag, line 62: Please update the resource_kind type. There are only 3 options.

tests/infrastructure/vm_console_proxy/utils.py
dbasunag: In that case please don't combine validations of all in one utility. This would lead to hiding bugs. e.g. there is no check in place that ensures status check does not get skipped for non clusterrole resources.

dbasunag (changes requested):

tests/infrastructure/vm_console_proxy/utils.py
geetikakay: yeah , it could happen, so you are suggesting to check for selective resources that have status.conditions?

tests/infrastructure/vm_console_proxy/conftest.py
```diff
@@ -140,15 +124,24 @@ def vm_service_account_role_binding(namespace, vm_console_proxy_service_account)
 
 
 @pytest.fixture(scope="class")
-def vm_console_proxy_service_account_role_binding(namespace, vm_console_proxy_service_account, vm_console_proxy_role):
+def assert_vm_console_proxy_cluster_role():
```
dbasunag, line 127: no verbs in fixture name please. May be name it `console_proxy_cluster_role` ? Also, return it if it exist, else raise resource not found exception.

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -13,20 +17,20 @@
...
-    duration,
-    runtime_headers=None,
-):
+    url: str,
+    endpoint: str,
+    api_version: str,
+    namespace: str,
+    virtual_machine: str,
+    duration: int,
+    runtime_headers: Dict[str, str] | None,
```
dbasunag, line 26: please use `dict` instead

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -66,3 +70,27 @@ def get_vm_console_proxy_resource(resource_kind, namespace=None):
...
+    assert resource.exists, (
+        f"Missing: {resource.kind}/{resource.name} under {getattr(resource, 'namespace', 'cluster')}"
+    )
+
+    valid_kinds = {"APIService", "Service", "Deployment"}  # valid kinds that require an availability check
+    if resource.kind not in valid_kinds:
+        LOGGER.info(f"Resource {resource.kind} is not in the list of valid resources for availability check. Skipping.")
+        return
+
+    if not hasattr(resource, "status"):
```
dbasunag, line 85: Can it not have status?

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -66,3 +70,27 @@ def get_vm_console_proxy_resource(resource_kind, namespace=None):
...
+        raise ValueError(f"Resource {resource.kind} does not have a 'status' attribute. Expected status")
+
+    conditions = getattr(resource.instance.status, "conditions", None)
```
dbasunag, line 88: ditto, can it not have status.conditions?

dbasunag (changes requested):

tests/infrastructure/vm_console_proxy/utils.py
geetikakay: we expected that it will have but in case there is a bug, we can catch it here

tests/infrastructure/vm_console_proxy/utils.py
geetikakay: we expected that it will have but in case there is a bug, we can catch it here

tests/infrastructure/vm_console_proxy/conftest.py
geetikakay: I just want to use VM_CONSOLE_PROXY_CLUSTER_ROLE in fixture vm_console_proxy_service_account_role_binding so just to be sure it exists.

tests/infrastructure/vm_console_proxy/utils.py
dbasunag: why would resource.instance.status not catch the same bug? Can't it take a few seconds to populate status? Are you sure that would be a bug?

tests/infrastructure/vm_console_proxy/utils.py
dbasunag: and why can't we use resource.instance.status.conditions to catch the same bug? Also, does not it take sometimes for conditions to get updated? Are you sure this would be a bug, if you check it too early?

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -66,3 +70,27 @@ def get_vm_console_proxy_resource(resource_kind, namespace=None):
             name=f"{TOKEN_API_VERSION}.{TOKEN_ENDPOINT}" if resource_kind == APIService else VM_CONSOLE_PROXY
         )
     return vm_console_proxy_resource_object
+
+
+def assert_resource_status_and_wait_for_availability(resource: Any, timeout: int) -> None:
+    assert resource.exists, (
+        f"Missing: {resource.kind}/{resource.name} under {getattr(resource, 'namespace', 'cluster')}"
+    )
+
```
dbasunag, line 79: This should not be part of this function. If you are checking status and waiting for conditions, the resource should exist at that point.

dbasunag (changes requested):

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-174 published

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-174 published

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-174 published

tests/infrastructure/vm_console_proxy/utils.py
```diff
@@ -66,3 +70,18 @@ def get_vm_console_proxy_resource(resource_kind, namespace=None):
...
+    if resource.kind in {"APIService", "Deployment"}:
+        LOGGER.info(f"Availability/condition check for {resource.kind}.")
+        resource.wait_for_condition(
+            condition=Resource.Condition.AVAILABLE,
+            status=Resource.Condition.Status.TRUE,
+            timeout=timeout,
+        )
+        return
+
+    assert resource.exists, (
```
dbasunag, line 85: Should this be moved above? If you are not sure if the resource actually exists, then using wait_for_condition() does not seem right.

geetikakay: resource.wait_for_condition will check if resource exists and then wait for condition 
if i move assert up , it will e checked for all resources so i will explictly check exists when resource kind != {"APIService", "Deployment"}

dbasunag (approved):

dshchedr (approved):

cnv-qe-bot1:
Failed to delete tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-174. Please delete it manually.

cnv-qe-bot1:

PR #186, opened 2025-01-27
Test kubevirt_vmi_number_of_outdated

OhadRevah:
##### Short description:
Test kubevirt_vmi_number_of_outdated metric.


tests/observability/metrics/test_vms_metrics.py
```diff
@@ -471,3 +471,14 @@ def test_kubevirt_vmi_non_evictable(
...
+
+class TestOutdatedVmi:
+    @pytest.mark.post_upgrade
+    @pytest.mark.polarion("CNV-11749")
+    def test_metric_kubevirt_vmi_number_of_outdated(self, prometheus, kubevirt_resource):
+        validate_metrics_value(
+            prometheus=prometheus,
+            metric_name="kubevirt_vmi_number_of_outdated",
+            expected_value=kubevirt_resource.instance.status.outdatedVirtualMachineInstanceWorkloads,
+        )
```
dbasunag, line 484: This test should be in upgrade lane. 1) before upgrade value should be 0 and after upgrade the value should be non-zero.

dbasunag (changes requested):

dbasunag (changes requested):
Open comment. PR needs to be changed significantly.

tests/observability/upgrade/test_upgrade_observability.py
```diff
@@ -0,0 +1,38 @@
...
+@pytest.mark.upgrade
+@pytest.mark.usefixtures("running_vm_with_bridge")
+class TestUpgradeObservability:
+    """Pre-upgrade tests"""
+
+    @pytest.mark.polarion("CNV-11749")
+    @pytest.mark.order(before=KUBEVIRT_VMI_NUMBER_OF_OUTDATED)
+    @pytest.mark.dependency(name=f"test_metric_{KUBEVIRT_VMI_NUMBER_OF_OUTDATED}")
+    def test_metric_kubevirt_vmi_number_of_outdated_before_upgrade(self, prometheus):
+        validate_metrics_value(
```
dbasunag, line 20: should you also consider validating that kubevirt_resource.instance.status.outdatedVirtualMachineInstanceWorkloads == 0

tests/observability/upgrade/test_upgrade_observability.py
```diff
@@ -0,0 +1,38 @@
...
+    """ Post-upgrade tests """
+
+    @pytest.mark.polarion("CNV-11758")
+    @pytest.mark.order(after=IUO_UPGRADE_TEST_ORDERING_NODE_ID)
+    @pytest.mark.dependency(
+        depends=[f"test_metric_{KUBEVIRT_VMI_NUMBER_OF_OUTDATED}", IUO_UPGRADE_TEST_DEPENDENCY_NODE_ID],
+        scope=DEPENDENCY_SCOPE_SESSION,
+    )
+    def test_metric_kubevirt_vmi_number_of_outdated_after_upgrade(self, prometheus):
+        wait_for_non_zero_metric_value(
```
dbasunag, line 35: This would never go to zero. network lane's vms are always going to report outdated. You should ensure the number is > 0

dbasunag (changes requested):

tests/observability/upgrade/utils.py
```diff
@@ -0,0 +1,26 @@
...
+        metrics_name=metric_name,
+    )
+    sample = None
+    try:
+        for sample in samples:
+            if sample and int(sample) > 0:
+                return
+    except TimeoutExpiredError:
+        LOGGER.info(f"Metric value of: {metric_name} is: {sample}, expected value: non zero")
+        raise
```
hmeir, line 26: Should we **wait** to 0?
Simple assert should be enough in my opinion

hmeir (changes requested):

tests/observability/upgrade/test_upgrade_observability.py
```diff
@@ -0,0 +1,39 @@
...
+from tests.observability.upgrade.utils import wait_for_greater_than_zero_metric_value
+from tests.observability.utils import validate_metrics_value
+from tests.upgrade_params import IUO_UPGRADE_TEST_DEPENDENCY_NODE_ID, IUO_UPGRADE_TEST_ORDERING_NODE_ID
+from utilities.constants import DEPENDENCY_SCOPE_SESSION
+
+KUBEVIRT_VMI_NUMBER_OF_OUTDATED = "kubevirt_vmi_number_of_outdated"
+
+
+@pytest.mark.upgrade
+@pytest.mark.usefixtures("running_vm_with_bridge")
```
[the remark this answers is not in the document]
dbasunag: This vm is not livemigratable. So as long as it is created, the metrics would be non-zero.

tests/observability/upgrade/test_upgrade_observability.py
```diff
@@ -0,0 +1,39 @@
...
+class TestUpgradeObservability:
+    """Pre-upgrade tests"""
+
+    @pytest.mark.polarion("CNV-11749")
+    @pytest.mark.order(before=KUBEVIRT_VMI_NUMBER_OF_OUTDATED)
```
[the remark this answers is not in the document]
dbasunag: +1

tests/observability/upgrade/test_upgrade_observability.py
```diff
@@ -0,0 +1,39 @@
...
+    @pytest.mark.dependency(name=f"test_metric_{KUBEVIRT_VMI_NUMBER_OF_OUTDATED}")
```
dbasunag, line 18: Please match the name with the test name.

tests/observability/upgrade/test_upgrade_observability.py
```diff
@@ -0,0 +1,39 @@
...
+        validate_metrics_value(
+            prometheus=prometheus,
+            metric_name=KUBEVIRT_VMI_NUMBER_OF_OUTDATED,
+            expected_value="0",
+        )
+
+    """ Post-upgrade tests """
+
+    @pytest.mark.polarion("CNV-11758")
+    @pytest.mark.order(after=IUO_UPGRADE_TEST_ORDERING_NODE_ID)
```
dbasunag, line 30: No need for order marker. Dependency would take care of it.

tests/observability/upgrade/test_upgrade_observability.py
```diff
@@ -0,0 +1,39 @@
...
+    @pytest.mark.dependency(
+        depends=[f"test_metric_{KUBEVIRT_VMI_NUMBER_OF_OUTDATED}", IUO_UPGRADE_TEST_DEPENDENCY_NODE_ID],
+        scope=DEPENDENCY_SCOPE_SESSION,
+    )
+    def test_metric_kubevirt_vmi_number_of_outdated_after_upgrade(self, prometheus):
+        wait_for_greater_than_zero_metric_value(
+            prometheus=prometheus,
+            metric_name=KUBEVIRT_VMI_NUMBER_OF_OUTDATED,
+        )
```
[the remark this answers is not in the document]
dbasunag: This test should not run on ocp upgrade. for EUS and CNV this is a legit scenario.

tests/observability/upgrade/utils.py
dbasunag: Nope. the network vms would never get updated - due to lack of matching resources. So waiting would be of no value.

dbasunag (changes requested):

tests/install_upgrade_operators/product_upgrade/test_upgrade_iuo.py
```diff
@@ -74,3 +75,15 @@ def test_nodes_labels_after_upgrade(
...
+    @pytest.mark.order(before=IMAGE_UPDATE_AFTER_UPGRADE_NODE_ID, after=IUO_CNV_ALERT_ORDERING_NODE_ID)
+    @pytest.mark.dependency(
+        depends=[IUO_UPGRADE_TEST_DEPENDENCY_NODE_ID],
+        scope=DEPENDENCY_SCOPE_SESSION,
+    )
+    def test_metric_kubevirt_vmi_number_of_outdated_after_upgrade(self, prometheus):
+        wait_for_greater_than_zero_metric_value(
+            prometheus=prometheus,
+            metric_name="kubevirt_vmi_number_of_outdated",
+        )
```
dshchedr, line 89: you are not creating any VMs, how you would like to check VMI metric without any VM?

dshchedr (changes requested):

tests/install_upgrade_operators/product_upgrade/test_upgrade_iuo.py
hmeir: We have the network VM's created befoer the upgrade, so the metric will shouldn't return 0

btw, maybe we should verify that the number of outdated VMS(either by iterating over them/getting the value from kv CR) is the same as the metric value?
non-zero is enough?
@dbasunag @OhadRevah

Ahmad-Hafe: is it always we will have outdated vms after upgrade? 
are we rely on the network vms that remain after upgarde?

dshchedr: we should never rely on other teams because we may run tests separately

dshchedr (changes requested):

tests/install_upgrade_operators/product_upgrade/test_upgrade_iuo.py
hmeir: @dshchedr 
While I understand your argument, totally a good one, we have vms_for_upgrade using RWO storage & network vm with bridge.
Do we really need to create another one which is kinda the same?

In addition this test testing the metric functionallity.
Maybe the right approach is not to test that it returns none zero value, but the correct value given the number of outdated VM's at the cluster?

dbasunag (changes requested):

tests/install_upgrade_operators/product_upgrade/test_upgrade_iuo.py
dshchedr: @dbasunag add dependency on other team VMs is an option. But in my opinion it is still non reliable: what if tomorrow network team decide to remove all VMs with bridge from test scope? Or for example these VMs become migratable?

I still think it would be better to create very tiny VM with node selector, as I understand - we even don't need to check if it is bootable.. 

But I'm not going to hold PR because of this

> Maybe the right approach is not to test that it returns none zero value, but the correct value given the number of outdated VM's at the cluster?

@hmeir Checking exact number would be definitely the best way, but it can be very difficult to predict how many VMs are non-migratable. Bridge - is not the only possible reason

tests/install_upgrade_operators/product_upgrade/test_upgrade_iuo.py
```diff
@@ -74,3 +75,15 @@ def test_nodes_labels_after_upgrade(
             nodes_labels_before_upgrade=nodes_labels_before_upgrade,
             cnv_upgrade=cnv_upgrade,
         )
+
+    @pytest.mark.polarion("CNV-11758")
```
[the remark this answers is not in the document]
Ahmad-Hafe: +1

tests/install_upgrade_operators/product_upgrade/conftest.py
```diff
@@ -603,3 +604,17 @@ def upgraded_odf(
     updated_odf_subscription_source,
 ):
     wait_for_odf_update(target_version=odf_version)
+
+
+@pytest.fixture(scope="session")
+def cirros_vm_with_node_selector(admin_client, namespace, worker_node1):
+    with create_cirros_vm(
+        storage_class=py_config["default_storage_class"],
```
dshchedr, line 612: create VM with DV seems too heavy
I think you can just create VM with `VirtualMachineForTests` class and pass node_selector there:
```
node_selector=get_node_selector_dict(node_selector=worker_node1.name)
```

or another option - use nodeAffinity `requiredDuringSchedulingIgnoredDuringExecution` option, we have some examples in repo

Ahmad-Hafe: +1 make sense 
if the purpose is just to create non migrate-able vm

OhadRevah:
Closing this PR and open a new one for this.

PR #240, opened 2025-02-05
Fix eus_hco_target_version

hmeir:
##### Short description:
_**"Cherry pick" from 4.16 branch.**_

 1. eus_hco_target_version should use fixtures from the same scope(might fail optin test)
2. cnv_version fixture is redudant - can be removed.
3. change the names of fixtures and util function to match their purpose:
   getting the cnv name of the target version

##### More details:
the name changes:

```
      hco_target_version -> hco_target_version_csv_name
      eus_hco_target_version -> eus_hco_target_version_csv_name
      get_hco_version -> get_hco_csv_name_by_version
```

post upgrade test using this fixture may fail.
In addition the fixture/naming test is misleading.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,7 +17,6 @@ from signal import SIGINT, SIGTERM, getsignal, signal
 from subprocess import check_output
 
 import bcrypt
-import packaging.version
 import paramiko
 import pytest
 import requests
@@ -67,6 +66,7 @@ from ocp_resources.virtual_machine_instance_migration import (
 from ocp_resources.virtual_machine_instancetype import VirtualMachineInstancetype
 from ocp_resources.virtual_machine_preference import VirtualMachinePreference
 from ocp_utilities.monitoring import Prometheus
+from packaging.version import Version, parse
 from pytest_testconfig import config as py_config
 from timeout_sampler import TimeoutSampler
 
@@ -176,7 +176,7 @@ from utilities.network import (
 from utilities.operator import (
     cluster_with_icsp,
     disable_default_sources_in_operatorhub,
-    get_hco_version_name,
+    get_hco_csv_name_by_version,
     get_machine_config_pool_by_name,
 )
 from utilities.ssp import get_data_import_crons, get_ssp_resource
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -1989,20 +1989,33 @@ def rhel_latest_os_params():
...
 @pytest.fixture(scope="session")
 def cnv_target_version(pytestconfig):
     return pytestconfig.option.cnv_version
 
 
+@pytest.fixture(scope="session")
+def eus_target_cnv_version(pytestconfig, cnv_current_version):
+    cnv_current_version = Version(version=cnv_current_version)
+    minor = cnv_current_version.minor
+    # EUS-to-EUS updates are only viable between even-numbered minor versions, exit if non-eus version
```
vsibirsk, line 2010:
```suggestion
    # EUS-to-EUS upgrades are only viable between even-numbered minor versions, exit if non-eus version
```

tests/conftest.py
```diff
@@ -1989,20 +1989,33 @@ def rhel_latest_os_params():
 
 
 @pytest.fixture(scope="session")
-def hco_target_version(cnv_target_version):
-    return get_hco_version_name(cnv_target_version=cnv_target_version)
+def hco_target_csv_name(cnv_target_version):
+    return get_hco_csv_name_by_version(cnv_target_version=cnv_target_version) if cnv_target_version else None
```
vsibirsk, line 1993: i think `get_hco_csv_name_by_version` should return `None` if not `cnv_target_version`

vsibirsk (changes requested):

tests/conftest.py
hmeir: But cnv_target_version(at the condition) is a fixture that can have a value/None
I done want to even execute the function if its None

dbasunag: I agree with Harel. We should not call the function if cnv_target_version does not exist.

dbasunag (approved):

tests/conftest.py
vsibirsk: ack

vsibirsk (approved):

dshchedr (approved):

PR #250, opened 2025-02-06
remove skip_if_not_three_nodes fixture

dbasunag:
##### Short description:
Test is not marked with sno marker. So it won't be executed on sno.

### DIFF
--- a/tests/network/nmstate/test_nncp_maxunavailable.py
+++ b/tests/network/nmstate/test_nncp_maxunavailable.py
@@ -53,12 +53,6 @@ def enable_threading_get_intermediate_nnce_nodes(policy, workers):
         return actual_state
 
 
-@pytest.fixture()
-def skip_if_not_three_nodes(schedulable_nodes):
-    if len(schedulable_nodes) != NUM_OF_DESIRED_WORKERS:
-        pytest.skip(f"Only run on {NUM_OF_DESIRED_WORKERS} worker nodes")
-
-
 @pytest.fixture(scope="session")
 def label_max_unavailable_nodes(schedulable_nodes):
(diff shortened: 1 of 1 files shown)


tests/network/nmstate/test_nncp_maxunavailable.py
```diff
@@ -111,7 +105,6 @@ def maxunavailable_input_for_bridge_creation(request, hosts_common_available_por
     indirect=["maxunavailable_input_for_bridge_creation"],
 )
 def test_create_policy_get_status(
-    skip_if_not_three_nodes,
```
jpeimer, line 114: The test params depend on the number of expected workers, so it will fail on a 4-node cluster for example

Ahmad-Hafe: +1

for example:
NUM_OF_DESIRED_WORKERS is set to 3

where: 
def enable_threading_get_intermediate_nnce_nodes(policy, workers):
    actual_state = []
    with ThreadPoolExecutor(max_workers=**NUM_OF_DESIRED_WORKERS**) as executor:

dbasunag: Do we support 4 nodes cluster currently @jpeimer / @Ahmad-Hafe ?

dbasunag:
/build-and-push-container --build-arg OPENSHIFT_PYTHON_WRAPPER_COMMIT=654b294182802b0ba0034388054cca414c0b2d0f

yossisegev (approved):

cnv-qe-bot1:
Failed to delete tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-250. Please delete it manually.

PR #266, opened 2025-02-10
remove skip_if_no_multinic_nodes and skip_no_bond_support

dbasunag:


README.md
```diff
@@ -8,6 +8,11 @@ When running Windows tests, the cluster should have at least 16GiB RAM (XL deplo
 and 80G volume size (default deployment configuration).
 
 Upgrade tests must be run against a large deployment(24GiB RAM, 250GB volume size)
+Upgrade network tests can't be run against a non-multi-nic cluster. To run upgrade against
+such clusters, we must ignore network component
+```bash
+--ignore=tests/network/
+```
```
EdDev, line 15: > To run upgrade against such clusters, we must ignore network component.

I do not understand this point and how it is related to the skips removed.
I thought there are dedicated upgrade tests for each SIG.

tests/conftest.py
```diff
@@ -646,20 +646,6 @@ def nodes_occupied_nics(nodes_active_nics):
     return {node: nodes_active_nics[node]["occupied"] for node in nodes_active_nics.keys()}
 
 
-@pytest.fixture(scope="session")
-def multi_nics_nodes(hosts_common_available_ports):
-    """
-    Check if nodes has any available NICs
-    """
-    return len(hosts_common_available_ports) > 1
```
EdDev, line 654: WDYT about adding an autouse with an assertion in the `network/condftest.py` about this? It will help fail faster in case the cluster does not confirm to the basic requirements.
The fixture can be named "base_requirements" to emphasize what are the checks all about. We can add here all relevant requirements that are common to the tests in this level.

EdDev (commented):
Thanks.
I had two inline question/comments.

tests/conftest.py
dbasunag: yeah, I think we should have a network sanity for network lane, that should ensure that the cluster is multi-nic. I will do that as a part of this PR.

dshchedr (approved):

tests/network/conftest.py
```diff
@@ -217,3 +221,18 @@ def cnao_deployment(hco_namespace):
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def network_sanity(hosts_common_available_ports, junitxml_plugin):
+    """
+    Perform verification that the cluster is a multi-nic one otherwise exit run
+    """
+    LOGGER.info("Verify cluster running network tests is a multi-nic one")
+    if len(hosts_common_available_ports) <= 1:
+        exit_pytest_execution(
```
EdDev, line 233: Can you please explain why use this one instead of a simple `assert hosts_common_available_ports > 1` ?
This one also seems to collect must-gather and do other stuff.

dbasunag: For consistency. It can be expanded for other checks. Must-gather collection would be skipped for this one.

tests/network/conftest.py
```diff
@@ -231,3 +221,20 @@ def cnao_deployment(hco_namespace):
...
         deployment_name=CLUSTER_NETWORK_ADDONS_OPERATOR,
     )
+
+
+@pytest.fixture(scope="session", autouse=True)
+def network_sanity(hosts_common_available_ports, junitxml_plugin):
+    """
+    Perform verification that the cluster is a multi-nic one otherwise exit run
+    """
+    network_sanity_failure_return_code = 91
```
yossisegev, line 231: Can you please explain what is this error value (91)?

dshchedr: may be it worth to add a short comment line?

dbasunag: exit_pytest_execution() already explains it.

dshchedr (approved):

tests/network/conftest.py
yossisegev: > exit_pytest_execution() already explains it.

I went over the function but didn't find an explanation for this return code.
And anyway - this code is explicitly used here, so I think this is the right place to explain its meaning.
Actually, I think that the more preferred way would be to store this number in a constant with a meaningful name (even if it is used once), otherwise it looks like a magic number.

yossisegev (changes requested):

yossisegev (approved):

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-266 published

dshchedr (approved):

EdDev (approved):

dshchedr (dismissed):

EdDev:
@myakove , there are so many votes here.. I invite you to join the party. :)

dbasunag:
> I am good with the content of this change. However, after learning that this branch is the one which the eminent release is tracking, I find it very risky.
> 
> We are already in a bad situation, I do not think we should make it worse. However, this may be too late to change now, so I am not blocking this. Just make sure you know the risks.

I don't like the idea of code freeze etc. on automation branches. Automation branches does not do releases the same way product does. Unless QE and Dev starts working on features and delivers tests and features together, for all the components, it is not a feasible idea.

EdDev:
> I don't like the idea of code freeze etc. on automation branches. Automation branches does not do releases the same way product does. Unless QE and Dev starts working on features and delivers tests and features together, for all the components, it is not a feasible idea.

It just happened, and the results caused so much harm, at the level that the release is at risk.
I do not know what else should happen to see the problem manifested.

dbasunag:
> > I don't like the idea of code freeze etc. on automation branches. Automation branches does not do releases the same way product does. Unless QE and Dev starts working on features and delivers tests and features together, for all the components, it is not a feasible idea.
> 
> It just happened, and the results caused so much harm, at the level that the release is at risk. I do not know what else should happen to see the problem manifested.

Let's move this conversation off this PR. Please feel free to schedule a meeting - including all the CNV QE and their managers.

tests/network/conftest.py
```diff
@@ -230,3 +220,21 @@ def cnao_deployment(hco_namespace):
 @pytest.fixture(scope="session")
 def ovn_kubernetes_cluster(admin_client):
     return get_cluster_cni_type(admin_client=admin_client) == "OVNKubernetes"
+
+
+@pytest.fixture(scope="session", autouse=True)
+def network_sanity(hosts_common_available_ports, junitxml_plugin):
```
myakove, line 226: I would name it something like `multi_nic_cluster`.

This is the opposite from CNV decision to mark tests that should run on specific lanes.
I understood that there are two clusters type, regular and special and assume that this full into special cluster type.
You eliminated the skip but introduced new problem that you don't know if those tests run or not (it will be dynamic)

Network tests should be marked as `@special`.

dbasunag, line 226: @myakove For some reason, I don't see your comment, even though I see it it email notification. `@special_infra` is for fully capable BM clusters. PSI currently supports multi-nic too. Network team choose to not have their multi-nic tests marked, since their test jobs always have the requirement to run against a multi-nic cluster. Keeping the name to generic, because it can be expanded later.

dshchedr (approved):

vsibirsk (approved):

yossisegev (approved):

redhat-cnv-qe-bot1:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-266.

redhat-cnv-qe-bot1:

PR #311, opened 2025-02-20
must-gather collection can hang on a cluster , so add timeout for run_command() call

dbasunag:
##### Short description:
subprocess.run would wait indefinitely when timeout=None. We should explicitly set timeout for must-gather calls.

### DIFF
--- a/utilities/must_gather.py
+++ b/utilities/must_gather.py
@@ -3,12 +3,18 @@ import shlex
 
 from pyhelper_utils.shell import run_command
 
-from utilities.constants import TIMEOUT_15MIN
+from utilities.constants import TIMEOUT_15MIN, TIMEOUT_20MIN
 
 
 def run_must_gather(
-    image_url="", target_base_dir="", script_name="", flag_names="", timeout=f"{TIMEOUT_15MIN}s", since=None
-):
+    image_url: str = "",
+    target_base_dir: str = "",
+    script_name: str = "",
+    flag_names: str = "",
+    timeout: str = f"{TIMEOUT_15MIN}s",
+    since: str | None = None,
+    run_command_timeout: int = TIMEOUT_20MIN,
+) -> str:
     """
     Run must gather command with an option to create target directory.
 
@@ -24,6 +30,8 @@ def run_must_gather(
             must-gather. However, flag_names can not be passed without script_name
         since (str, optional): since when the data should be collected. format is: '(+|-)[0-9]+(s|m|h|d)'
         timeout (str, optional): runs the debug pods for specified duration
+        run_command_timeout(str, optional): timeout value for run_command call
+
     Returns:
         str: command output
     """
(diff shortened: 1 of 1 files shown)


cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-311 published

utilities/must_gather.py
```diff
@@ -3,12 +3,18 @@
...
 def run_must_gather(
-    image_url="", target_base_dir="", script_name="", flag_names="", timeout=f"{TIMEOUT_15MIN}s", since=None
-):
+    image_url: str = "",
+    target_base_dir: str = "",
+    script_name: str = "",
+    flag_names: str = "",
+    timeout: str = f"{TIMEOUT_15MIN}s",
+    since: str | None = None,
+    run_command_timeout: int = TIMEOUT_20MIN,
```
dshchedr, line 16: just a nit: do you think we need to have it as argument? not sure if someone is going to pass it...

dshchedr (changes requested):

utilities/must_gather.py
dbasunag: I think so. My guess it, for OCP gather you may need to adjust it at some point.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-311 published

utilities/must_gather.py
```diff
@@ -3,12 +3,17 @@
...
 
 
 def run_must_gather(
-    image_url="", target_base_dir="", script_name="", flag_names="", timeout=f"{TIMEOUT_15MIN}s", since=None
-):
+    image_url: str = "",
+    target_base_dir: str = "",
+    script_name: str = "",
+    flag_names: str = "",
+    timeout: str = f"{TIMEOUT_15MIN}s",
```
hmeir, line 14: should be str | None?

hmeir (changes requested):

utilities/must_gather.py
```diff
@@ -42,7 +48,7 @@ def run_must_gather(
     if flag_names:
         flag_string = "".join([f" --{flag_name}" for flag_name in flag_names.split(",")])
         base_command += f" {flag_string}"
-    return run_command(command=shlex.split(base_command), check=False)[1]
+    return run_command(command=shlex.split(base_command), check=False, timeout=TIMEOUT_20MIN)[1]
```
vsibirsk, line 51: `timeout=timeout`?

vsibirsk (changes requested):

vsibirsk (approved):

utilities/must_gather.py
Ahmad-Hafe: I think its better to keep it as 
timeout: str = f"{TIMEOUT_15MIN}s",

this PR is about subprocess.run would wait indefinitely when timeout=None
so we dont want timeout=None, and declareing str | None can be falsy

Ahmad-Hafe (approved):

utilities/must_gather.py
dbasunag: This is default value. So no `None`

cnv-qe-bot1:

PR #364, opened 2025-02-26
remove skip fixtures from some network tests

dbasunag:

### DIFF
--- a/tests/network/ovs_bond/test_bond_link_down.py
+++ b/tests/network/ovs_bond/test_bond_link_down.py
@@ -9,11 +9,8 @@ import pytest
 import utilities.network
 
 
-@pytest.mark.gating
 @pytest.mark.polarion("CNV-3296")
 def test_connectivity_over_pod_network(
-    skip_when_no_bond,
-    skip_if_workers_bms,
     disconnected_bond_port,
     running_ovs_bond_vma,
     running_ovs_bond_vmb,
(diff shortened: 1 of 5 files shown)


vsibirsk (approved):

jpeimer (dismissed):

dshchedr (approved):

Ahmad-Hafe (approved):

dshchedr (approved):

dshchedr (dismissed):

tests/network/ovs_bond/test_bond_link_down.py
```diff
@@ -9,11 +9,8 @@
 import utilities.network
 
 
-@pytest.mark.gating
 @pytest.mark.polarion("CNV-3296")
 def test_connectivity_over_pod_network(
-    skip_when_no_bond,
-    skip_if_workers_bms,
```
hmeir, line 16: This fixture itself can be removed from tests/conftest

hmeir (changes requested):

yossisegev:
> I have concerns with this change. I think that bond tests should have a dedicated marker. The reason is that having multi NICs is not enough. Multi NIC node could mean that the node has only one extra interface, which is enough for all the standard bridge tests, but not for bond configuration.

Following an offline discussion, we found that the bond requirement is already [covered in the sanity check that verifies multi-NIC support](https://github.com/RedHatQE/openshift-virtualization-tests/blob/0ce1686650a4ba6cc91adb9660d92739522377a3/tests/network/conftest.py#L209).

yossisegev (approved):

vsibirsk (approved):

redhat-cnv-qe-bot1:

PR #385, opened 2025-03-02
Fix upgrade_custom collection

hmeir:
##### Short description:
Fixing the upgrade_custom collection.
In addition adding upgrade-custom runs to tox.

Mypy requires line 386, because of: 
```
    ocp_upgrade_test = None
    cnv_upgrade_test_with_prod_src = None
    cnv_upgrade_test_no_prod_src = None
    eus_upgrade_test = None
```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -354,10 +354,10 @@ def remove_upgrade_tests_based_on_config(
         cnv_source(str): cnv source option.
         upgrade_tests(list): list of upgrade tests.
     """
-    ocp_upgrade_test: Item
-    cnv_upgrade_test_with_prod_src: Item
-    cnv_upgrade_test_no_prod_src: Item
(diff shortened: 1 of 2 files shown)


vsibirsk (approved):

conftest.py
```diff
@@ -354,10 +354,10 @@ def remove_upgrade_tests_based_on_config(
         cnv_source(str): cnv source option.
         upgrade_tests(list): list of upgrade tests.
     """
```
jpeimer, line 356: lgtm
(Is it intentionally in 4.18 branch?)

dbasunag: +1 to what @jpeimer said. PRs should be started with main, unless unblocking a release. 4.18 is already out. So this PR should be onhold against 4.18 and we should start with main.

PR #395, opened 2025-03-04
Remove obsolete cpu models: w/a for jira CNV-57387

dshchedr:
Remove obsolete cpu models: w/a for jira CNV-57387

### DIFF
--- a/utilities/infra.py
+++ b/utilities/infra.py
@@ -1426,6 +1426,18 @@ def get_nodes_cpu_model(nodes):
 
     :return: Dict of nodes and associated cpu models
     """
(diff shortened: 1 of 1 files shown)


utilities/infra.py
```diff
@@ -1426,6 +1426,18 @@ def get_nodes_cpu_model(nodes):
...
     """
+    if is_jira_open(jira_id="CNV-57387"):
+        obsolete_cpu = [
+            "core",
+            "kvm",
+            "n270",
+            "pentium",
+            "qemu",
+            "486",
+        ]
```
geetikakay, line 1437: i see Conroe and some other models too as part of obsolete cpu

PR #397, opened 2025-03-04
Test vmi_migration_end_time and vmi_migration_start_time

OhadRevah:
##### Short description:
Test the metrics
kubevirt_vmi_migration_end_time_seconds
kubevirt_vmi_migration_start_time_seconds

### DIFF
--- a/tests/observability/metrics/test_migration_metrics.py
+++ b/tests/observability/metrics/test_migration_metrics.py
@@ -12,8 +12,14 @@ from tests.observability.metrics.constants import (
     KUBEVIRT_VMI_MIGRATION_DIRTY_MEMORY_RATE_BYTES,
     KUBEVIRT_VMI_MIGRATION_DISK_TRANSFER_RATE_BYTES,
 )
-from tests.observability.metrics.utils import get_metric_sum_value, wait_for_non_empty_metrics_value
-from utilities.constants import MIGRATION_POLICY_VM_LABEL, TIMEOUT_2MIN, TIMEOUT_3MIN
+from tests.observability.metrics.utils import (
+    get_metric_sum_value,
+    timestamp_to_seconds,
+    validate_metric_value_within_range,
+    wait_for_non_empty_metrics_value,
+)
+from tests.observability.utils import validate_metrics_value
+from utilities.constants import MIGRATION_POLICY_VM_LABEL, TIMEOUT_2MIN, TIMEOUT_3MIN, TIMEOUT_6MIN
 from utilities.infra import get_node_selector_dict, get_pods
 from utilities.virt import VirtualMachineForTests, fedora_vm_body, running_vm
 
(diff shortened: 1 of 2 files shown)


tests/observability/metrics/test_migration_metrics.py
```diff
@@ -252,3 +277,40 @@ def test_kubevirt_vmi_migration_data_processed_bytes(
...
+class TestKubevirtVmiMigrationStartAndEnd:
+    @pytest.mark.polarion("CNV-11809")
+    def test_metric_kubevirt_vmi_migration_start_time_seconds(
+        self,
+        prometheus,
+        migration_policy_with_bandwidth_scope_class,
+        vm_for_migration_metrics_test,
+        vm_migration_metrics_vmim_scope_class,
+    ):
+        validate_metric_value_within_range(
```
dbasunag, line 291: Why do you need a range here? Are we seeing difference in value between prometheus and what vm_for_migration_metrics_test.vmi.instance.status.migrationState.startTimestamp is reporting?

dbasunag (changes requested):

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -264,3 +276,42 @@ def test_kubevirt_vmi_migration_metrics(
...
+                    timestamp=vm_for_migration_metrics_test.vmi.instance.status.migrationState.startTimestamp
+                ),
+            ),
+        )
+
+    @pytest.mark.polarion("CNV-11810")
+    def test_metric_kubevirt_vmi_migration_end_time_seconds(
+        self,
+        prometheus,
+        migration_policy_with_bandwidth_scope_class,
```
dshchedr, line 304: this migration policy has `bandwidth_per_migration="128Ki"` - it is too slow and it should be used in cases when we don't need to wait for migration to complete.  I'm pretty sure your test may be very flaky (even 5 minute timeout could be not enough for migration to succeed). 
I would suggest to use some bigger value for bandwidth, at least starting from Mi

dshchedr (changes requested):

tests/observability/metrics/test_migration_metrics.py
OhadRevah: The migration here takes about a minute, I did the difference between the timestamps of the start and end of the migration.
I need it to be at least 30 seconds so the prometheus will be able to catch it.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-397 published

tests/observability/metrics/test_migration_metrics.py
dshchedr: I will leave it up to you.. 
just keep in mind - in some cases (on different environments in different situations) it may be a flaky test

dshchedr (approved):

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-397 published

dbasunag (approved):

PR #287, opened 2025-02-13
remove dpdk, jumbo frame, service mesh related skips

dbasunag:

PR #379, opened 2025-02-27
[virt] drop cluster related programmatic skips

vsibirsk:
##### Short description:
  - added special_infra marker for all tests that require fully capable cluster to run
  - added specific feature-related markers
  - added virt_special_infra_sanity

##### Special notes for reviewer:
This PR doesn't handle CPU and storage related skips. Those will be addressed in a follow up PR.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-55884
https://issues.redhat.com/browse/CNV-55883

PR #408, opened 2025-03-05
Add DPDK validation to network sanity check

sbahar619:
##### Short description:
Add DPDK validation to network sanity check
##### More details:
Add DPDK validation to network sanity check
    
    The network_sanity fixture was updated to include validation for
    DPDK support. If any test is marked with the dpdk marker, it checks
    whether the cluster supports running DPDK tests by verifying the
    existence of the dpdk performance profile. If the profile does not
    exist, the test execution is halted with an appropriate error message.
    The logging was updated to reflect the DPDK validation status.

This is a follow up PR after: #287 , which I used the same logic as it was for skipping, but now for assertion.

### DIFF
--- a/tests/network/conftest.py
+++ b/tests/network/conftest.py
@@ -10,6 +10,7 @@ import pytest
 from kubernetes.dynamic.exceptions import ResourceNotFoundError
 from ocp_resources.namespace import Namespace
 from ocp_resources.network_config_openshift_io import Network
+from ocp_resources.performance_profile import PerformanceProfile
 from ocp_resources.pod import Pod
 from pytest_testconfig import config as py_config
 
@@ -199,7 +200,7 @@ def ovn_kubernetes_cluster(admin_client):
 
 
 @pytest.fixture(scope="session", autouse=True)
-def network_sanity(hosts_common_available_ports, junitxml_plugin):
+def network_sanity(hosts_common_available_ports, junitxml_plugin, request):
     """
     Perform verification that the cluster is a multi-nic one otherwise exit run
     """
@@ -214,3 +215,16 @@ def network_sanity(hosts_common_available_ports, junitxml_plugin):
             junitxml_property=junitxml_plugin,
         )
     LOGGER.info(f"Validated network lane is running against a multinic-cluster: {hosts_common_available_ports}")
+
+    if any(item.get_closest_marker("dpdk") for item in request.session.items):
+        LOGGER.info("Verifying if the cluster supports running DPDK tests...")
+        dpdk_performance_profile_name = "dpdk"
+        if not PerformanceProfile(name=dpdk_performance_profile_name).exists:
+            exit_pytest_execution(
+                filename="network_cluster_sanity_failure.txt",
+                return_code=network_sanity_failure_return_code,
+                message=f"DPDK is not configured, the {PerformanceProfile.kind}/{dpdk_performance_profile_name} "
+                "does not exist.",
+                junitxml_property=junitxml_plugin,
+            )
+        LOGGER.info("Validated network lane is running against a DPDK-enabled cluster")


cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-408 published

tests/network/checkup_framework/test_checkup_dpdk.py
```diff
@@ -13,10 +14,18 @@
         "dpdk_checkup_traffic_generator_service_account",
         "dpdk_checkup_configmap_role_binding",
         "dpdk_checkup_resources_role_binding",
+        "check_when_no_dpdk",
```
yossisegev, line 17: Under the assumption that these fixtures are called in the order they appear, I think it should be the first fixture in the list. If DPDK is not installed, there's no point in executing all the other fixtures.

yossisegev (changes requested):

tests/network/checkup_framework/test_checkup_dpdk.py
EdDev: I also agree.

However, the usage of `usefixture` here is redundant when the content ref fixtures in the same module.
It should be enough (and cleaner) to just pass on the relevant fixtures a `autouse=True`.

It makes sense to define it first so it will be processed first (hopefully that is the case).

tests/network/checkup_framework/test_checkup_dpdk.py
```diff
@@ -13,10 +14,18 @@
...
     ),
 ]
 
 
+@pytest.fixture(scope="session")
+def check_when_no_dpdk(admin_client):
```
EdDev, line 23: I would called it `dpdk_support` or `dpdk_support_validation`. Fixtures are not suppose to be verbs.

yossisegev: > It should be enough (and cleaner) to just pass on the relevant fixtures a autouse=True.

@EdDev I am not sure it is necessarily clearer the way you are suggesting.
There's an advantage in gathering all the auto-used fixtures in one central place.
Bottom line - IMO it's more a matter of personal preference, and not a matter of right-vs.-wrong.

tests/network/checkup_framework/test_checkup_dpdk.py
sbahar619: Done, I reorder the fixures.

tests/network/checkup_framework/test_checkup_dpdk.py
sbahar619: @EdDev 
Done, I renamed the fixture.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/53ef7215380eb7bd6d17b99e5661089c8156f317..8fd15a39af45d4f89c0e705d8ce57acd0ebf4a9e
Resolve @EdDev and @yossisegev comments,

tests/network/checkup_framework/test_checkup_dpdk.py
sbahar619: @EdDev 
I think using autouse=True for session scoped fixture means the fixture will be applied to all the tests in the suite, which we don't want to do.

EdDev: > There's an advantage in gathering all the auto-used fixtures in one central place.

This implies that I need to define a fixture and then also ref it from some place.
That is 1) Less elegant and 2) contradicts the ability to distribute fixture all over the project (vs centralizing it).
With the same reasoning, you could argue that we should "show" and explicitly ref fixtures at all the package levels above, including root, just to see all fixtures that run. I'm less fun of that. 

> I think using autouse=True for session scoped fixture means the fixture will be applied to all the tests in the suite, which we don't want to do.

If that is the case, then I would expect you to change it to module level.
Setting a session scope under a package or module makes little sense. But if has some usage, I will love to learn.

This is not a blocker for me, if anything, it should be something to consider for any new code added.

EdDev (approved):
Thanks!

yossisegev (approved):

tests/network/checkup_framework/test_checkup_dpdk.py
```diff
@@ -17,6 +19,13 @@
 ]
 
 
+@pytest.fixture(scope="session")
+def dpdk_support_validation(admin_client):
+    assert any(profile.name == "dpdk" for profile in list(PerformanceProfile.get(dyn_client=admin_client))), (
+        "DPDK is not configured"
+    )
```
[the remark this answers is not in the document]
sbahar619: Done.

sbahar619:
Change: 
https://github.com/RedHatQE/openshift-virtualization-tests/compare/8fd15a39af45d4f89c0e705d8ce57acd0ebf4a9e..d1d6e4191a305595cb13390ae2360f284281d890
Resolve @vsibirsk comment.

sbahar619:
Change: rebase.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-408 published

tests/network/checkup_framework/test_checkup_dpdk.py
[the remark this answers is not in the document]
EdDev: Nice one!

tests/network/checkup_framework/test_checkup_dpdk.py
```diff
@@ -17,6 +19,11 @@
 ]
 
 
+@pytest.fixture(scope="session")
+def dpdk_support_validation(admin_client):
```
EdDev, line 23: I guess you do not need the `admin_client` now, right?

vsibirsk: +1

tests/network/checkup_framework/test_checkup_dpdk.py
```diff
@@ -17,6 +19,11 @@
...
+    assert PerformanceProfile(name="dpdk").exists, "DPDK is not configured"
```
vsibirsk, line 24: on second thought i don't think assert is good call here,
IMHO
assertions are part of the test (check no bugs and the feature is working)
pytest.fail is for an undesired state (operator not deployed, cluster of wrong type, etc)

vsibirsk (changes requested):

tests/network/checkup_framework/test_checkup_dpdk.py
dbasunag: This should be part of network_sanity fixture. There is no point running T3, if dpdk is not present. So if tier3 marker was passed in command line this should be checked.

dbasunag (changes requested):

tests/network/checkup_framework/test_checkup_dpdk.py
sbahar619: Right, thanks, done.

tests/network/checkup_framework/test_checkup_dpdk.py
EdDev: I am not familiar with such a distinction between asserting and explicit call to `fail`.
Can you please ref something that describes/explains/recommends it?

vsibirsk: i don't think there are hard guidelines about it (thats why it's an IMHO :) )

my view is based on pytest docs:
https://docs.pytest.org/en/stable/explanation/anatomy.html#test-anatomy
https://docs.pytest.org/en/stable/explanation/fixtures.html#what-fixtures-are
my point is that fixture is an arrange/act phase, that's why the distinction - `fail` for fixtures, `assert` for check/validate test functions
in other words - fixture for

EdDev: This is surely not what happens in practice.
I mean, an `assert` in a fixture will raise ERROR in case of failure and a FAIL in case it runs and fails in a test body.

We have this all over the codebase.

But fine, it is opinionated and we need this in, lets move things on and accept a condition + fail() to get this in.

dbasunag: +1 to what Vasiliy said. 
Also, please do this as a part of network sanity. T3 network should not even begin execution if dpdk is not installed. It should be similar to what virt team did here: 
https://github.com/RedHatQE/openshift-virtualization-tests/pull/379/files#diff-b1be3011e49a9ad9c396ea72c9895f6fa0e7b23ee1ed84204ac58c4398e95f01R15

dbasunag (changes requested):

tests/network/checkup_framework/test_checkup_dpdk.py
sbahar619: Sure @vsibirsk , I will apply your comment.

tests/network/checkup_framework/test_checkup_dpdk.py
```diff
@@ -17,6 +19,12 @@
 ]
 
 
+@pytest.fixture(scope="session")
```
dbasunag, line 22: please do this as a part of network sanity. T3 network should not even begin execution if dpdk is not installed. It should be similar to what virt team did here:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/379/files#diff-b1be3011e49a9ad9c396ea72c9895f6fa0e7b23ee1ed84204ac58c4398e95f01R15

dbasunag (changes requested):

tests/network/checkup_framework/test_checkup_dpdk.py
EdDev: Can you explain what we gain by this?
As it is not, it says something like this: If any test under this module is executed, we will fail on this fixture.
That sounds ok to me. If tier-2 tests are executed, tests here will not run and therefore the fixture will not run.

sbahar619:
Change:
https://github.com/RedHatQE/openshift-virtualization-tests/compare/8e4b392a1b38481b2e2dd28606e8425b32a1f978..e96c2486cb2c345ecdb08a1e8551d3c663e9114f
Resolve @vsibirsk comment.

tests/network/checkup_framework/test_checkup_dpdk.py
sbahar619: Added logic to fail DPDK tests if they are being collected or if the user explicitly sets the dpdk marker. The tests will now trigger validation if the marker is found in the collected tests or specified via the -m flag.

tests/network/conftest.py
```diff
@@ -199,7 +199,7 @@ def ovn_kubernetes_cluster(admin_client):
 
 
 @pytest.fixture(scope="session", autouse=True)
-def network_sanity(hosts_common_available_ports, junitxml_plugin):
+def network_sanity(hosts_common_available_ports, junitxml_plugin, pytestconfig, request):
     """
     Perform verification that the cluster is a multi-nic one otherwise exit run
```
Anatw, line 204: This need to be changed in line with the DPDK addition.

dbasunag: Why? dpdk is part of network sanity, no?

tests/network/conftest.py
```diff
@@ -214,3 +214,8 @@ def network_sanity(hosts_common_available_ports, junitxml_plugin):
             junitxml_property=junitxml_plugin,
         )
     LOGGER.info(f"Validated network lane is running against a multinic-cluster: {hosts_common_available_ports}")
+
+    if "dpdk" in pytestconfig.getoption("-m") or any(item.get_closest_marker("dpdk") for item in request.session.items):
```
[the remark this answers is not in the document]
sbahar619: Done.

sbahar619:
Change:  https://github.com/RedHatQE/openshift-virtualization-tests/compare/08c8cac99047522a609b50b87cb872016f2bc678..d577d388c377513f279e7de37c61de209d8a2660
Resolve @EdDev comment.

dbasunag (approved):

tests/network/conftest.py
EdDev: The text mentions "multi-nic", so I guess Anat mentioned it should also mention "dpdk" now.

IMO we should drop the specifics, i.e. remove the "multi-nic" text.

Anyway, it is a nit.

EdDev (approved):
One comment about the function doctext is not treated, but it is a nit which we can follow up with later.

yossisegev (approved):

tests/network/conftest.py
dbasunag: If it is about updating the comment, may be in a follow up PR, remove it or update it. I won't hold the PR for that, but it is your team's decision.

tests/network/utils.py
```diff
@@ -423,3 +425,8 @@ def get_service(name, namespace):
         return service
 
     raise ResourceNotFoundError(f"Service {name}.")
+
+
+def dpdk_support_validation():
+    if not PerformanceProfile(name="dpdk").exists:
+        pytest.fail(reason="DPDK is not configured")
```
dbasunag, line 432: Ok, I just noticed this. You need to fail here. using exit_pytest_execution()

dbasunag (changes requested):

vsibirsk (changes requested):

tests/network/conftest.py
```diff
@@ -214,3 +214,8 @@ def network_sanity(hosts_common_available_ports, junitxml_plugin):
             junitxml_property=junitxml_plugin,
         )
     LOGGER.info(f"Validated network lane is running against a multinic-cluster: {hosts_common_available_ports}")
+
+    if any(item.get_closest_marker("dpdk") for item in request.session.items):
```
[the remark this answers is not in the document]
EdDev: I asked to remove it: https://github.com/RedHatQE/openshift-virtualization-tests/pull/408#discussion_r1983303455

tests/network/conftest.py
```diff
@@ -199,7 +199,7 @@ def ovn_kubernetes_cluster(admin_client):
 
 
 @pytest.fixture(scope="session", autouse=True)
-def network_sanity(hosts_common_available_ports, junitxml_plugin):
+def network_sanity(hosts_common_available_ports, junitxml_plugin, pytestconfig, request):
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/conftest.py
sbahar619: I will handle it in a follow up PR.

tests/network/conftest.py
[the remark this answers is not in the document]
sbahar619: @vsibirsk 
The rationale for removing this logic is that we’re currently querying all collected tests, making the check for whether the user explicitly set the flag redundant.

tests/network/conftest.py
```diff
@@ -214,3 +214,8 @@ def network_sanity(hosts_common_available_ports, junitxml_plugin):
             junitxml_property=junitxml_plugin,
         )
     LOGGER.info(f"Validated network lane is running against a multinic-cluster: {hosts_common_available_ports}")
+
+    if any(item.get_closest_marker("dpdk") for item in request.session.items):
+        LOGGER.info("Verifying if the cluster supports running DPDK tests...")
+        dpdk_support_validation()
+        LOGGER.info("DPDK support validation completed successfully.")
```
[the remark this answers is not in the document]
sbahar619: Done, I removed the function call and moved the logic directly into the sanity fixture.

tests/network/utils.py
sbahar619: Done, thanks.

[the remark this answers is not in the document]
sbahar619: Good idea, I will handle it in a follow up PR.

sbahar619:
Change: 
https://github.com/RedHatQE/openshift-virtualization-tests/compare/d577d388c377513f279e7de37c61de209d8a2660..ea40522416eed1fafd12ad9c796721e77a29e091
Resolve @vsibirsk and @dbasunag comments.

sbahar619:
Change: rebase.

sbahar619:
Change: update commit message.

tests/network/conftest.py
Anatw: Thank you.

Anatw (approved):
Thank you

yossisegev (approved):

EdDev (approved):
Thanks!

vsibirsk (approved):
Just so it's not lost in the conversations, please don't forget to do these things in follow up PR:
 - update doc str in `network_sanity` fixture (that it also validated dpdk now)
 - refactor `network_sanity` so that it runs all checks and then fails with listing all failed items

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-408 published

dshchedr (approved):

dbasunag (approved):

redhat-cnv-qe-bot1:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-408.

redhat-cnv-qe-bot1:

redhat-cnv-qe-bot1:
Cherry-picked PR Add DPDK validation to network sanity check into cnv-4.18

PR #405, opened 2025-03-04
change ipv4/ipv6 matrix related skips to fail

dbasunag:

PR #410, opened 2025-03-05
Add fixture to check if Service Mesh is installed

sbahar619:
##### Short description:
Add fixture to check if Service Mesh is installed
##### More details:
Introduced a new pytest fixture to verify that the Service Mesh is installed before running tests.
The fixture is used in service_mesh_member_roll to ensure that the tests will not run if the Service Mesh is not present.
This is a follow up PR after: https://github.com/RedHatQE/openshift-virtualization-tests/pull/287 , which I used the same logic as it was for skipping, but now for assertion.

### DIFF
--- a/tests/network/conftest.py
+++ b/tests/network/conftest.py
@@ -100,6 +100,12 @@ def istio_system_namespace(admin_client):
     return Namespace(name=ISTIO_SYSTEM_DEFAULT_NS, client=admin_client).exists
 
 
+@pytest.fixture(scope="session")
+def service_mesh_installed_validation(istio_system_namespace):
+    if not istio_system_namespace:
+        pytest.fail(reason="Service Mesh is not installed.")
+
+
 @pytest.fixture(scope="module")
 def sriov_workers_node1(sriov_workers):
     """
--- a/tests/network/service_mesh/test_service_mesh.py
+++ b/tests/network/service_mesh/test_service_mesh.py
@@ -6,7 +6,7 @@ from tests.network.service_mesh.utils import (
 )
 from tests.network.utils import assert_authentication_request
 
-pytestmark = pytest.mark.special_infra
+pytestmark = [pytest.mark.usefixtures("service_mesh_installed_validation"), pytest.mark.special_infra]
 
 
 class TestSMTrafficManagement:


cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-410 published

tests/network/service_mesh/conftest.py
```diff
@@ -215,7 +215,7 @@ def httpbin_service_service_mesh(httpbin_deployment_service_mesh, httpbin_servic
 
 
 @pytest.fixture(scope="module")
-def service_mesh_member_roll(namespace):
+def service_mesh_member_roll(namespace, check_if_service_mesh_not_installed):
```
yossisegev, line 218: The fixture should be called independently (from the test module), and not via another fixture - either via `usefixtures` (which would be my preferred option) or `autouse=True`.
When getting to the ServiceMesh package/module, we should first check if it is installed, and continue to the next package if it's not.

yossisegev (changes requested):

EdDev (changes requested):

tests/network/service_mesh/conftest.py
sbahar619: Agree, I already did it and forgot to push.
About the scope, it should be session because we don't need to call the API to check the existence of the SM NS for each module, even though we only have one now,

sbahar619: Done.

tests/network/conftest.py
```diff
@@ -100,6 +100,11 @@ def istio_system_namespace(admin_client):
     return Namespace(name=ISTIO_SYSTEM_DEFAULT_NS, client=admin_client).exists
 
 
+@pytest.fixture(scope="session")
+def service_mesh_installed_validation(istio_system_namespace):
+    assert istio_system_namespace, "Service Mesh is not installed."
```
hmeir, line 105: I think that according to https://github.com/RedHatQE/openshift-virtualization-tests/pull/405#discussion_r1981303387
We should `pytest.fail`

tests/network/conftest.py
```diff
@@ -100,6 +100,11 @@ def istio_system_namespace(admin_client):
     return Namespace(name=ISTIO_SYSTEM_DEFAULT_NS, client=admin_client).exists
 
 
+@pytest.fixture(scope="session")
```
dbasunag, line 103: https://github.com/RedHatQE/openshift-virtualization-tests/pull/379/files#diff-b1be3011e49a9ad9c396ea72c9895f6fa0e7b23ee1ed84204ac58c4398e95f01R15

Please add this to network_sanity, like virt team did. Network special infra tests should not even begin running, if the cluster does not have everything it needs.

dbasunag (changes requested):

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/43247383b4c6f7bd2479887688eb9d7ab6ca93b7..c7badfa95134a0a076bc46e210c2ab2c6fc6a4df
Resolve @vsibirsk comment.

EdDev (approved):
I'm good with this approach.

I understand that the goal is to fail even before a single test got executed in a run.
I believe that is achievable and we will send something to handle it in an elegant manner.
For now, I think it is ok to fail in the network tests.
If anything, it is a sig-network burden to resolve quickly.
It is also better than what we have now, where we will fail and it will not be easy to understand why.

Please consider taking in this one with the promise to follow up with a better solution later.

tests/network/conftest.py
EdDev: I will try to do solve it in a different manner with the help of the test collector and a generic sanity validator.
In the meantime, I suggest to let this in as it does solve the problem of debugging why something failed due to a missing cluster pre-requirement.

sbahar619:
I will open new PR, after related PR #408 comments indicating this PR will need its core changes and rebasing.

cnv-qe-bot1:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-410.

PR #414, opened 2025-03-05
Test kubevirt_vmi_migration_data_total_bytes

OhadRevah:
##### Short description:
Test for kubevirt_vmi_migration_data_total_bytes metric.

### DIFF
--- a/tests/observability/metrics/test_migration_metrics.py
+++ b/tests/observability/metrics/test_migration_metrics.py
@@ -10,7 +10,7 @@ from tests.observability.metrics.constants import (
     KUBEVIRT_VMI_MIGRATION_DATA_PROCESSED_BYTES,
     KUBEVIRT_VMI_MIGRATION_DATA_REMAINING_BYTES,
     KUBEVIRT_VMI_MIGRATION_DIRTY_MEMORY_RATE_BYTES,
-    KUBEVIRT_VMI_MIGRATION_DISK_TRANSFER_RATE_BYTES,
+    KUBEVIRT_VMI_MIGRATION_DISK_TRANSFER_RATE_BYTES, KUBEVIRT_VMI_MIGRATION_DATA_TOTAL_BYTES,
 )
 from tests.observability.metrics.utils import get_metric_sum_value, wait_for_non_empty_metrics_value
 from utilities.constants import MIGRATION_POLICY_VM_LABEL, TIMEOUT_2MIN, TIMEOUT_3MIN
@@ -247,6 +247,10 @@ class TestKubevirtVmiMigrationMetrics:
             pytest.param(
                 KUBEVIRT_VMI_MIGRATION_DIRTY_MEMORY_RATE_BYTES,
                 marks=(pytest.mark.polarion("CNV-11599")),
+            ),
+            pytest.param(
+                KUBEVIRT_VMI_MIGRATION_DATA_TOTAL_BYTES,
+                marks=(pytest.mark.polarion("CNV-11802")),
             ),
         ],
     )
--- a/tests/observability/metrics/constants.py
+++ b/tests/observability/metrics/constants.py
@@ -80,3 +80,4 @@ KUBEVIRT_VMI_MIGRATION_DIRTY_MEMORY_RATE_BYTES = "kubevirt_vmi_migration_dirty_m
 KUBEVIRT_VMSNAPSHOT_PERSISTENTVOLUMECLAIM_LABELS = (
     "kubevirt_vmsnapshot_persistentvolumeclaim_labels{{vm_name='{vm_name}'}}"
 )
+KUBEVIRT_VMI_MIGRATION_DATA_TOTAL_BYTES = "kubevirt_vmi_migration_data_total_bytes{{name='{vm_name}'}}"


cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-414 published

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -248,6 +249,10 @@ class TestKubevirtVmiMigrationMetrics:
                 KUBEVIRT_VMI_MIGRATION_DIRTY_MEMORY_RATE_BYTES,
                 marks=(pytest.mark.polarion("CNV-11599")),
             ),
+            pytest.param(
+                KUBEVIRT_VMI_MIGRATION_DATA_TOTAL_BYTES,
+                marks=(pytest.mark.polarion("CNV-11802")),
+            ),
```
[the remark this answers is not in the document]
OhadRevah: In some cases like this one yes, because the values always different, we just can see that it gets non zero value and it's ok.

dbasunag (approved):

hmeir (approved):

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-414 published

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-414 published

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -68,10 +79,18 @@ def assert_metrics_values(prometheus, migration_metrics_dict, initial_values, me
         initial_value = initial_values[metric]
         expected_value = initial_value + 1 if metric == metric_to_check else initial_value
         try:
+            time_passed_from_migration_start = None
+            if vmi:
+                time_passed_from_migration_start = time_passed_from_timestamp_until_now_minutes(
+                    timestamp=vmi.instance.status.migrationState.startTimestamp
+                )
```
dshchedr, line 86: With this change the function logic looks more complicated.. and even doc string does not reflect what is checked here.

Instead of obfustating current function by optional argument `vmi` - I would suggest to create additional function for `time_passed_from_migration_start` verification.

dshchedr (changes requested):

tests/observability/metrics/test_migration_metrics.py
OhadRevah: I will do it in other PR because it is not related to this test, removed it.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-414 published

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -260,7 +269,11 @@ def test_kubevirt_vmi_migration_metrics(
...
         query,
     ):
+        time_passed_from_starting_migration = time_passed_from_timestamp_until_now_minutes(
+            timestamp=vm_for_migration_metrics_test.instance.status.migrationState.startTimestamp
+        )
         wait_for_non_empty_metrics_value(
             prometheus=prometheus,
-            metric_name=f"last_over_time({query.format(vm_name=vm_for_migration_metrics_test.name)}[8m])",
+            metric_name=f"last_over_time({query.format(vm_name=vm_for_migration_metrics_test.name)}"
+            f"[{time_passed_from_starting_migration if time_passed_from_starting_migration != 0 else 8}m])",
```
dbasunag, line 278: What would be the point of using [0m]?

dbasunag (changes requested):

tests/observability/metrics/test_migration_metrics.py
OhadRevah: It will not use [0m], because if `time_passed_from_starting_migration ` is zero than it will be [8m].

jpeimer: Why is it 8 minutes?

OhadRevah: @jpeimer Because the default is 5m so do last_over_time is not relevant if its less or equal to 5, so I gave it extra 3 mins.

tests/observability/metrics/utils.py
```diff
@@ -1132,6 +1132,13 @@ def timestamp_to_seconds(timestamp: str) -> int:
     return int(dt.timestamp())
 
 
+def time_passed_from_timestamp_until_now_minutes(timestamp: str) -> int:
+    time_passed = (int(datetime.now(timezone.utc).timestamp()) - timestamp_to_seconds(timestamp=timestamp)) // 60
+    if time_passed > 8:
+        return time_passed
+    return 0
```
vsibirsk, line 1139: why 0 if not > 8?

vsibirsk (changes requested):

tests/observability/metrics/utils.py
OhadRevah: Because 5 mins is the default when running prometheus so I gave it extra 3, so if it is more than this is the time I will use in last_over_time.

vsibirsk: then just return `time_passed` and do the check > 8 where it's needed.

tests/observability/metrics/test_migration_metrics.py
vsibirsk: `f"[{time_passed_from_starting_migration if time_passed_from_starting_migration > 8 else 8}m])",`

vsibirsk (changes requested):

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-414 published

tests/observability/metrics/utils.py
```diff
@@ -1132,6 +1132,10 @@ def timestamp_to_seconds(timestamp: str) -> int:
     return int(dt.timestamp())
 
 
+def time_passed_from_timestamp_until_now_minutes(timestamp: str) -> int:
+    return (int(datetime.now(timezone.utc).timestamp()) - timestamp_to_seconds(timestamp=timestamp)) // 60
```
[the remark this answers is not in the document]
OhadRevah: Yes, I am sure because when you wrap it with int it's possible.

jpeimer (dismissed):

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-414 published

tests/observability/metrics/utils.py
```diff
@@ -1134,6 +1134,10 @@ def timestamp_to_seconds(timestamp: str) -> int:
     return int(dt.timestamp())
 
 
+def time_passed_from_timestamp_until_now_minutes(timestamp: str) -> int:
```
hmeir, line 1137: This function is used only once - do we need it?

OhadRevah: Yes because otherwise the code will be less readable

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -272,9 +284,13 @@ def test_kubevirt_vmi_migration_metrics(
         vm_migration_metrics_vmim_scope_class,
         query,
     ):
+        time_passed_from_starting_migration = time_passed_from_timestamp_until_now_minutes(
+            timestamp=vm_for_migration_metrics_test.vmi.instance.status.migrationState.startTimestamp
```
hmeir, line 288: this is used twice below - how about re-using it?

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -47,21 +50,26 @@ def metric_value_sampler(prometheus, metric, expected_value):
...
     Check all migration metrics do not change from initial values,
     except for specified metric which must increase by 1.
 
     Args:
         initial_values: Dictionary representing initial values of metrics
         metric_to_check: metric expected to be increased by 1
-        vm: vm object
-
+        migration_metrics_dict: Dictionary with metrics by the status it checks
+        prometheus: Prometheus object
```
hmeir, line 72: Can you add those before the other two args?
So the order will be the same

hmeir (changes requested):

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -47,21 +50,26 @@ def metric_value_sampler(prometheus, metric, expected_value):
...
             current_check = 0
 
 
-def assert_metrics_values(prometheus, migration_metrics_dict, initial_values, metric_to_check):
+def assert_metrics_values(
+    prometheus: Prometheus,
+    migration_metrics_dict: dict[str, str],
+    initial_values: dict[str, int],
+    metric_to_check: str,
+):
```
rnetser, line 63: missing return type

tests/observability/metrics/utils.py
rnetser: not sure i understand how it will be less readable, just set `time_passed_from_starting_migration` to the below.

rnetser (changes requested):

tests/observability/metrics/utils.py
dshchedr: +1
one line function that is used in one place

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -34,7 +38,7 @@ def delete_failed_migration_target_pod(admin_client, namespace, vm_name):
             pod.delete(wait=True)
 
 
-def metric_value_sampler(prometheus, metric, expected_value):
+def metric_value_sampler(prometheus: Prometheus, metric: str, expected_value: int) -> None:
```
hmeir, line 41: qq,
Why do we need this function when we have `wait_for_expected_metric_value_sum`?

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -273,9 +286,16 @@ def test_kubevirt_vmi_migration_metrics(
         vm_migration_metrics_vmim_scope_class,
         query,
     ):
+        time_passed_from_starting_migration = (
```
hmeir, line 289: since you are dividing by 60 - how about more declerative name?
```suggestion
        minutes_passed_since_migration_start = (
```\

**Review by hmeir — CHANGES_REQUESTED** [review]

**Review by rnetser — DISMISSED** [review]

**Review by rnetser — APPROVED** [review]

File: tests/observability/metrics/test_migration_metrics.py [hunk 12]
```diff
@@ -273,9 +269,16 @@ def test_kubevirt_vmi_migration_metrics(
...
+            int(datetime.now(timezone.utc).timestamp())
+            - timestamp_to_seconds(
+                timestamp=vm_for_migration_metrics_test.vmi.instance.status.migrationState.startTimestamp
+            )
+        ) // 60
         wait_for_non_empty_metrics_value(
             prometheus=prometheus,
-            metric_name=f"last_over_time({query.format(vm_name=vm_for_migration_metrics_test.name)}[8m])",
+            metric_name=f"last_over_time({query.format(vm_name=vm_for_migration_metrics_test.name)}"
+            f"[{minutes_passed_since_migration_start if minutes_passed_since_migration_start > 8 else 8}m])",
```
rnetser, line 281: why set it to 8 minutes?

OhadRevah: Because the metric value validated while the migration is running, so because it tests multiple metrics I want to make sure the test won't fail if the migration is done and at that moment there is no value, and the prometheus default is 5mins so I give it extra 3 to be sure test won't fail for that

hmeir: Maybe it would be better to sample this metric starting from 5/6 min, and if there's no value then try again every minute until you reach 8?

rnetser (changes requested):
D/S test `tox -e verify-bugs-are-open` failed: cnv-tests-tox-executor/11510

tests/observability/metrics/test_migration_metrics.py
rnetser: > prometheus default is 5mins default is 5mins
not for this PR but can this value be reduced in prometheus config?

rnetser: as this does not increase execution time, why not set the value to 10m (and not be at the upper bound of the flakiness you witnessed)?

cnv-qe-bot1 (approved):

cnv-qe-bot1 (approved):

dshchedr (approved):

tests/observability/metrics/test_migration_metrics.py
```diff
@@ -273,9 +269,16 @@ def test_kubevirt_vmi_migration_metrics(
         vm_migration_metrics_vmim_scope_class,
         query,
     ):
+        minutes_passed_since_migration_start = (
+            int(datetime.now(timezone.utc).timestamp())
+            - timestamp_to_seconds(
+                timestamp=vm_for_migration_metrics_test.vmi.instance.status.migrationState.startTimestamp
+            )
+        ) // 60
```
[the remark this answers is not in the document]
OhadRevah: Fixed

cnv-qe-bot1 (approved):

cnv-qe-bot1 (approved):

hmeir (approved):

PR #464, opened 2025-03-13
 Add test for connectivity between localnet VMs

sbahar619:
##### Short description:
Add test for connectivity between localnet VMs
##### More details:
Add test for connectivity between localnet VMs
    
    Introduced a new test to verify connectivity between VMs on the
    localnet network. This test ensures that VMs within the localnet
    network have connectivity.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-55290

### DIFF
--- a/libs/net/netattachdef.py
+++ b/libs/net/netattachdef.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 from dataclasses import asdict, dataclass, field
+from enum import Enum
 from typing import Any, Final
 
 from kubernetes.dynamic import DynamicClient
@@ -28,6 +29,25 @@ class CNIPluginBridgeConfig(CNIPluginConfig):
     vlan: int | None = None
     macspoofchk: bool | None = None
     disableContainerInterface: bool | None = None  # noqa: N815
+
+
+@dataclass
+class CNIPluginOvnK8sConfig(CNIPluginConfig):
+    """
+    CNI OVN-Kubernetes Plugin
+    Ref:
+    https://docs.openshift.com/container-platform/4.14/networking/multiple_networks/
+    configuring-additional-network.html#configuration-ovnk-network-plugin-json-object_
+    configuring-additional-network
+    """
+
+    type: str = field(default="ovn-k8s-cni-overlay", init=False)
+    topology: str
+    netAttachDefName: str  # noqa: N815
+    vlanID: int | None = None  # noqa: N815
+
+    class Topology(Enum):
+        LOCALNET = "localnet"
 
 
 @dataclass
--- a/libs/net/vmspec.py
+++ b/libs/net/vmspec.py
@@ -4,11 +4,12 @@ from typing import Any, Final
 from kubernetes.dynamic.client import ResourceField
 from timeout_sampler import TimeoutExpiredError, retry
 
-from libs.vm.spec import Network
+from libs.vm.spec import Devices, Interface, Network, SpecDisk, VMISpec, Volume
 from libs.vm.vm import BaseVirtualMachine
 
 LOOKUP_IFACE_STATUS_TIMEOUT_SEC: Final[int] = 30
 RETRY_INTERVAL_SEC: Final[int] = 5
+IP_ADDRESS: Final[str] = "ipAddress"
 
 
 class VMInterfaceSpecNotFoundError(Exception):
@@ -38,7 +39,7 @@ def lookup_iface_status(vm: BaseVirtualMachine, iface_name: str) -> ResourceFiel
         return _lookup_iface_status(
             vm=vm,
             iface_name=iface_name,
-            predicate=lambda interface: "guest-agent" in interface["infoSource"] and interface["ipAddress"],
+            predicate=lambda interface: "guest-agent" in interface["infoSource"] and interface[IP_ADDRESS],
         )
     except TimeoutExpiredError:
         raise VMInterfaceStatusNotFoundError(f"Network interface named {iface_name} was not found in VM {vm.name}.")
@@ -78,3 +79,21 @@ def lookup_primary_network(vm: BaseVirtualMachine) -> Network:
         if network.pod is not None:
             return Network(**network)
     raise VMInterfaceSpecNotFoundError(f"No interface connected to the primary network was found in VM {vm.name}.")
+
+
+def add_network_interface(vmi_spec: VMISpec, network: Network, interface: Interface) -> VMISpec:
+    vmi_spec.networks = vmi_spec.networks or []
+    vmi_spec.networks.append(network)
+    vmi_spec.domain.devices = vmi_spec.domain.devices or Devices()
+    vmi_spec.domain.devices.interfaces = vmi_spec.domain.devices.interfaces or []
+    vmi_spec.domain.devices.interfaces.append(interface)
+    return vmi_spec
+
+
+def add_volume_disk(vmi_spec: VMISpec, volume: Volume, disk: SpecDisk) -> VMISpec:
+    vmi_spec.volumes = vmi_spec.volumes or []
+    vmi_spec.volumes.append(volume)
+    vmi_spec.domain.devices = vmi_spec.domain.devices or Devices()
+    vmi_spec.domain.devices.disks = vmi_spec.domain.devices.disks or []
+    vmi_spec.domain.devices.disks.append(disk)
+    return vmi_spec
--- a/libs/vm/factory.py
+++ b/libs/vm/factory.py
@@ -38,8 +38,8 @@ def _fill_vm_spec_defaults(spec: VMSpec | None) -> VMSpec:
     vmi_spec.volumes = vmi_spec.volumes or []
 
     disk, volume = containerdisk_storage(image=fedora_image())
-    vmi_spec.domain.devices.disks.append(disk)
-    vmi_spec.volumes.append(volume)
+    vmi_spec.domain.devices.disks.insert(0, disk)
+    vmi_spec.volumes.insert(0, volume)
 
     vmi_spec.domain.cpu = vmi_spec.domain.cpu or CPU(cores=1)
     vmi_spec.domain.memory = vmi_spec.domain.memory or Memory(guest="1Gi")
--- a/libs/vm/spec.py
+++ b/libs/vm/spec.py
@@ -108,6 +108,7 @@ class PodAntiAffinity:
 class PodAffinityTerm:
     labelSelector: LabelSelector  # noqa: N815
     topologyKey: str  # noqa: N815
+    namespaceSelector: dict[str, Any] | None = None  # noqa: N815
 
 
 @dataclass
@@ -137,3 +138,4 @@ class ContainerDisk:
 @dataclass
 class CloudInitNoCloud:
     networkData: str  # noqa: N815
+    userData: str | None = None  # noqa: N815
--- a/libs/vm/vm.py
+++ b/libs/vm/vm.py
@@ -9,15 +9,9 @@ from ocp_resources.node import Node
 from ocp_resources.virtual_machine import VirtualMachine, VirtualMachineInstance
 from pytest_testconfig import config as py_config
 
-from libs.vm.spec import (
-    ContainerDisk,
-    Disk,
-    SpecDisk,
-    VMSpec,
-    Volume,
-)
+from libs.vm.spec import CloudInitNoCloud, ContainerDisk, Disk, SpecDisk, VMSpec, Volume
 from utilities import infra
-from utilities.virt import get_oc_image_info, vm_console_run_commands
+from utilities.virt import CLOUD_INIT_DISK_NAME, get_oc_image_info, vm_console_run_commands
 
 
 class BaseVirtualMachine(VirtualMachine):
@@ -93,3 +87,9 @@ def container_image(base_image: str) -> str:
 def containerdisk_storage(image: str) -> tuple[SpecDisk, Volume]:
     name = "containerdisk"
     return SpecDisk(name=name, disk=Disk(bus="virtio")), Volume(name=name, containerDisk=ContainerDisk(image=image))
+
+
+def cloudinitdisk_storage(data: CloudInitNoCloud) -> tuple[SpecDisk, Volume]:
+    return SpecDisk(name=CLOUD_INIT_DISK_NAME, disk=Disk(bus="virtio")), Volume(
+        name=CLOUD_INIT_DISK_NAME, cloudInitNoCloud=data
+    )
--- a/tests/network/libs/cloudinit.py
+++ b/tests/network/libs/cloudinit.py
@@ -1,6 +1,8 @@
 from dataclasses import asdict, dataclass, field
 from typing import Any, Final
 
+import yaml
+
 from tests.network.libs.apimachinery import dict_normalization_for_dataclass
 
 NETWORK_DATA: Final[str] = "networkData"
@@ -51,9 +53,33 @@ class NetworkData:
     ethernets: dict[str, EthernetDevice]
 
 
+@dataclass
+class UserData:
+    """Represents user configuration for cloud-init."""
+
+    users: list[Any]
+    """
+    Part of cloud-init's 'users and groups' module:
+    https://cloudinit.readthedocs.io/en/latest/reference/modules.html#users-and-groups
+    """
+
+
+def todict(no_cloud: NetworkData | UserData) -> dict[str, Any]:
+    return asdict(obj=no_cloud, dict_factory=dict_normalization_for_dataclass)
+
+
+def asyaml(no_cloud: NetworkData | UserData) -> str:
+    return yaml.safe_dump(todict(no_cloud=no_cloud), sort_keys=False)
+
+
+def format_cloud_config(userdata: UserData) -> str:
+    """
+    Formats UserData as a cloud-init '#cloud-config' file.
+    See: https://cloudinit.readthedocs.io/en/latest/explanation/format.html#headers-and-content-types
+    """
+    body = yaml.safe_dump(todict(no_cloud=userdata), sort_keys=False)
+    return f"#cloud-config\n{body}"
+
+
 def cloudinit(netdata: NetworkData) -> dict[str, Any]:
-    return {NETWORK_DATA: todict(netdata)}
-
-
-def todict(netdata: NetworkData) -> dict[str, Any]:
-    return asdict(obj=netdata, dict_factory=dict_normalization_for_dataclass)
+    return {NETWORK_DATA: todict(no_cloud=netdata)}
--- a/tests/network/libs/nodenetworkconfigurationpolicy.py
+++ b/tests/network/libs/nodenetworkconfigurationpolicy.py
@@ -3,6 +3,7 @@ from collections.abc import Iterator
 from copy import deepcopy
 from dataclasses import asdict, dataclass
 from datetime import datetime
+from enum import Enum
 from typing import Any
 
 from ocp_resources.exceptions import NNCPConfigurationFailed
@@ -14,6 +15,7 @@ from tests.network.libs.apimachinery import dict_normalization_for_dataclass
 
 WAIT_FOR_STATUS_TIMEOUT_SEC = 90
 WAIT_FOR_STATUS_INTERVAL_SEC = 5
+DEFAULT_OVN_EXTERNAL_BRIDGE = "br-ex"  # Default name for OVN-Kubernetes external bridge
 
 
 @dataclass
@@ -66,6 +68,21 @@ class Interface:
 
 
 @dataclass
+class BridgeMappings:
+    class State(Enum):
+        PRESENT = "present"
+
+    localnet: str
+    bridge: str
+    state: str
+
+
+@dataclass
+class OVN:
+    bridge_mappings: list[BridgeMappings]
+
+
+@dataclass
 class DesiredState:
     """
     Represents the desired network configuration for NMstate.
@@ -73,7 +90,8 @@ class DesiredState:
     https://nmstate.io/devel/yaml_api.html
     """
 
-    interfaces: list[Interface]
+    interfaces: list[Interface] | None = None
+    ovn: OVN | None = None
 
 
 class NodeNetworkConfigurationPolicy(Nncp):
@@ -109,11 +127,14 @@ class NodeNetworkConfigurationPolicy(Nncp):
         return self._desired_state
 
     def clean_up(self, wait: bool = True, timeout: int | None = None) -> bool:
+        desired_state: dict[str, Any] = {}
         if to_delete_interfaces := self._interfaces_for_deletion():
+            desired_state["interfaces"] = to_delete_interfaces
+        if to_delete_ovn := self._ovn_for_deletion():
+            desired_state["ovn"] = to_delete_ovn
+        if desired_state:
             with self._confirm_status_success_on_update():
-                ResourceEditor(
-                    patches={self: {"spec": {"desiredState": {"interfaces": to_delete_interfaces}}}}
-                ).update()
+                ResourceEditor(patches={self: {"spec": {"desiredState": desired_state}}}).update()
         return super().clean_up(wait=wait, timeout=timeout)
 
     @retry(
@@ -146,10 +167,16 @@ class NodeNetworkConfigurationPolicy(Nncp):
         return {}
 
     def _interfaces_for_deletion(self) -> list[Interface]:
-        to_delete_interfaces = deepcopy(self.desired_state["interfaces"])
+        to_delete_interfaces = deepcopy(self.desired_state.get("interfaces", []))
         for iface in to_delete_interfaces:
             iface["state"] = Nncp.Interface.State.ABSENT
         return to_delete_interfaces
+
+    def _ovn_for_deletion(self) -> OVN:
+        to_delete_ovn = deepcopy(self.desired_state.get("ovn", {}))
+        for bridge_mapping in to_delete_ovn.get("bridge-mappings", []):
+            bridge_mapping["state"] = Nncp.Interface.State.ABSENT
+        return to_delete_ovn
 
     @contextlib.contextmanager
     def _confirm_status_success_on_update(self) -> Iterator[None]:
--- a/tests/network/localnet/conftest.py
+++ b/tests/network/localnet/conftest.py
@@ -1,0 +1,118 @@
+from collections.abc import Generator
+
+import pytest
+from ocp_resources.namespace import Namespace
+
+import tests.network.libs.nodenetworkconfigurationpolicy as libnncp
+from libs.net import netattachdef
+from libs.net.traffic_generator import Client, Server
+from libs.net.vmspec import IP_ADDRESS, lookup_iface_status
+from libs.vm.vm import BaseVirtualMachine
+from tests.network.localnet.liblocalnet import NETWORK_NAME, localnet_nad, localnet_vm
+from utilities.constants import (
+    WORKER_NODE_LABEL_KEY,
+)
+from utilities.infra import create_ns
+
+_IPERF_SERVER_PORT = 5201
+
+
+@pytest.fixture(scope="module")
+def nncp_localnet() -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
+    desired_state = libnncp.DesiredState(
+        ovn=libnncp.OVN([
+            libnncp.BridgeMappings(
+                localnet=NETWORK_NAME,
+                bridge=libnncp.DEFAULT_OVN_EXTERNAL_BRIDGE,
+                state=libnncp.BridgeMappings.State.PRESENT.value,
+            )
+        ])
+    )
+
+    with libnncp.NodeNetworkConfigurationPolicy(
+        name="test-localnet-nncp",
+        desired_state=desired_state,
+        node_selector={WORKER_NODE_LABEL_KEY: ""},
+    ) as nncp:
+        nncp.wait_for_status_success()
+        yield nncp
+
+
+@pytest.fixture(scope="module")
+def namespace_localnet_1() -> Generator[Namespace]:
+    yield from create_ns(name="test-localnet-ns1")  # type: ignore
+
+
+@pytest.fixture(scope="module")
+def namespace_localnet_2() -> Generator[Namespace]:
+    yield from create_ns(name="test-localnet-ns2")  # type: ignore
+
+
+@pytest.fixture(scope="module")
+def vlan_id(vlan_index_number: Generator[int]) -> int:
+    return next(vlan_index_number)
+
+
+@pytest.fixture(scope="module")
+def nad_localnet_1(
+    namespace_localnet_1: Namespace, vlan_id: int
+) -> Generator[netattachdef.NetworkAttachmentDefinition]:
+    with localnet_nad(namespace=namespace_localnet_1.name, name="test-localnet-nad1", vlan_id=vlan_id) as nad:
+        yield nad
+
+
+@pytest.fixture(scope="module")
+def nad_localnet_2(
+    namespace_localnet_2: Namespace, vlan_id: int
+) -> Generator[netattachdef.NetworkAttachmentDefinition]:
+    with localnet_nad(namespace=namespace_localnet_2.name, name="test-localnet-nad2", vlan_id=vlan_id) as nad:
+        yield nad
+
+
+@pytest.fixture(scope="module")
+def vm_localnet_1(nad_localnet_1: netattachdef.NetworkAttachmentDefinition) -> Generator[BaseVirtualMachine]:
+    with localnet_vm(
+        namespace=nad_localnet_1.namespace, name="test-vm1", network=nad_localnet_1.name, cidr="10.0.0.1/24"
+    ) as vm:
+        yield vm
+
+
+@pytest.fixture(scope="module")
+def vm_localnet_2(nad_localnet_2: netattachdef.NetworkAttachmentDefinition) -> Generator[BaseVirtualMachine]:
+    with localnet_vm(
+        namespace=nad_localnet_2.namespace, name="test-vm2", network=nad_localnet_2.name, cidr="10.0.0.2/24"
+    ) as vm:
+        yield vm
+
+
+@pytest.fixture(scope="module")
+def vms_localnet(
+    vm_localnet_1: BaseVirtualMachine, vm_localnet_2: BaseVirtualMachine
+) -> Generator[tuple[BaseVirtualMachine, BaseVirtualMachine]]:
+    vms = (vm_localnet_1, vm_localnet_2)
+    for vm in vms:
+        vm.start()
+    for vm in vms:
+        vm.wait_for_ready_status(status=True)
+        vm.vmi.wait_for_condition(condition=vm.Condition.Type.AGENT_CONNECTED, status=vm.Condition.Status.TRUE)
+    yield vms
+
+
+@pytest.fixture()
+def localnet_server(vms_localnet: tuple[BaseVirtualMachine, BaseVirtualMachine]) -> Generator[Server]:
+    vm1, _ = vms_localnet
+    with Server(vm=vm1, port=_IPERF_SERVER_PORT) as server:
+        assert server.is_running()
+        yield server
+
+
+@pytest.fixture()
+def localnet_client(vms_localnet: tuple[BaseVirtualMachine, BaseVirtualMachine]) -> Generator[Client]:
+    vm1, vm2 = vms_localnet
+    with Client(
+        vm=vm2,
+        server_ip=lookup_iface_status(vm=vm1, iface_name=NETWORK_NAME)[IP_ADDRESS],
+        server_port=_IPERF_SERVER_PORT,
+    ) as client:
+        assert client.is_running()
+        yield client
--- a/tests/network/localnet/liblocalnet.py
+++ b/tests/network/localnet/liblocalnet.py
@@ -1,0 +1,74 @@
+from libs.net import netattachdef
+from libs.net.vmspec import add_network_interface, add_volume_disk
+from libs.vm.affinity import new_pod_anti_affinity
+from libs.vm.factory import base_vmspec, fedora_vm
+from libs.vm.spec import CloudInitNoCloud, Interface, Metadata, Multus, Network
+from libs.vm.vm import BaseVirtualMachine, cloudinitdisk_storage
+from tests.network.libs import cloudinit
+
+NETWORK_NAME = "localnet-network"
+
+
+def localnet_vm(namespace: str, name: str, network: str, cidr: str) -> BaseVirtualMachine:
+    """
+    Create a Fedora-based Virtual Machine connected to a given localnet network with a static IP configuration.
+
+    The VM will:
+    - Attach to a Multus network using a bridge interface.
+    - Apply a specific label for anti-affinity scheduling.
+    - Use cloud-init to configure a static IP address.
+    - Based on a standard Fedora VM template.
+
+    Args:
+        namespace (str): The namespace where the VM should be created.
+        name (str): The name of the VM.
+        network (str): The name of the Multus network to attach.
+        cidr (str): The CIDR address to assign to the VM's interface.
+
+    Returns:
+        BaseVirtualMachine: The configured VM object ready for creation.
+    """
+    spec = base_vmspec()
+    spec.template.metadata = spec.template.metadata or Metadata()
+    spec.template.metadata.labels = spec.template.metadata.labels or {}
+    localnet_test_label = {"test": "localnet"}
+    spec.template.metadata.labels.update(localnet_test_label)
+    vmi_spec = spec.template.spec
+
+    vmi_spec = add_network_interface(
+        vmi_spec=vmi_spec,
+        network=Network(name=NETWORK_NAME, multus=Multus(networkName=network)),
+        interface=Interface(name=NETWORK_NAME, bridge={}),
+    )
+
+    netdata = cloudinit.NetworkData(ethernets={"eth0": cloudinit.EthernetDevice(addresses=[cidr])})
+    # Prevents cloud-init from overriding the default OS user credentials
+    userdata = cloudinit.UserData(users=[])
+    disk, volume = cloudinitdisk_storage(
+        data=CloudInitNoCloud(
+            networkData=cloudinit.asyaml(no_cloud=netdata), userData=cloudinit.format_cloud_config(userdata=userdata)
+        )
+    )
+    vmi_spec = add_volume_disk(vmi_spec=vmi_spec, volume=volume, disk=disk)
+
+    vmi_spec.affinity = new_pod_anti_affinity(label=next(iter(localnet_test_label.items())))
+    vmi_spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].namespaceSelector = {}
+
+    return fedora_vm(namespace=namespace, name=name, spec=spec)
+
+
+def localnet_nad(namespace: str, name: str, vlan_id: int) -> netattachdef.NetworkAttachmentDefinition:
+    return netattachdef.NetworkAttachmentDefinition(
+        namespace=namespace,
+        name=name,
+        config=netattachdef.NetConfig(
+            NETWORK_NAME,
+            [
+                netattachdef.CNIPluginOvnK8sConfig(
+                    topology=netattachdef.CNIPluginOvnK8sConfig.Topology.LOCALNET.value,
+                    netAttachDefName=f"{namespace}/{name}",
+                    vlanID=vlan_id,
+                )
+            ],
+        ),
+    )
--- a/tests/network/localnet/test_localnet.py
+++ b/tests/network/localnet/test_localnet.py
@@ -1,0 +1,11 @@
+import pytest
+
+from libs.net.traffic_generator import is_tcp_connection
+from utilities.virt import migrate_vm_and_verify
+
+
+@pytest.mark.ipv4
+@pytest.mark.polarion("CNV-11775")
+def test_connectivity_over_migration_between_localnet_vms(nncp_localnet, localnet_server, localnet_client):
+    migrate_vm_and_verify(vm=localnet_client.vm)
+    assert is_tcp_connection(server=localnet_server, client=localnet_client)
--- a/libs/net/traffic_generator.py
+++ b/libs/net/traffic_generator.py
@@ -77,6 +77,10 @@ class Client:
     def __exit__(self, exc_type: BaseException, exc_value: BaseException, traceback: object) -> None:
         _stop_process(vm=self._vm, cmd=self._cmd)
 
+    @property
+    def vm(self) -> BaseVirtualMachine:
+        return self._vm
+
     def is_running(self) -> bool:
         return _is_process_running(vm=self._vm, cmd=self._cmd)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,8 @@ warn_redundant_casts = true
 [[tool.mypy.overrides]]
 module = [
   "libs.*",
-  "tests.network.libs.*"
+  "tests.network.libs.*",
+  "tests.network.localnet.*"
 ]
 disallow_untyped_calls = true
 disallow_untyped_defs = true


cnv-qe-bot1:

The following are automatically added:
 * Add reviewers from OWNER file (in the root of the repository) under reviewers section.
 * Set PR size label.
 * New issue is created for the PR. (Closed when PR is merged/closed)
 * Run [pre-commit](https://pre-commit.ci/) if `.pre-commit-config.yaml` exists in the repo.

Available user actions:
 * To mark PR as WIP comment `/wip` to the PR, To remove it from the PR comment `/wip cancel` to the PR.
 * To block merging of PR comment `/hold`, To un-block merging of PR comment `/hold cancel`.
 * To mark PR as verified comment `/verified` to the PR, to un-verify comment `/verified cancel` to the PR.
        verified label removed on each new commit push.
 * To cherry pick a merged PR comment `/cherry-pick <target branch to cherry-pick to>` in the PR.
    * Multiple target branches can be cherry-picked, separated by spaces. (`/cherry-pick branch1 branch2`)
    * Cherry-pick will be started when PR is merged
 * To build and push container image command `/build-and-push-container` in the PR (tag will be the PR number).
    * You can add extra args to the Podman build command
        * Example: `/build-and-push-container --build-arg OPENSHIFT_PYTHON_WRAPPER_COMMIT=<commit_hash>`
 * To add a label by comment use `/<label name>`, to remove, use `/<label name> cancel`
 * To assign reviewers based on OWNERS file use `/assign-reviewers`
 * To check if PR can be merged use `/check-can-merge`
 * to assign reviewer to PR use `/assign-reviewer @<reviewer>`

<details>
<summary>Supported /retest check runs</summary>

 * `/retest tox`: Retest tox
 * `/retest build-container`: Retest build-container
 * `/retest all`: Retest all

</details>

<details>
<summary>Supported labels</summary>

 * hold
 * verified
 * wip
 * lgtm

</details>

sbahar619:
Change: rebase.

libs/vm/vm.py
```diff
@@ -87,3 +81,10 @@ def container_image(base_image: str) -> str:
 def containerdisk_storage(image: str) -> tuple[SpecDisk, Volume]:
     name = "containerdisk"
     return SpecDisk(name=name, disk=Disk(bus="virtio")), Volume(name=name, containerDisk=ContainerDisk(image=image))
+
+
+def cloudinitdisk_storage(data: dict[Any, Any]) -> tuple[SpecDisk, Volume]:
```
EdDev, line 86: If you intended for the input to be a dict, I think it should be more accurate to do: `dict[str, Any]`.
We could make it even more accurate with a recursion definition, but it is not a must for now. We should struct it anyway in the near future.

libs/vm/factory.py
```diff
@@ -38,8 +38,8 @@ def _fill_vm_spec_defaults(spec: VMSpec | None) -> VMSpec:
     vmi_spec.volumes = vmi_spec.volumes or []
 
     disk, volume = containerdisk_storage(image=fedora_image())
-    vmi_spec.domain.devices.disks.append(disk)
-    vmi_spec.volumes.append(volume)
+    vmi_spec.domain.devices.disks.insert(0, disk)
+    vmi_spec.volumes.insert(0, volume)
```
EdDev, line 42: Should be explained in its own commit, this is not trivial.

libs/vm/spec.py
```diff
@@ -108,6 +108,7 @@ class PodAntiAffinity:
 class PodAffinityTerm:
     labelSelector: LabelSelector  # noqa: N815
     topologyKey: str  # noqa: N815
+    namespaceSelector: dict[Any, Any] | None = None  # noqa: N815
```
EdDev, line 111: Maybe you mean `dict[str, str]`.
I also do not understand why is it needed, please provide some reasoning to have it.

libs/vm/spec.py
```diff
@@ -126,14 +127,9 @@ class LabelSelectorRequirement:
 class Volume:
     name: str
     containerDisk: ContainerDisk | None = None  # noqa: N815
-    cloudInitNoCloud: CloudInitNoCloud | None = None  # noqa: N815
+    cloudInitNoCloud: str | None = None  # noqa: N815
```
EdDev, line 130: You will have to explain this change as well. From the first look, it flattens the data instead of keeping network and user data separated, which is not a good direction IMO.

libs/net/vmspec.py
```diff
@@ -9,6 +9,7 @@
 
 LOOKUP_IFACE_STATUS_TIMEOUT_SEC: Final[int] = 30
 RETRY_INTERVAL_SEC: Final[int] = 5
+IP_ADDRESS: str = "ipAddress"
```
EdDev, line 12: It should be final.

libs/net/netattachdef.py
```diff
@@ -30,6 +30,22 @@ class CNIPluginBridgeConfig(CNIPluginConfig):
...
+    """
+    CNI OVN-Kubernetes Plugin
+    Ref:
+    https://docs.openshift.com/container-platform/4.14/networking/multiple_networks/
+    configuring-additional-network.html#configuration-ovnk-network-plugin-json-object_
+    configuring-additional-network
+    """
+
+    type: str = field(default="ovn-k8s-cni-overlay", init=False)
+    topology: str
```
EdDev, line 44: We should limit this to known values somehow. If not limit, then at least provide the options as some "empty" classes like it was done with conditions.

utilities/virt.py
```diff
@@ -142,7 +142,7 @@ def wait_for_vm_interfaces(vmi: VirtualMachineInstance, timeout: int = TIMEOUT_1
     return False
 
 
-def generate_cloud_init_data(data):
+def generate_cloud_init_data(data: dict[Any, Any]):
```
EdDev, line 145: I have a PR in the working to replace this mess.
But that should not be a blocker for this PR. If it will be ready before this one, I would prefer using that one instead.

EdDev (changes requested):
I have reviewed only the libs/tools. Need to continue with the test file itself.
Posting this for now to allow the author to start responding, but it may be useful to wait before posting again on the PR  (and wait for a complete review).

libs/net/netattachdef.py
sbahar619: ack

libs/net/vmspec.py
sbahar619: ack

libs/vm/factory.py
sbahar619: Do you mean new PR for this change?

libs/vm/spec.py
sbahar619: No, dict[str, str] doesn't match the expected structure of this API, which includes nested objects and dictionaries.

We need to use namespaceSelector because we deploy two VMs in different namespaces. To ensure the pod anti-affinity rule applies across both namespaces, namespaceSelector must be set.

Ref:
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#namespace-selector

libs/vm/spec.py
sbahar619: Right, I adjusted it  for convenient so we can use the existing utilities which expect str.
Without it, we need to manipulate the data somehow so the utilities will know how to work with it or implement our own utilities.

libs/vm/vm.py
sbahar619: ack

tests/network/libs/nodenetworkconfigurationpolicy.py
```diff
@@ -134,7 +148,14 @@ def wait_for_status_success(self) -> dict[str, Any]:
 
     def _delete_interfaces(self) -> None:
         desired_state = deepcopy(self.desired_state)
-        for iface in desired_state["interfaces"]:
+        for iface in desired_state.get("interfaces", []):
```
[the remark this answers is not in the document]
sbahar619: ack

tests/network/libs/nodenetworkconfigurationpolicy.py
```diff
@@ -134,7 +148,14 @@ def wait_for_status_success(self) -> dict[str, Any]:
...
             iface["state"] = Nncp.Interface.State.ABSENT
-        if desired_state["interfaces"]:
+        if desired_state.get("interfaces"):
             ResourceEditor(patches={self: {"spec": {"desiredState": desired_state}}}).update()
+
+    def _delete_bridge_mappings(self) -> None:
```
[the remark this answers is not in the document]
sbahar619: ack

utilities/virt.py
sbahar619: Great thanks!
Any action from my side to do?

tests/network/libs/nodenetworkconfigurationpolicy.py
```diff
@@ -134,7 +148,14 @@ def wait_for_status_success(self) -> dict[str, Any]:
 
     def _delete_interfaces(self) -> None:
         desired_state = deepcopy(self.desired_state)
-        for iface in desired_state["interfaces"]:
+        for iface in desired_state.get("interfaces", []):
             iface["state"] = Nncp.Interface.State.ABSENT
-        if desired_state["interfaces"]:
+        if desired_state.get("interfaces"):
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/libs/nodenetworkconfigurationpolicy.py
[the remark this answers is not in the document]
sbahar619: Done.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/a942c3c825fcc88cf910325ddf8ec55b14be59b5..c9edfaeb31d97351338e0e4910399f010f2a855e
Resolve @EdDev comment.

sbahar619:
Change: rebase.

libs/vm/factory.py
EdDev: I mean commit.

libs/net/netattachdef.py
sbahar619: Done.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/c188889e34a6f7ab10e6e0ae3b1b4f064833de35..a3820bf1463f16d6f7654d02db0ae9e96dbb8352
Address @EdDev comment.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/a3820bf1463f16d6f7654d02db0ae9e96dbb8352..556402aac5977c4a0e47a0c769b717ff2cb709b3
Revert previous change.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/556402aac5977c4a0e47a0c769b717ff2cb709b3..2bbb1086e5c2703ff8eec2cca7a5cc79dd8105cf
Addess @EdDev comment.

libs/net/vmspec.py
sbahar619: Done.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/2bbb1086e5c2703ff8eec2cca7a5cc79dd8105cf..0284abef9b35b54d4db312cd897a14f4463882ee
Address @EdDev comment.

libs/vm/factory.py
sbahar619: Done.

tests/network/libs/nodenetworkconfigurationpolicy.py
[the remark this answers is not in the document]
sbahar619: After a discussion, we agreed to wait for this change to merge.

libs/vm/spec.py
sbahar619: After a discussion, we agreed to wait with this changes for your future cloud-init PR.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+                    netAttachDefName=f"{namespace}/{name}",
+                    vlanID=vlan_id,
+                )
+            ],
+        ),
+    )
+
+
+@pytest.fixture(scope="module")
+def nncp():
```
Anatw, line 68: Please use more meaningful names, containing 'localnet' with specification where possible. For example here, you can name it localnet_nncp_on_brex

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+        name="test-localnet-nncp",
+        desired_state=desired_state,
+        node_selector={WORKER_NODE_LABEL_KEY: ""},
+    ) as nncp:
+        nncp.wait_for_status_success()
+        yield nncp
+
+
+@pytest.fixture(scope="module")
+def ns1():
```
Anatw, line 89: ditto

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+    yield from create_ns(
+        name="test-localnet-ns1",
+    )
+
+
+@pytest.fixture(scope="module")
```
Anatw, line 95: ditto

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+def nad2(ns2, vlan_id):
+    with localnet_nad(namespace=ns2.name, name="test-localnet-nad2", vlan_id=vlan_id) as nad:
+        yield nad
+
+
+@pytest.fixture(scope="module")
+def vm1(nad1):
```
Anatw, line 120: ditto

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+    with localnet_vm(namespace=nad1.namespace, name="test-vm1", network=nad1.name, cidr="10.0.0.1/24") as vm:
+        vm.start(wait=True)
+        yield vm
+
+
+@pytest.fixture(scope="module")
+def vm2(nad2):
```
Anatw, line 127: ditto

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+)
+from utilities.constants import WORKER_NODE_LABEL_KEY
+from utilities.infra import create_ns
+from utilities.network import cloud_init_network_data
+from utilities.virt import migrate_vm_and_verify
+
+LABEL = {"test": "localnet"}
+TOPOLOGY = "localnet"
+NETWORK_NAME = "localnet-network"
+_SERVER_PORT = 5201
```
Anatw, line 25: Why is this port chosen?

Anatw (changes requested):

tests/network/localnet/test_localnet.py
sbahar619: It's the default port used by iperf.

tests/network/localnet/test_localnet.py
sbahar619: Since this fixture is already defined in a module with "localnet" in the name, I felt it would be redundant to include it again. Regarding br-ex, it’s a user-defined value that might change, and tying it to the fixture name could make it harder to maintain. That said, I’m open to a more descriptive name if we feel it's unclear in context.

sbahar619:
Change: rebase.

sbahar619:
Change: rebase + resolve conflicts.
https://github.com/RedHatQE/openshift-virtualization-tests/compare/edc9cc18b4d566feab0b7c1b6c62f75ef752c760..394237eb2c1558ddbad146e178cb6d91ba4c98c8

sbahar619:
Change:  resolve conflicts.

utilities/virt.py
sbahar619: Thanks!
I pushed a commit based on your PR.
#754

sbahar619:
Change: fix pre-commit.

sbahar619:
Change: rebase.

EdDev (changes requested):
Very nice, thank you!

I still need to go over the test itself, I mainly reviewed the libs.
I am posting what I have now and will continue to the test.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/290734c560280f85616fa3ddc70f6d05e4ae6a8d..2b785ba4eaf7ff57e7129919fe696f044488a30f
Address @EdDev comment.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+from libs.vm.affinity import new_pod_anti_affinity
+from libs.vm.factory import base_vmspec, fedora_vm
+from libs.vm.spec import Interface, Multus, Network
+from libs.vm.vm import cloudinitdisk_storage
+from tests.network.libs.nodenetworkconfigurationpolicy import (
+    OVN,
+    BridgeMappings,
+    DesiredState,
+    NodeNetworkConfigurationPolicy,
+)
```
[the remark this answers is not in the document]
Anatw: +1 this will help make it clear that this is not the nncp from the wrapper which is currently in use in most our tests.

tests/network/libs/cloudinit.py
```diff
@@ -1,14 +1,45 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
sbahar619: Fixed.

tests/network/localnet/test_localnet.py
Anatw: I think that function/fixture names should be clear on their own.
br-ex has been here for some time and if it changes the name should be changed, which is not a big deal. If we have a clear name you don't need to dig into the function/fixture to understand what it does, and reading the code is easier.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/93f95b1d68159d0fae6fb1ffa8199e5283d0371b..2ef25579a1e1dcbe6f9dc4d62b12bc888303cd6e
Address @EdDev comment.

tests/network/localnet/test_localnet.py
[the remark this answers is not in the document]
sbahar619: Done.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/2ef25579a1e1dcbe6f9dc4d62b12bc888303cd6e..67fcb29335b59ec5db0dac482c56d2d3797e3b2c
Address @EdDev comment.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,160 @@
...
+    ) as nncp:
+        nncp.wait_for_status_success()
+        yield nncp
+
+
+@pytest.fixture(scope="module")
+def ns1():
+    yield from create_ns(
+        name="test-localnet-ns1",
+    )
```
EdDev, line 89: nit: Why this is not one-liner? Same for the other one.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,160 @@
...
+        name="test-localnet-nncp",
+        desired_state=desired_state,
+        node_selector={WORKER_NODE_LABEL_KEY: ""},
+    ) as nncp:
+        nncp.wait_for_status_success()
+        yield nncp
+
+
+@pytest.fixture(scope="module")
+def ns1():
```
EdDev, line 86: Based on the current codebase, ns1 is used only at the NAD fixture, which means you do not need this fixture. Same for ns2.

sbahar619:
Change: rebase.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,160 @@
...
+@pytest.fixture(scope="module")
+def nad2(ns2, vlan_id):
+    with localnet_nad(namespace=ns2.name, name="test-localnet-nad2", vlan_id=vlan_id) as nad:
+        yield nad
+
+
+@pytest.fixture(scope="module")
+def vm1(nad1):
+    with localnet_vm(namespace=nad1.namespace, name="test-vm1", network=nad1.name, cidr="10.0.0.1/24") as vm:
+        vm.start(wait=True)
```
EdDev, line 119: I see that you added a different fixture to wait for the VMs to be ready.
And I assume you did it so we will not wait for each serially.
Therefore, this wait here also needs to be removed and passed to the same location.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,160 @@
...
+
+@pytest.fixture(scope="module")
+def vm2(nad2):
+    with localnet_vm(namespace=nad2.namespace, name="test-vm2", network=nad2.name, cidr="10.0.0.2/24") as vm:
+        vm.start(wait=True)
+        yield vm
+
+
+@pytest.fixture()
+def wait_for_vms_to_be_ready(vm1, vm2):
```
EdDev, line 131: This is an action/verb, it does not fit a fixture.
There are two options:
- Move it to the test.
- Call it `localnet_vms` that return the two vms in a tuple and remove the other specific vm fixtures.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,160 @@
...
+        server_port=_IPERF_SERVER_PORT,
+    ) as client:
+        assert client.is_running()
+        yield client
+
+
+@pytest.mark.ipv4
+@pytest.mark.polarion("CNV-10804")
+def test_connectivity_between_localnet_vms(nncp, wait_for_vms_to_be_ready, localnet_server, localnet_client):
+    migrate_vm_and_verify(vm=localnet_client.vm)
```
EdDev, line 159: Why did you ref the vm from `localnet_client.vm`?

EdDev (changes requested):

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/277462dde4d20e9e6ff822c815c2da612a685a39..4b9c66d98692e81f31469e6151ce05de746d3937
Adress @Anatw comments.

tests/network/localnet/test_localnet.py
sbahar619: Done.

tests/network/localnet/test_localnet.py
sbahar619: Done.

tests/network/localnet/test_localnet.py
sbahar619: Done.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+
+
+@pytest.fixture(scope="module")
+def vlan_id(vlan_index_number):
+    return next(vlan_index_number)
+
+
+@pytest.fixture(scope="module")
+def nad1(ns1, vlan_id):
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/localnet/test_localnet.py
sbahar619: Done.

tests/network/localnet/test_localnet.py
sbahar619: Done.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,165 @@
...
+_SERVER_PORT = 5201
+
+
+def localnet_vm(namespace, name, network, cidr):
+    spec = base_vmspec()
+    spec.template.metadata.labels = LABEL
+
+    vmi_spec = spec.template.spec
+    vmi_spec.domain.devices.interfaces = [Interface(name=NETWORK_NAME, bridge={})]
+    vmi_spec.networks = [Network(name=NETWORK_NAME, multus=Multus(networkName=network))]
```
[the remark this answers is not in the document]
sbahar619: Done.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/4b9c66d98692e81f31469e6151ce05de746d3937..9c1a6b43b458716b26b8d2766a1c4d61d2c48fac
Address @EdDev comment.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/9c1a6b43b458716b26b8d2766a1c4d61d2c48fac..08d312381f14778dcb2f2429a5fe91ccd02bbe6b
Address @EdDev comment.

tests/network/localnet/test_localnet.py
sbahar619: I used localnet_client.vm because I needed to pass the VM instance itself to migrate_vm_and_verify. While I could have used the standalone 'vm_localnet_2' fixture, it felt out of context in this case. Since localnet_client already encapsulates the VM, referencing it directly from there felt more natural and consistent within the test flow.

libs/net/netattachdef.py
sbahar619: I used Enum:
https://github.com/RedHatQE/openshift-virtualization-tests/blob/d4caa0c8bdc74b2317de1f24df71e21b69c12455/tests/network/localnet/test_localnet.py#L56

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+
+@pytest.fixture(scope="module")
+def nncp():
+    desired_state = DesiredState(
+        ovn=OVN([
+            BridgeMappings(
+                localnet=NETWORK_NAME,
+                bridge="br-ex",
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/localnet/test_localnet.py
sbahar619: Right, thanks!
Done.

tests/network/localnet/test_localnet.py
sbahar619: Good point, thanks, done.

sbahar619:
Change: rebase.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/5b6a9babe4dd884a9374ff2ac66145ba5527121f..c6053673785cfabcdea2351d4a2e3832e135d946
Refactor.

tests/network/libs/nodenetworkconfigurationpolicy.py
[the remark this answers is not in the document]
sbahar619: Update, where do you except to do it?
Currently I do use it:
https://github.com/RedHatQE/openshift-virtualization-tests/blob/c6053673785cfabcdea2351d4a2e3832e135d946/tests/network/libs/nodenetworkconfigurationpolicy.py#L173
Because 'interfaces' is optional, therefore it have a None default value, as it is immutable, if I set it as empty list, it will be mutable and not safe (will be shared across object instances).

tests/network/libs/nodenetworkconfigurationpolicy.py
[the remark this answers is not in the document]
sbahar619: Done.

libs/vm/vm.py
sbahar619: Now we use: CloudInitNoCloud type.

tests/network/localnet/test_localnet.py
sbahar619: Done.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,166 @@
...
+    vmi_spec.domain.devices.interfaces = [Interface(name=NETWORK_NAME, bridge={})]
+    vmi_spec.networks = [Network(name=NETWORK_NAME, multus=Multus(networkName=network))]
+
+    vmi_spec.affinity = new_pod_anti_affinity(label=next(iter(LABEL.items())))
+    vmi_spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].namespaceSelector = {}
+
+    netdata = cloudinit.NetworkData(ethernets={"eth0": cloudinit.EthernetDevice(addresses=[cidr])})
+    # Prevents cloud-init from overriding the default OS user credentials
+    userdata = cloudinit.UserData(users=[])
+    cloud_init_data = CloudInitNoCloud(networkData=netdata.asyaml(), userData=userdata.asyaml())
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,165 @@
...
+
+    disk, volume = cloudinitdisk_storage(data=cloud_init_data)
+    vmi_spec.domain.devices.disks = vmi_spec.domain.devices.disks or []
+    vmi_spec.volumes = vmi_spec.volumes or []
+    vmi_spec.domain.devices.disks.append(disk)
+    vmi_spec.volumes.append(volume)
```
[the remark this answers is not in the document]
sbahar619: Done.

libs/vm/spec.py
sbahar619: I squashed the commits, as when I sart working on it, I used it as separate commit because some changes were on your unmerged PR, now that its merged, no need a separate commit.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,163 @@
...
+    BridgeMappings,
+    DesiredState,
+    NodeNetworkConfigurationPolicy,
+)
+from utilities.constants import WORKER_NODE_LABEL_KEY
+from utilities.infra import create_ns
+from utilities.network import cloud_init_network_data
+from utilities.virt import migrate_vm_and_verify
+
+LABEL = {"test": "localnet"}
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/libs/nodenetworkconfigurationpolicy.py
```diff
@@ -108,11 +121,18 @@ def desired_state_spec(self) -> DesiredState:
...
+        to_delete_interfaces = self._interfaces_for_deletion()
+        to_delete_ovn = self._ovn_for_deletion()
+
+        if to_delete_interfaces or to_delete_ovn:
+            desired_state = {}
+
+            if to_delete_interfaces:
+                desired_state["interfaces"] = to_delete_interfaces
+            if to_delete_ovn:
+                desired_state["ovn"] = to_delete_ovn
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,166 @@
...
+TOPOLOGY = "localnet"
+NETWORK_NAME = "localnet-network"
+_SERVER_PORT = 5201
+
+
+def localnet_vm(namespace, name, network, cidr):
+    spec = base_vmspec()
+    spec.template.metadata.labels = LABEL
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/localnet/test_localnet.py
sbahar619: Since create_ns is a generator-based fixture, removing the separate ns fixture would require manually handling the generator (e.g., next() and StopIteration) within the nad fixture, adding unnecessary complexity. Keeping the ns fixture maintains cleaner, more readable code.

tests/network/localnet/test_localnet.py
sbahar619: Done, I chose option 2.

sbahar619:
Change: rebase.

sbahar619:
Change: rebase.

tests/network/libs/nodenetworkconfigurationpolicy.py
```diff
@@ -14,6 +15,7 @@
 
 WAIT_FOR_STATUS_TIMEOUT_SEC = 90
 WAIT_FOR_STATUS_INTERVAL_SEC = 5
+OVN_EXTERNAL_BRIDGE = "br-ex"  # Default name for OVN-Kubernetes external bridge
```
EdDev, line 18: If it can be also a different name, then maybe add `DEFAULT` into the name of the const?

tests/network/localnet/test_localnet.py
EdDev, line 1: This is a new package, please cover it with mypy.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,157 @@
...
+        name="test-localnet-nncp",
+        desired_state=desired_state,
+        node_selector={WORKER_NODE_LABEL_KEY: ""},
+    ) as nncp:
+        nncp.wait_for_status_success()
+        yield nncp
+
+
+@pytest.fixture(scope="module")
+def ns_localnet_1():
```
EdDev, line 87: I won't insist, but I believe all these fixtures are useless and provide no benefit because they are used by only one other fixture. It also adds a lot of code-lines for no good reason.

Look:
```python
pytest.fixture(scope="module")
def nad_localnet_1(vlan_index_number):
  ns = next(create_ns(name="test-localnet-ns1"))
  vlan_id = next(vlan_index_number)
  with localnet_nad(namespace=ns.name, name="test-localnet-nad1", vlan_id=vlan_id) as nad:
    yield nad
```

But you decide, it can be changed later as well.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,157 @@
...
+def localnet_vms(nad_localnet_1, nad_localnet_2):
+    with (
+        localnet_vm(
+            namespace=nad_localnet_1.namespace, name="test-vm1", network=nad_localnet_1.name, cidr="10.0.0.1/24"
+        ) as vm1,
+        localnet_vm(
+            namespace=nad_localnet_2.namespace, name="test-vm2", network=nad_localnet_2.name, cidr="10.0.0.2/24"
+        ) as vm2,
+    ):
+        vms = [vm1, vm2]
```
EdDev, line 123: nit: Can be tuple and you can return it at the end.

tests/network/localnet/test_localnet.py
EdDev: I understand, but it does break the encapsulation a bit.
Ok, lets have it as you suggested. Thanks.

tests/network/libs/cloudinit.py
```diff
@@ -50,10 +52,23 @@ class NetworkData:
...
     ethernets: dict[str, EthernetDevice]
 
+    def todict(self) -> dict[str, Any]:
+        return asdict(obj=self, dict_factory=dict_normalization_for_dataclass)
+
+    def asyaml(self) -> str | bytes:
+        return yaml.safe_dump(self.todict(), sort_keys=False)
+
+    def cloudinit(self) -> dict[str, Any]:
+        return {NETWORK_DATA: self.todict()}
```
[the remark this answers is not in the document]
vsibirsk: where is it said that dataclass is only a storage with no behavior?

tests/network/localnet/test_localnet.py
vsibirsk: it might save up some code lines, but fixture should do one thing only.
also how (and when) does in your example `ns` being teared-down?

tests/network/localnet/test_localnet.py
sbahar619: Great, thanks!

tests/network/localnet/test_localnet.py
sbahar619: I added __init__.py, is that what you meant?

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/464b4d23dc3679a8ac5eb172b2e93e9e62dd3ef6..5be350f8fccd6153539dee2bb88f2445b8fd268a
Address @EdDev comments.

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,157 @@
...
+        server_port=_IPERF_SERVER_PORT,
+    ) as client:
+        assert client.is_running()
+        yield client
+
+
+@pytest.mark.ipv4
+@pytest.mark.polarion("CNV-10804")
+def test_connectivity_between_localnet_vms(nncp_ovn_bridge_mapping, localnet_server, localnet_client):
+    migrate_vm_and_verify(vm=localnet_client.vm)
```
yossisegev, line 156: Why migrating?
Unless I am missing something, this is not part of this test case.

tests/network/libs/nodenetworkconfigurationpolicy.py
sbahar619: Done.

yossisegev (changes requested):

tests/network/libs/cloudinit.py
```diff
@@ -50,10 +52,23 @@ class NetworkData:
...
 
-def cloudinit(netdata: NetworkData) -> dict[str, Any]:
-    return {NETWORK_DATA: todict(netdata)}
 
+@dataclass
+class UserData:
+    users: list
+
+    def todict(self) -> dict[str, Any]:
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/libs/cloudinit.py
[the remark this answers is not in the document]
sbahar619: Done.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-464 published

tests/network/localnet/test_localnet.py
Anatw: The test tests connectivity over migration, so the migration is a crucial part.

tests/network/libs/cloudinit.py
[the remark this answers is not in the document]
EdDev: > where is it said that dataclass is only a storage with no behavior?

The [PEP](https://peps.python.org/pep-0557/#rationale) mentions:
> There have been numerous attempts to define classes which exist primarily **to store values** which are accessible by attribute lookup.

An unformal description of these, without considering a specific language can be found [here](https://refactoring.guru/smells/data-class).

Unfortunately, python does not like to limit and give flexibility to developers, assuming they have good reasons to do unorthodox things. Some find it as valuable and others (like me) find it challenging as it gets very easy to abuse tools and make the codebase harder to maintain/evolve.

Hope this helps.

tests/network/localnet/test_localnet.py
EdDev: No, please add this package here: https://github.com/RedHatQE/openshift-virtualization-tests/blob/8251443b5abbca38ae988960c0a0a03db9de8bf9/pyproject.toml#L19.
Then you will be asked to add typing all over.

tests/network/libs/cloudinit.py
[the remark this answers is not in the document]
vsibirsk: same PEP has methods defined almost in every code snip :)
agree that python is (too) flexible in some cases. so it's my IMHO, if the method is just a presentation of class own data (to_dict() for instance), then it's fine to have it (same as regular dict also has methods like keys(), pop() etc)

tests/network/localnet/test_localnet.py
EdDev: Fixtures are overused in this project and in fact, it is a known cons of them.

Fixtures should be limited to real test setup and not to getters or to manage resources of indexes or any other thing. For those you can use the right python tools, e.g. context-manager, pools, objects, etc.

The one thing a test needs here is a NAD, all the other stuff that fixture with or without other helpers can do.
The meaning of "one thing" is sometime wrongly interpreted as a single action, single object, etc. It is not accurate, the "one thing" is an abstraction or representation that can be grouped as one thing.

Anyway, I hoped you python experts can step up and solve my indeed wrong code above.
But your challenge is accepted :)
The whole evil comes from the fact that generators are not decorated as a `contextmanager`, which seems to be the real problem here. I cannot understand why otherwise it is used as a "naked" generator.
So we will just wrap them correctly:

So here it is the correct alternative:
```python
@contextmanager
def create_namespace(name: str):
  yield from create_ns(name=name)

@contextmanager
def vlan_indexer():
  yield from vlan_index_number

pytest.fixture(scope="module")
def nad_localnet_1(vlan_index_number):
  with create_namespace(name="test-localnet-ns1") as ns, vlan_indexer() as vlan_id:
    with localnet_nad(namespace=ns.name, name="test-localnet-nad1", vlan_id=vlan_id) as nad:
      yield nad
```

tests/network/localnet/test_localnet.py
EdDev: You should prefer immutable containers over mutable ones, it is safer. The fact you can return them in the end was a nice side effect over specifying them again.
But it is a nit, you do not have to do it.

tests/network/libs/cloudinit.py
```diff
@@ -51,9 +53,21 @@ class NetworkData:
...
+def asyaml(obj: Any) -> str | bytes:
+    if isinstance(obj, UserData):
+        body = yaml.safe_dump(todict(obj=obj), sort_keys=False)
+        return f"#cloud-config\n{body}"
+    return yaml.safe_dump(todict(obj=obj), sort_keys=False)
 
 
-def todict(netdata: NetworkData) -> dict[str, Any]:
-    return asdict(obj=netdata, dict_factory=dict_normalization_for_dataclass)
+def cloudinit(obj: Any) -> dict[str, Any]:
```
EdDev, line 72: Please keep the original signature, we want to work with structures and not with "whatever".

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,157 @@
...
+    assert is_tcp_connection(server=localnet_server, client=localnet_client)
```
EdDev, line 157: Don't you need to check this before the migration as well?

EdDev (changes requested):
The changes in the libs/cloudinit seem problematic to me now.
They lost structure and typing, please see what can be done.

tests/network/libs/cloudinit.py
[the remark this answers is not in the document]
EdDev: As I see it, the methods mentioned in the pep and that are indeed available as part of the dataclass itself come to serve its objective, i.e. to be a data class.
Same for the dict, all its methods are there to serve its purpose to act as a dict.

Anyway, I haved not argued it is forbidden, I just think that with such a opinionated rule, there is less chance things will leak in which should not be there.
BTW, if python thought that it is reasonable to export data as a dict or yaml from any dataclass and mix these, then they would have added `asdict` as a method and not as a [function](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict). IMO they have tried to keep clear separation between the data and operations on it.

tests/network/libs/cloudinit.py
```diff
@@ -51,9 +53,21 @@ class NetworkData:
...
+class UserData:
+    users: list
+
+
+def todict(obj: Any) -> dict[str, Any]:
+    return asdict(obj=obj, dict_factory=dict_normalization_for_dataclass)
+
+
+def asyaml(obj: Any) -> str | bytes:
+    if isinstance(obj, UserData):
```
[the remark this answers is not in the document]
sbahar619: Please explain why isinstance can be used in extreme cases when there is no other possible way to get out of the situation .
What alternative solution do you have in mind?

tests/network/localnet/test_localnet.py
sbahar619: No, because the is_tcp_connection function is executing:
server.is_running() and client.is_running()
The localnet_client and localnet_server fixtures each already perform those checks.

tests/network/localnet/test_localnet.py
sbahar619: @yossisegev 
Right, thanks, we need to update it.

tests/network/localnet/test_localnet.py
sbahar619: Actually no, because:
https://github.com/RedHatQE/openshift-virtualization-tests/blob/00a0b583aa13c53d3089d96cdeff44c81ba522cb/.pre-commit-config.yaml#L71
But I will add it anyway for future mudules added not as part of test_*

tests/network/libs/cloudinit.py
```diff
@@ -51,9 +53,21 @@ class NetworkData:
     ethernets: dict[str, EthernetDevice]
 
 
-def cloudinit(netdata: NetworkData) -> dict[str, Any]:
-    return {NETWORK_DATA: todict(netdata)}
+@dataclass
+class UserData:
+    users: list
```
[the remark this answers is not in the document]
sbahar619: https://cloudinit.readthedocs.io/en/latest/reference/modules.html#users-and-groups
To align with the configuration schema, we can add the relevant typing annotations, but it will be complicated as it supports many types and we currently not using any of them, we really just use empty list right now, so who ever in the future will need to expand it and use it deeper, and add the relevant types.

tests/network/libs/cloudinit.py
```diff
@@ -51,9 +53,21 @@ class NetworkData:
...
+
+
+def todict(obj: Any) -> dict[str, Any]:
+    return asdict(obj=obj, dict_factory=dict_normalization_for_dataclass)
+
+
+def asyaml(obj: Any) -> str | bytes:
+    if isinstance(obj, UserData):
+        body = yaml.safe_dump(todict(obj=obj), sort_keys=False)
+        return f"#cloud-config\n{body}"
```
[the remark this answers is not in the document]
sbahar619: It is required:
https://cloudinit.readthedocs.io/en/latest/explanation/format.html#headers-and-content-types

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/ab983b0b6aee650c4fabed92c7cc76194299f2bf..1b5d86854fb7a29652dd7d5428466fc8957227a5
Address @EdDev comments.

tests/network/libs/cloudinit.py
```diff
@@ -51,9 +53,21 @@ class NetworkData:
...
 
 
-def cloudinit(netdata: NetworkData) -> dict[str, Any]:
-    return {NETWORK_DATA: todict(netdata)}
+@dataclass
+class UserData:
+    users: list
+
+
+def todict(obj: Any) -> dict[str, Any]:
```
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/libs/cloudinit.py
sbahar619: Done.

tests/network/localnet/test_localnet.py
sbahar619: Done, thanks!

tests/network/libs/cloudinit.py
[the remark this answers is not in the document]
EdDev: Why? OO basics. Contact me offline if you need details.

I thought I gave alternatives in other comments, but here some:
- Remove the differences to treat both the same.
- Use two different functions.
- Let the caller add or change non-common things.

tests/network/localnet/test_localnet.py
sbahar619: Done.

sbahar619:
Change: rebase.

tests/network/localnet/test_localnet.py
EdDev: Ohh my... I did not expect that at all.
@myakove , @rnetser , @vsibirsk , how do we encourage typing and not the other way around?

tests/network/localnet/test_localnet.py
yossisegev: Why should the test case be updated?
The case that is automated here - CNV-10804 - is about basic VMs connectivity when their secondaries are connected via localnet. For testing it over migration we need a separate test case.

EdDev (approved):
Thank you very much!

The comments that have not been agreed or treated can be passed to a follow up.

I'm listing the ones I remember for my own records or for anyone that want to pick them up:
- mypy: https://github.com/RedHatQE/openshift-virtualization-tests/pull/464#discussion_r2057529869
- Drop `vlan_id` fixture.
- Drop `isinstance` usage from libs/cloudinit.

tests/network/localnet/test_localnet.py
EdDev: AFAIK there is a trace added on the failure and it can be easily seen.

+1 for small building blocks, but it does not have to be a fixture. In this case, there is a contextmanager.

> For example, ns -> you can have one fixture for ns creation which get the name via request.param.

I guess you mean by marking it on the caller fixture. I have no issue with this, but if a regular contextmanager can do it as well, I see not added value to make it also a fixture.
If the fixture exists already, fine, lets use it and pass the name.

tests/network/libs/cloudinit.py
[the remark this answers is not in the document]
rnetser: why not list[Any]?
it will also be good to include the link as a comment

libs/net/vmspec.py
```diff
@@ -78,3 +79,19 @@ def lookup_primary_network(vm: BaseVirtualMachine) -> Network:
...
     raise VMInterfaceSpecNotFoundError(f"No interface connected to the primary network was found in VM {vm.name}.")
+
+
+def add_network_interface(vmi_spec: VMISpec, network: Network, interface: Interface) -> None:
+    vmi_spec.networks = vmi_spec.networks or []
+    vmi_spec.networks.append(network)
+    vmi_spec.domain.devices = vmi_spec.domain.devices or Devices()
+    vmi_spec.domain.devices.interfaces = vmi_spec.domain.devices.interfaces or []
+    vmi_spec.domain.devices.interfaces.append(interface)
+
```
rnetser, line 90: isn't `return` missing?
what am i missing?

tests/network/localnet/test_localnet.py
rnetser: i'm ok with tests files not having typing.
i am not in favor of having everything needed for a test in the test file. i.e functions should be in a separate file and fixtures in another  and those should be checked for typing.

libs/net/vmspec.py
EdDev: It is added to the inputted spec, e.g. `vmi_spec.domain.devices.interfaces.append(interface)`.
Maybe we should also return the `vmi_spec` to make it clearer.

IMO it can be done as a follow up, but up to you and the author.

sbahar619: The function operates directly on the object instance pointer, which is why the function name is prefixed with add_ rather than get_.
Adding a return statement would shift the function’s purpose and potentially cause confusion, as it would no longer reflect a modification-oriented operation.

tests/network/localnet/test_localnet.py
Anatw: Oh, I missed that. I thought we agreed on automating CNV-11775 as the first scenario. Good catch @yossisegev!

libs/net/vmspec.py
EdDev: > Adding a return statement would shift the function’s purpose and potentially cause confusion, as it would no longer reflect a modification-oriented operation.

It will clarify that the output is a new updated spec.
This is a known pattern, but there is no hard rule/policy AFAIK.

sbahar619: Right, but it means adding a return statement without usage for the sake of clarity, I prefer not adding it, but I will not insist.

tests/network/localnet/test_localnet.py
Anatw: Shahaf, 10804 is inactive, please change the Polarion to CNV-11775.

tests/network/libs/cloudinit.py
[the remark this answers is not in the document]
sbahar619: Done.

tests/network/libs/cloudinit.py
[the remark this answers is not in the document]
sbahar619: I added it a a comment.

tests/network/localnet/test_localnet.py
sbahar619: Thanks I updated the correct Polarion ID.

sbahar619:
> Thank you very much!
> 
> The comments that have not been agreed or treated can be passed to a follow up.
> 
> I'm listing the ones I remember for my own records or for anyone that want to pick them up:
> 
>     * mypy: [ Add test for connectivity between localnet VMs #464 (comment)](https://github.com/RedHatQE/openshift-virtualization-tests/pull/464#discussion_r2057529869)
> 
>     * Drop `vlan_id` fixture.
> 
>     * Drop `isinstance` usage from libs/cloudinit.

I dropped isinstance.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-464 published

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,157 @@
...
+def nad_localnet_2(ns_localnet_2, vlan_id):
+    with localnet_nad(namespace=ns_localnet_2.name, name="test-localnet-nad2", vlan_id=vlan_id) as nad:
+        yield nad
+
+
+@pytest.fixture(scope="module")
+def localnet_vms(nad_localnet_1, nad_localnet_2):
+    with (
+        localnet_vm(
+            namespace=nad_localnet_1.namespace, name="test-vm1", network=nad_localnet_1.name, cidr="10.0.0.1/24"
```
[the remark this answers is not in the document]
Anatw: We will need a const once we have more scenarios, but we can change it than. Shahaf, you can resolve this comment, thanks.

tests/network/localnet/test_localnet.py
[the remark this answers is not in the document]
sbahar619: I split the utilities and fixtures.
The ns fixture you asked seems to me too complicated compared to the alternative to keep it as it is now or just splitting the fixture to ns and nad.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/afbb37b500c8310070c63ed4ee3070d0b7ecfddd..b467a5e5379dee09951f95ec25caf13d5fd1753b
Address @EdDev comment.

sbahar619:
Change: rebase.

yossisegev (approved):

tests/network/localnet/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+        yield nncp
+
+
+@pytest.fixture(scope="module")
+def vlan_id(vlan_index_number: Iterator[int]) -> int:
+    return next(vlan_index_number)
+
+
+@contextmanager
+def create_namespace(name: str) -> Iterator[Namespace]:
```
EdDev, line 49: Make it a fixture please and use it directly to create the VMs (instead of interpreting it from the NAD).

tests/network/localnet/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+        ) as vm1,
+        localnet_vm(
+            namespace=nad_localnet_2.namespace, name="test-vm2", network=nad_localnet_2.name, cidr="10.0.0.2/24"
+        ) as vm2,
+    ):
+        vms = (vm1, vm2)
+        for vm in vms:
+            vm.start()
+        for vm in vms:
+            vm.vmi.wait_for_condition(
```
EdDev, line 84: We agreed some time ago that we want to do this in two stages: Wait for it to be ready (`wait_for_ready_status`) and then wait for the agent-connected, just to make it easier to debug in case of problems.

EdDev (commented):
Very nice, thank you.
Some minor comments inline.

tests/network/localnet/conftest.py
sbahar619: Done.

tests/network/localnet/conftest.py
sbahar619: Done.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/008b919d978db97c920d9ac8bb6f2426009e7f3a..9b8e439bee03b5c46bd1a821ede4bf81dd639257
Address @EdDev comments.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-464 published

EdDev (approved):
Thanks!

/lgtm
/approve

tests/network/localnet/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+from utilities.constants import (
+    WORKER_NODE_LABEL_KEY,
+)
+from utilities.infra import create_ns
+
+_IPERF_SERVER_PORT = 5201
+
+
+@pytest.fixture(scope="module")
+def nncp_ovn_bridge_mapping() -> Iterator[libnncp.NodeNetworkConfigurationPolicy]:
```
rnetser, line 22: nitpick - `yield` returns a Generator

tests/network/localnet/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@pytest.fixture(scope="module")
+def vlan_id(vlan_index_number: Iterator[int]) -> int:
+    return next(vlan_index_number)
+
+
+@pytest.fixture(scope="module")
+def namespace_localnet_1() -> Iterator[Namespace]:
+    yield from create_ns(name="test-localnet-ns1")  # type: ignore
```
rnetser, line 49: why do you need ignore typing here?

tests/network/localnet/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+            namespace=nad_localnet_2.namespace, name="test-vm2", network=nad_localnet_2.name, cidr="10.0.0.2/24"
+        ) as vm2,
+    ):
+        vms = (vm1, vm2)
+        for vm in vms:
+            vm.start()
+        for vm in vms:
+            vm.wait_for_ready_status(status=True)
+            vm.vmi.wait_for_condition(
+                condition=Resource.Condition.Type.AGENT_CONNECTED, status=Resource.Condition.Status.TRUE
```
rnetser, line 88:
```suggestion
                condition=vm.Condition.Type.AGENT_CONNECTED, status=vm.Condition.Status.TRUE
```
(removes extra import that is not needed)

tests/network/localnet/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+    nad_localnet_2: netattachdef.NetworkAttachmentDefinition,
+) -> Iterator[tuple[BaseVirtualMachine, BaseVirtualMachine]]:
+    with (
+        localnet_vm(
+            namespace=nad_localnet_1.namespace, name="test-vm1", network=nad_localnet_1.name, cidr="10.0.0.1/24"
+        ) as vm1,
+        localnet_vm(
+            namespace=nad_localnet_2.namespace, name="test-vm2", network=nad_localnet_2.name, cidr="10.0.0.2/24"
+        ) as vm2,
+    ):
```
rnetser, line 81: i'd suggest splitting to:
- 2 fixtures which create each VM (this way, if one VM creation fails, you know where the failure is, making triaging easier)
- `localnet_vms` that does that start etc

tests/network/localnet/liblocalnet.py
```diff
@@ -0,0 +1,56 @@
+from libs.net import netattachdef
+from libs.net.vmspec import add_network_interface, add_volume_disk
+from libs.vm.affinity import new_pod_anti_affinity
+from libs.vm.factory import base_vmspec, fedora_vm
+from libs.vm.spec import CloudInitNoCloud, Interface, Metadata, Multus, Network
+from libs.vm.vm import BaseVirtualMachine, cloudinitdisk_storage
+from tests.network.libs import cloudinit
+
+LOCALNET_TEST_LABEL = {"test": "localnet"}
```
rnetser, line 9: used in one function, should be defined there

rnetser (changes requested):

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from libs.net.traffic_generator import is_tcp_connection
+from utilities.virt import migrate_vm_and_verify
+
+
+@pytest.mark.ipv4
+@pytest.mark.polarion("CNV-11775")
+def test_connectivity_between_localnet_vms(nncp_ovn_bridge_mapping, localnet_server, localnet_client):
```
Anatw, line 9: I suggest renaming it to be aligned with the test - test_connectivity_over_migration_between_localnet_vms

tests/network/localnet/conftest.py
sbahar619: Because:
tests/network/localnet/conftest.py:49: error: Call to untyped function "create_ns" in typed context  [no-untyped-call]
Which means adding type annotation to a utility function outside the scope of my PR, so to keep the PR in scope, I decided to ignore typing here.

tests/network/localnet/conftest.py
sbahar619: I already had this kind of version you suggested, but @EdDev comment to use this version:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/464#discussion_r2052133873

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from libs.net.traffic_generator import is_tcp_connection
+from utilities.virt import migrate_vm_and_verify
+
+
+@pytest.mark.ipv4
```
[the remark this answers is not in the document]
sbahar619: The PR is not merged yet:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/724

tests/network/localnet/conftest.py
sbahar619: Done.

tests/network/localnet/conftest.py
sbahar619: Done.

tests/network/localnet/liblocalnet.py
sbahar619: Done.

tests/network/localnet/test_localnet.py
sbahar619: Done.

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/9b8e439bee03b5c46bd1a821ede4bf81dd639257..d58701e01cf04f5119ed636c0efaacf4ceb76cc1
Address @rnetser and @Anatw comments.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-464 published

tests/network/localnet/conftest.py
rnetser: had this discussion with @EdDev ; please split.

sbahar619: Done.

rnetser (dismissed):

sbahar619:
Change: https://github.com/RedHatQE/openshift-virtualization-tests/compare/9c7dfc29bd659b8654a4ad06df42f49373eb4912..f406bbc870ae92564c0f6991bbb7cfa9f09bd423
Minor refactor, renaming and reordering.
@EdDev @rnetser

sbahar619:
Change: rebase.

redhat-qe-bot2:
No build-and-push-container configured for this repository

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-464 published

tests/network/localnet/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+        name="test-localnet-nncp",
+        desired_state=desired_state,
+        node_selector={WORKER_NODE_LABEL_KEY: ""},
+    ) as nncp:
+        nncp.wait_for_status_success()
+        yield nncp
+
+
+@pytest.fixture(scope="module")
+def namespace_localnet_1() -> Generator[Namespace]:
```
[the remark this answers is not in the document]
EdDev: I guess we are suffering from this fixture hell which I do not seem to manage and take down.
Due to the way fixtures are referenced, a general name can collide with other future fixtures.

Do not change it.

tests/network/localnet/test_localnet.py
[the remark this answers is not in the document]
Anatw: Ack.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-464 published

EdDev (approved):
Thank you.

/approve
/lgtm

Passed verification on CI: `openshift-virtualization-tests-runner/2226`

/verified

EdDev:
> Passed verification on CI: `openshift-virtualization-tests-runner/2226`
> 
> /verified

/verified

sbahar619:
Change: rebase.

tests/network/localnet/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+from utilities.constants import (
+    WORKER_NODE_LABEL_KEY,
+)
+from utilities.infra import create_ns
+
+_IPERF_SERVER_PORT = 5201
+
+
+@pytest.fixture(scope="module")
+def nncp_ovn_bridge_mapping() -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
```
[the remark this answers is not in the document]
sbahar619: https://github.com/RedHatQE/openshift-virtualization-tests/pull/464#discussion_r2047093581

tests/network/localnet/conftest.py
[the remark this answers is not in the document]
sbahar619: Ack.

tests/network/localnet/test_localnet.py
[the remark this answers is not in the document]
sbahar619: It's not the scope of this PR, and if I will implement it here, and it will be merged before:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/724
I will cause conflicts and take away the core context of this PR.
Let's be patient.

tests/network/localnet/conftest.py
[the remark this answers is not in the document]
sbahar619: Done.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-464 published

EdDev (approved):
/approve

tests/network/localnet/test_localnet.py
```diff
@@ -0,0 +1,11 @@
...
+@pytest.mark.polarion("CNV-11775")
+def test_connectivity_over_migration_between_localnet_vms(nncp_localnet, localnet_server, localnet_client):
```
Anatw, line 9: I think it might be confusing that the nncp is called here but not used and I'm afraid that in the future it can be mistakenly removed from here.
I think it should be called to in the NAD and not in the test. I'll do that in a follow up PR if agreed.

Anatw (approved):
Thank you!

rnetser (approved):

vsibirsk (approved):

yossisegev (approved):

PR #511, opened 2025-03-17
Update default storage class setup for upgrade

hmeir:
##### Short description:
Currently, when we are switching to ocs-virt sc before an upgrade, we are deleting the bootsources as well.
For 4.18+, switching the sc should be enough.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-56878

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -187,10 +187,10 @@ from utilities.storage import (
     get_storage_class_with_specified_volume_mode,
     get_test_artifact_server_url,
     is_snapshot_supported_by_sc,
-    re_import_boot_sources,
     remove_default_storage_classes,
     sc_is_hpp_with_immediate_volume_binding,
     update_default_sc,
+    verify_boot_sources_reimported,
 )
 from utilities.virt import (
     VirtualMachineForCloning,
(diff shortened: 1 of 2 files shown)


redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-511 published

tests/conftest.py
```diff
@@ -2665,22 +2665,22 @@ def updated_default_storage_class_ocs_virt(
...
                     namespace=golden_images_namespace.name,
                 )
-                if not failed_update_boot_sources:
+                if boot_source_imported_successfully:
                     yield
 
-        # on teardown, delete the boot sources and wait for the original sources to re-create
-        re_import_boot_sources(
+        # on teardown, wait for the original sources to re-create
+        verify_boot_sources_reimported(
```
vsibirsk, line 2679: what if they fail to restore?

jpeimer: yeah, maybe we should save the return value here?
`boot_source_imported_successfully = verify_boot_sources_reimported()`

tests/conftest.py
```diff
@@ -2665,22 +2665,22 @@ def updated_default_storage_class_ocs_virt(
         )
         == "false"
     ):
-        failed_update_boot_sources = False
+        boot_source_imported_successfully = False
         with remove_default_storage_classes(cluster_storage_classes=cluster_storage_classes):
             with update_default_sc(default=True, storage_class=ocs_storage_class):
-                failed_update_boot_sources = re_import_boot_sources(
```
geetikakay, line 2671: if update_default_sc fails , i know chances are usually less , then there will be no default sc left in cluster?

jpeimer: if there will be no default sc in the cluster - boot source images will not be ready, so it will fail anyway

geetikakay: but if we catch early we will reduce wait time and fail right away.? wdyt

tests/conftest.py
hmeir: I dont think that we should take specific action.
If it fails also on teardown, we have the logger error. we pytest.exit in this case, so not much we can do

jpeimer: So in line 2683 we say `if not boot_source_imported_successfully: exit`

but we do not update the value of the `boot_source_imported_successfully`, so if it was False before yiled, but succeeded on teardown, we will not catch it, and will exit anyway

jpeimer (changes requested):

tests/conftest.py
jpeimer: Oh okay, if we want to exit anyway, we are doing that.

jpeimer: But if on teardown they failed to restore, we will not exit. 
We will only catch that if the cluster will run a health check afterwards. 
(that's how it worked before this PR as well, so I'm okay with keeping it this way for now)

jpeimer (approved):

dshchedr (changes requested):
we should create patch for main branch first, no?

jpeimer:
> we should create patch for main branch first, no?
@dshchedr For upgrade, when there's no way to actually verify the patch in main, it's okay to merge in main-1 and then port-forward

dshchedr (approved):

cnv-qe-bot1:
Cherry-picked PR Update default storage class setup for upgrade into main

PR #478, opened 2025-03-16
Add jumbo_frame marker and mark related tests

sbahar619:
##### Short description:
Add jumbo_frame marker and mark related tests
##### More details:
Add jumbo_frame marker and mark related tests
    
    Added a new jumbo_frame pytest marker to categorize tests
    requiring network configurations that support jumbo frames.
    
    Marked existing jumbo frame related tests with the jumbo_frame
    marker for better classification. Also, marked the jumbo frame
    tests with the special_infra marker as they require infrastructure
    network NICs to support jumbo frames.

PR #525, opened 2025-03-18
Optimize session item access in network_sanity

sbahar619:
##### Short description:
Optimize session item access in network_sanity
##### More details:
Optimize session item access in network_sanity
    
    Avoid repeated calls to request.session.items by storing the result in a
    variable. This reduces redundant API calls and improves
    performance when verifying network test conditions.

PR #529, opened 2025-03-19
Validate jumbo frame support in network sanity check

sbahar619:
##### Short description:
Validate jumbo frame support in network sanity check
##### More details:
Validate jumbo frame support in network sanity check
    
    Added a check to ensure the cluster supports jumbo frame tests.
    If the network MTU is too low, a failure message is recorded.
    This prevents tests from running when the network is not capable
    of handling jumbo frames.
This PR is based on: #478

### DIFF
--- a/tests/network/conftest.py
+++ b/tests/network/conftest.py
@@ -200,7 +200,14 @@ def ovn_kubernetes_cluster(admin_client):
 
 
 @pytest.fixture(scope="session", autouse=True)
-def network_sanity(hosts_common_available_ports, junitxml_plugin, request, istio_system_namespace):
+def network_sanity(
+    hosts_common_available_ports,
+    junitxml_plugin,
+    request,
+    istio_system_namespace,
+    cluster_network_mtu,
+    network_overhead,
+):
     """
     Ensures the test cluster meets network requirements before executing tests.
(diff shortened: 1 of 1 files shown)


tests/network/conftest.py
```diff
@@ -242,9 +249,22 @@ def _verify_service_mesh():
                     f"'{ISTIO_SYSTEM_DEFAULT_NS}' namespace"
                 )
 
+    def _verify_jumbo_frame():
+        if any(item.get_closest_marker("jumbo_frame") for item in request.session.items):
```
vsibirsk, line 253: this should be aligned now with:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/525/files#diff-553c5bd873d0bdf44449163e92ab12a9e2d440f4ce07d111f4830959e88f80d6R222

vsibirsk (changes requested):

tests/network/conftest.py
sbahar619: Thanks, done.

vsibirsk (approved):

cnv-qe-bot1:
Cherry-picked PR Validate jumbo frame support in network sanity check into cnv-4.18

PR #586, opened 2025-03-25
Split pod network connectivity test per IP family

yossisegev:
Towards stabilizing network tier-2 tests, and specifically the gating ones, this PR
- splits the pod network connectivity test to IPv4 and IPv6 cases
- marks the IPv6 tests with the ticket that trackes this limitation.

### DIFF
--- a/tests/network/connectivity/test_pod_network.py
+++ b/tests/network/connectivity/test_pod_network.py
@@ -4,6 +4,7 @@ VM to VM connectivity
 
 import pytest
 
+from utilities.constants import IPV4_STR, IPV6_STR
 from utilities.infra import get_node_selector_dict
 from utilities.network import (
     compose_cloud_init_data_dict,
@@ -73,12 +74,30 @@ def cloud_init_ipv6_network_data(dual_stack_network_data):
     return compose_cloud_init_data_dict(ipv6_network_data=dual_stack_network_data)
 
 
+@pytest.mark.parametrize(
+    "ip_family",
+    [
+        pytest.param(
+            IPV4_STR,
+            marks=[
+                pytest.mark.polarion("CNV-2332"),
+                pytest.mark.ipv4,
+            ],
+        ),
+        pytest.param(
+            IPV6_STR,
+            marks=[
+                pytest.mark.polarion("CNV-11845"),
+                pytest.mark.ipv6,
+                pytest.mark.jira("CNV-58529", run=True),
+            ],
+        ),
+    ],
+    indirect=False,
+)
 @pytest.mark.gating
-@pytest.mark.polarion("CNV-2332")
 def test_connectivity_over_pod_network(
-    fail_if_not_ipv4_supported_cluster_from_mtx,
-    fail_if_not_ipv6_supported_cluster_from_mtx,
-    ip_stack_version_matrix__module__,
+    ip_family,
     pod_net_vma,
     pod_net_vmb,
     pod_net_running_vma,
@@ -88,7 +107,7 @@ def test_connectivity_over_pod_network(
     """
     Check connectivity
     """
-    dst_ip = get_ip_from_vm_or_virt_handler_pod(family=ip_stack_version_matrix__module__, vm=pod_net_running_vmb)
+    dst_ip = get_ip_from_vm_or_virt_handler_pod(family=ip_family, vm=pod_net_running_vmb)
     assert dst_ip, f"Cannot get valid IP address from {pod_net_running_vmb.vmi.name}."
 
     ping_cmd = f"ping -c 3 {dst_ip}"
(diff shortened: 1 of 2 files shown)


tests/network/connectivity/test_pod_network.py
```diff
@@ -88,8 +87,28 @@ def test_connectivity_over_pod_network(
...
+        src_vm=pod_net_running_vma,
+        dst_vm=pod_net_running_vmb,
+    )
+
 
-    ping_cmd = f"ping -c 3 {dst_ip}"
-    vm_console_run_commands(vm=pod_net_running_vma, commands=[ping_cmd])
+@pytest.mark.gating
+@pytest.mark.polarion("CNV-11845")
+def test_connectivity_over_pod_ipv6_network(
```
EdDev, line 99: I think you can just use parametrize [1] , it is used frequently [2] and fits here well.
The relevant parametrize input is `ip_family`.

I'm also not clear what is the purpose of `fail_if_not_ipv4_supported_cluster_from_mtx` and its IPv6 counterpart.

[1] https://docs.pytest.org/en/stable/how-to/parametrize.html#parametrize
[2] https://github.com/RedHatQE/openshift-virtualization-tests/blob/4f8d53ca767a9c9813b0240748145aedc7df3748/tests/virt/cluster/vm_lifecycle/test_vm_run_strategy.py#L297

yossisegev: > I think you can just use parametrize [1] , it is used frequently [2] and fits here well. The relevant parametrize input is `ip_family`.

Because keeping these cases under the same test, which is marked as gating, necessarily means that the IPv6 cases will be executed as part of gating - but will always fail when IPv6 is not supported (i.e. in PSI).

> I'm also not clear what is the purpose of `fail_if_not_ipv4_supported_cluster_from_mtx` and its IPv6 counterpart.

These are used to make sure each of these tests run on a cluster that can support them, i.e. IPv6 test must be executed on a cluster that supports IPv6, and same for IPv4; otherwise - these tests should fail.

EdDev: > > I think you can just use parametrize [1] , it is used frequently [2] and fits here well. The relevant parametrize input is `ip_family`.
> 
> Because keeping these cases under the same test, which is marked as gating, necessarily means that the IPv6 cases will be executed as part of gating - but will always fail when IPv6 is not supported (i.e. in PSI).

Parametrize means that they are treated as multiple tests, and you can mark each per what you need.

> > I'm also not clear what is the purpose of `fail_if_not_ipv4_supported_cluster_from_mtx` and its IPv6 counterpart.
> 
> These are used to make sure each of these tests run on a cluster that can support them, i.e. IPv6 test must be executed on a cluster that supports IPv6, and same for IPv4; otherwise - these tests should fail.

I suggest to follow the dpdk sanity check [1], checking such things at the session level before running any test.
If the test is collected, the validation is per the marker.
If the cluster is wrong, no tests will run and therefore it will be clear that the problem is infrastructure and not a test problem.

[1] https://github.com/RedHatQE/openshift-virtualization-tests/blob/c25cc5009496ed3308f2af7ae31f51c2c7ae955b/tests/network/conftest.py#L221

tests/network/connectivity/test_pod_network.py
```diff
@@ -75,10 +76,8 @@ def cloud_init_ipv6_network_data(dual_stack_network_data):
 
 @pytest.mark.gating
 @pytest.mark.polarion("CNV-2332")
-def test_connectivity_over_pod_network(
+def test_connectivity_over_pod_ipv4_network(
```
geetikakay, line 79: do you think marker approach will work in your case ipv4,ipv6, dual?

yossisegev: Done.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-586 published

EdDev (approved):
Very elegant and clean, thank you!

EdDev:
Something seems wrong but I do not understand what it is: https://results.pre-commit.ci/run/github/716108424/1742998439.Zu8ysiyoQ3ShuA9DIjbBbg

EdDev:
> Something seems wrong but I do not understand what it is: https://results.pre-commit.ci/run/github/716108424/1742998439.Zu8ysiyoQ3ShuA9DIjbBbg

@myakove , can you have a look?

myakove:
> > Something seems wrong but I do not understand what it is: https://results.pre-commit.ci/run/github/716108424/1742998439.Zu8ysiyoQ3ShuA9DIjbBbg
> 
> @myakove , can you have a look?

Fixed in flake8-plugins.git@v0.0.10
https://github.com/RedHatQE/openshift-virtualization-tests/pull/601

dshchedr (approved):

sbahar619 (approved):
Looks good, thanks!

EdDev:
pre-commit.ci run

myakove:
pre-commit.ci run

PR #651, opened 2025-03-31
[4.18] Resolve IPv6 gating tests in non-supporting clusters

yossisegev:
Towards stabilizing network tier-2 tests, and specifically the gating ones, this PR
- splits the pod network connectivity test to IPv4 and IPv6 cases
- marks the IPv6 tests with the ticket that tracks this limitation.

### DIFF
--- a/tests/network/connectivity/test_ovs_linux_bridge.py
+++ b/tests/network/connectivity/test_ovs_linux_bridge.py
@@ -50,6 +50,7 @@ class TestConnectivityLinuxBridge:
     @pytest.mark.post_upgrade
     @pytest.mark.polarion("CNV-11125")
     @pytest.mark.ipv6
+    @pytest.mark.jira("CNV-58530", run=True)
     def test_ipv6_linux_bridge(
         self,
         fail_if_not_ipv6_supported_cluster,
@@ -144,6 +145,7 @@ class TestConnectivityOVSBridge:
     @pytest.mark.post_upgrade
     @pytest.mark.polarion("CNV-11128")
     @pytest.mark.ipv6
+    @pytest.mark.jira("CNV-58530", run=True)
     def test_ipv6_ovs_bridge(
         self,
         fail_if_not_ipv6_supported_cluster,
--- a/tests/network/connectivity/test_pod_network.py
+++ b/tests/network/connectivity/test_pod_network.py
@@ -4,6 +4,7 @@ VM to VM connectivity
 
 import pytest
 
+from utilities.constants import IPV4_STR, IPV6_STR
 from utilities.infra import get_node_selector_dict
 from utilities.network import (
     compose_cloud_init_data_dict,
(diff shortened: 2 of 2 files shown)


tests/network/connectivity/test_ovs_linux_bridge.py
```diff
@@ -50,6 +50,7 @@ def test_ipv4_linux_bridge(
     @pytest.mark.post_upgrade
     @pytest.mark.polarion("CNV-11125")
     @pytest.mark.ipv6
+    @pytest.mark.jira("CNV-58530", run=True)
```
jpeimer, line 53: Shouldn't it be `run=False`?

yossisegev: No, I'll explain:
IPv6 is supported on bare-metal, so we want these tests to run on these clusters and expect them to pass.
On virtual clusters (PSI) IPv6 is not supported, so marking these tests with `jira` and let them run means xailing them.

vsibirsk: so since IPv6 can't run on PSI, you'll need to remove `gating` marker here (as gating is ran on PSI only now)

tests/network/connectivity/test_pod_network.py
```diff
@@ -73,12 +74,30 @@ def cloud_init_ipv6_network_data(dual_stack_network_data):
...
+            marks=[
+                pytest.mark.polarion("CNV-11845"),
+                pytest.mark.ipv6,
+                pytest.mark.jira("CNV-58530 ", run=True),
+            ],
+        ),
+    ],
+    indirect=False,
+)
 @pytest.mark.gating
```
vsibirsk, line 98: ditto here, only v4 part should have gating marker

vsibirsk (changes requested):

tests/network/connectivity/test_ovs_linux_bridge.py
yossisegev: > so since IPv6 can't run on PSI, you'll need to remove `gating` marker here (as gating is ran on PSI only now)

@vsibirsk IPv6 support becomes a must, so that's why we issued https://issues.redhat.com/browse/CNV-56595 to limit gating tests only for bare-metals. You can go over the issue description and comments to understand why running gating only on PSI is no longer legit (if something is not clear, please reach out to me and I'll happily explain).
Just like this PR was [merged in main](https://github.com/RedHatQE/openshift-virtualization-tests/pull/586/files), it can and should be merged in 4.18 as well.
Thank you

tests/network/connectivity/test_ovs_linux_bridge.py
```diff
@@ -144,6 +145,7 @@ def test_ipv4_ovs_bridge(
     @pytest.mark.post_upgrade
     @pytest.mark.polarion("CNV-11128")
     @pytest.mark.ipv6
+    @pytest.mark.jira("CNV-58530", run=True)
```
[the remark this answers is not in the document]
yossisegev: Answered above

tests/network/connectivity/test_pod_network.py
yossisegev: Answered above

tests/network/connectivity/test_ovs_linux_bridge.py
vsibirsk: ack. wasn't aware of that network task. thanks for the clarification @yossisegev

vsibirsk (approved):

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:cnv-4.18 published

PR #669, opened 2025-04-02
Automate Fedora VM build for the tests

EdDev:
##### Short description:
Automate Fedora VM build for the tests

Preparing the Fedora VM image for the tests is currently carried out manually.
This is problematic as the human factor here presents a high risk (for mistakes.

This change aims to automate the process.

As a follow up, consider:
- Prepare arm64.
- Push the container image artifact to production quay registry on merge.

## Summary by CodeRabbit

- **Chores**
  - Introduced an automated workflow to build and push Fedora-based container images, supporting manual and pull request triggers.
  - Improved container build script to handle virtualization type selection and Fedora version compatibility more robustly.
  - Enhanced security in build logs by limiting sensitive output exposure during password handling.
  - Updated cloud-init configuration to simplify command execution by removing unnecessary privilege escalation steps.

### DIFF
--- a/.github/workflows/component-builder.yml
+++ b/.github/workflows/component-builder.yml
@@ -1,0 +1,51 @@
+name: "Component Builder"
+on:
+  pull_request_target:
+    paths:
+      - 'containers/fedora/**'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  guest-fedora-amd64:
+    runs-on: ubuntu-latest
+    env:
+      FEDORA_IMAGE: Fedora-Cloud-Base-Generic-41-1.4.x86_64.qcow2
+      FEDORA_VERSION: 41
+      CPU_ARCH: amd64
+      FULL_EMULATION: "true"
+      DEBIAN_FRONTEND: noninteractive
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Install dependencies for VM build
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            qemu-system-x86 \
+            libvirt-daemon-system \
+            virtinst cloud-image-utils \
+            libguestfs-tools
+      - name: Tweak hosted runner to enable 'virt-sysprep'
+        # https://bugs.launchpad.net/ubuntu/+source/linux/+bug/759725
+        run: sudo chmod 0644 /boot/vmlinuz*
+      - name: Fetch base Fedora image
+        working-directory: ./containers/fedora
+        run: wget -q "https://download.fedoraproject.org/pub/fedora/linux/releases/41/Cloud/x86_64/images/${{ env.FEDORA_IMAGE }}"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "0.6.11"
+      - name: Create VM
+        working-directory: ./containers/fedora
+        env:
+          PYTHONPATH: ${{ github.workspace }}
+          ORGANIZATION_ID: ${{ secrets.BITWARDEN_ORGANIZATION_ID }}
+          ACCESS_TOKEN: ${{ secrets.BITWARDEN_ACCESS_TOKEN }}
+        run: ./build.sh
+      - name: Report artifacts
+        working-directory: ./containers/fedora
+        run: |
+          ls -la
+          podman images
--- a/.github/workflows/component-builder.yml
+++ b/.github/workflows/component-builder.yml
@@ -44,8 +44,16 @@ jobs:
           ORGANIZATION_ID: ${{ secrets.BITWARDEN_ORGANIZATION_ID }}
           ACCESS_TOKEN: ${{ secrets.BITWARDEN_ACCESS_TOKEN }}
         run: ./build.sh
-      - name: Report artifacts
-        working-directory: ./containers/fedora
+      - name: Logging to quay.io
+        run: podman login -u ${{ secrets.QUAY_USER }} -p ${{ secrets.QUAY_TOKEN }} quay.io
+      - name: Tag & Push image to staging
+        env:
+          local_repository: "localhost/fedora"
+          remote_repository: "quay.io/openshift-cnv/qe-cnv-tests-fedora-staging"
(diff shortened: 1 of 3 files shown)


myakove (approved):

EdDev:
Verified on my fork: https://github.com/EdDev/openshift-virtualization-tests/actions/runs/14236588937/job/39896941625

/verified

jpeimer (dismissed):

EdDev:
[change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/34dc97379f3b9a42424275c0444c410e8b670419..fc9366a9f6f79d4d0f93fd133348cd34989386b4): Limit execution on PRs that change the `conteiners/fedora` folder.

.github/workflows/component-builder.yml
yossisegev, line 1: Where/how is this resource used?

.github/workflows/component-builder.yml
```diff
@@ -0,0 +1,59 @@
...
+      - 'containers/fedora/**'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  guest-fedora-amd64:
+    runs-on: ubuntu-latest
+    env:
+      FEDORA_IMAGE: Fedora-Cloud-Base-Generic-41-1.4.x86_64.qcow2
```
yossisegev, line 14: I think that the resource for building an image shouldn't be hard-coded and limited to a specific version (41).
You probably should take it from [fedora_os_matrix](https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/tests/global_config.py#L363) (via [FEDORA_LATEST](https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/tests/os_params.py#L27)).

containers/fedora/build.sh
```diff
@@ -35,13 +37,26 @@ else
...
 sed -i "s/password: $PASSWORD_PLACEHOLDER/password: $FEDORA_PASSWORD/" user-data
+set -x
 
 echo "Create cloud-init user data ISO"
 cloud-localds $CLOUD_INIT_ISO user-data
 
+# When running on CI, the latest Fedora may not be yet supported by virt-install
+# therefor use a lower one. It should have no influence on the result.
+OS_VARIANT=$NAME
+if [ $FEDORA_VERSION -gt 40 ]; then
```
yossisegev, line 56: Same here.
You are assuming that the latest available fedora version is  41, so when fedora advances, it will be have to be updated in several places. We better keep the parameter that should be manually changed in one place (in the matrices in [global_config.py](https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/tests/global_config.py)).

yossisegev (changes requested):

EdDev:
https://github.com/EdDev/openshift-virtualization-tests/actions/runs/14291076104/job/40052381233

/verified

.github/workflows/component-builder.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          version: "0.6.11"
+      - name: Create VM
+        working-directory: ./containers/fedora
+        env:
+          PYTHONPATH: ${{ github.workspace }}
+          ORGANIZATION_ID: ${{ secrets.BITWARDEN_ORGANIZATION_ID }}
+          ACCESS_TOKEN: ${{ secrets.BITWARDEN_ACCESS_TOKEN }}
+        run: ./build.sh
+      - name: Logging to quay.io
```
Anatw, line 47: I think we should add a verification step before pushing the image, to verify tat the build was successful. We should prevent unhealthy images to be pushed to Quay.

EdDev: You will have to explain what that means exactly and suggest what you expect to do.
First, it pushed to the staging repo, not to the production one, allowing verification against the staging before it is manually pushed to production. In fact, the follow up possible steps express it in the PR description.

myakove: I think another action in pr merged to push to prod.

While the pr is open you will have the image on stage to verify it, once verified the pr is merged then the second action push it to prod.

WDYT?

EdDev: Yes, that makes sense.
But that will go all the way to automate it from change to deployment.
I think it will be safer to start with this one, then add `arm64` support and only everything is good, to add the second job.

BTW, it will be interesting if we can pass the container images between jobs or cache them.

rnetser (dismissed):

.github/workflows/component-builder.yml
yossisegev: But you do set `FEDORA_VERSION` below in line 15. So why not setting it before, and use it dynamically?
```
FEDORA_IMAGE: Fedora-Cloud-Base-Generic-${{ env.FEDORA_IMAGE }}-1.4.x86_64.qcow2
```
or something like that?

yossisegev (changes requested):

.github/workflows/component-builder.yml
EdDev: The line you commented on is the source of the Fedora cloud image.
The `FEDORA_VERSION` below is used to name the different files and tags.

They could be linked, but it is the other way around: Parse the image name from the URL provided to the `FEDORA_IMAGE`, extract the major version and use it as the `FEDORA_VERSION`.

yossisegev: OK, I think I see what you mean here.
Thanks for the explanation.

yossisegev (approved):

EdDev:
Verify on the forked repo: https://github.com/EdDev/openshift-virtualization-tests/actions/runs/15095444595

containers/fedora/build.sh
```diff
@@ -43,13 +45,26 @@ case "$CPU_ARCH" in
 	;;
 esac
 
+if [ $FULL_EMULATION = "true" ]; then
+    VIRT_TYPE="qemu"
+fi
```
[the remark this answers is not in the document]
EdDev: Done

EdDev:
> Verify on the forked repo: https://github.com/EdDev/openshift-virtualization-tests/actions/runs/15095444595

/verified

cnv-qe-bot1:

PR #705, opened 2025-04-07
 Must gather fix flags and node role

hmeir:
##### Short description:
Fixing two issues with must gather.

##### More details:
1. Must gather collected both at failure point & IUO tests.But each time we run must-gather, it results in this message:

> error: Warning: spec.nodeSelector[node-role.kubernetes.io/master]: use "node-role.kubernetes.io/control-plane" instead

**Solution:** We should use control-plane selector.
can be done by passing specific node name/ setting up a node selector.

2. Currently, --since-time and --timeout flags are passed after "–" options seperator:

> oc adm must-gather --dest-dir=$PWD/regular --image=registry.redhat.io/container-native-virtualization/cnv-must-gather-rhel9@sha256:520577044baddc484233e22feede7c67d72b18958ac36811367ee58c4ffe2a0f -- --since=0s --timeout=900s  /usr/bin/gather --vm-details

**Solution**: They should be passed before the "–". both for IUO tests & infra must gather collection via failure.

1. Infra - each must gather collected on failure prints 10000 lines of output
2. Must gather IUO tests might have false positive.

### DIFF
--- a/tests/install_upgrade_operators/must_gather/conftest.py
+++ b/tests/install_upgrade_operators/must_gather/conftest.py
@@ -605,7 +605,8 @@ def collected_vm_details_must_gather_from_vm_node(
     yield collect_must_gather(
         must_gather_tmpdir=target_path,
         must_gather_image_url=must_gather_image_url,
-        flag_names=f"node-name={must_gather_vm.vmi.node.name},vms_details",
+        flag_names="vms_details",
+        node_name=must_gather_vm.vmi.node.name,
     )
     clean_up_collected_must_gather(failed=(request.session.testsfailed - before_fail_count), target_path=target_path)
 
--- a/utilities/must_gather.py
+++ b/utilities/must_gather.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import shlex
 
@@ -5,11 +6,14 @@ from pyhelper_utils.shell import run_command
 
 from utilities.constants import TIMEOUT_15MIN, TIMEOUT_20MIN
 
+LOGGER = logging.getLogger(__name__)
+
 
 def run_must_gather(
     image_url: str = "",
     target_base_dir: str = "",
     script_name: str = "",
+    node_name: str = "",
     flag_names: str = "",
     timeout: str = f"{TIMEOUT_15MIN}s",
     since: str | None = None,
@@ -22,6 +26,7 @@ def run_must_gather(
             If not specified, OpenShift's default must-gather image will be used.
         target_base_dir (str, optional): path to base directory
         script_name (str, optional): must-gather script name or path
+        node_name (str, optional): node name to run the must gather pod on.
         flag_names (str, optional): comma separated list of must-gather flags
             Examples: "oc adm must-gather --image=quay.io/kubevirt/must-gather -- /usr/bin/gather --default"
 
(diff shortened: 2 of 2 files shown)


cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-705 published

utilities/data_collector.py
```diff
@@ -101,10 +102,16 @@ def collect_default_cnv_must_gather_with_vm_gather(since_time, target_dir):
     must_gather_image = [
         image["image"] for image in cnv_csv.instance.spec.relatedImages if "must-gather" in image["name"]
     ][0]
+
+    # run on a control plane node
+    control_plane_node = list(Node.get(client=get_client(), label_selector=f"{NODE_ROLE_KUBERNETES_IO}/control-plane"))[
```
dshchedr, line 107: Probably, it would be good to check that this control-plane node is active and schedulable
if the node is not in a good shape - collection may fail

dshchedr (changes requested):

utilities/data_collector.py
hmeir: If any node not activate/in good shape - won't the sanity check fail anyway?

dshchedr: sanity is a scope=module, so it is checked at the beginning of the module.

what if the master node is broken because of some test?  

I'm not very familiar with must-gather arguments, may be you can pass list of nodes where it can be executed?

dshchedr: actually, based on the error message - I think you may need to pass `--node-selector node-role.kubernetes.io/control-plane` instead of single `--node-name`

dbasunag: This error message 
```
error: Warning: spec.nodeSelector[node-role.kubernetes.io/master]: use "node-role.kubernetes.io/control-plane" instead
```
is coming from actual must-gather. User is not required to pass a node to spin the must-gather pod. We should stick with user scenario and not work around to hide issue. . I believe I already logged a [bug](https://issues.redhat.com/browse/CNV-53368) for this warning, but it is for CNV, it should actually go to OCP. It is not specific to CNV.

dshchedr: @dbasunag not sure it will be fixed in some near future since it is a warning message..

but this patch looks useful - currently we have hundreds of lines of logs after each must-gather run which makes test logs pretty messy.

@hmeir I didn't check - may be we can simply ignore the warning message and do not print the full output?

hmeir: @dbasunag 
Yeah you are right- its general OCP must-gather issue.

@dshchedr 
The issue is with our automation that returns error upon this must gather warning.
Passing node selector will collect must gather on 3 nodes, instead of picking one.
I guess the output can be ignored like your said, i'll upload a patch for it.

hmeir: Now it looks like this:

```
13:23:41  2025-04-20T10:23:41.097616 pyhelper_utils.shell INFO Running oc adm must-gather --dest-dir=/data/tests-collected-info/install_upgrade_operators/feature_gates/test_default_featuregates/test_default_featuregates_by_resource[verify_default_featuregates_hco_cr]/pytest_exception_interact --image=registry.redhat.io/container-native-virtualization/cnv-must-gather-rhel9@sha256:ffc5d921ad3645a1471bcf1a82028af0c5d15b247c58c28f64a3e91c21cfe340 --since=37s --timeout=900s -- /usr/bin/gather --vms_details command
13:25:17  2025-04-20T10:25:10.040541 utilities.must_gather WARNING must-gather raised the following error: Warning: spec.nodeSelector[node-role.kubernetes.io/master]: use "node-role.kubernetes.io/control-plane" instead
```

jpeimer (approved):

utilities/data_collector.py
dshchedr: so, if you ignoring the output - do you still need to pass `--node-name=`? 

must-gather documentation is not very clear, I didn't get why node-name is selects the node WHERE to start pod but don't limit the list of collectable nodes, and at the same time node-selector does vice versa

I was always sure - node-selector is the same as node-name, just accepts the list instead of one node.

my main concern with passing only one node: if for some reason it is not available (lack of resources, broken, etc...) - nothing will be collected

hmeir: > so, if you ignoring the output - do you still need to pass --node-name=?

Yes for specific cases, because we have a test that is using that option(and it did wrong).
But for the general case, I keep it as before(picking automaticly good master/cp node), but ignoring the output

dshchedr (approved):

utilities/data_collector.py
dshchedr: ty!

PR #711, opened 2025-04-08
 Fix ServiceMesh tests and adjust to OSSM 3

yossisegev:
OSSM moved from ver. 2 to 3. This fix adjusts to ver. 3 by enabling gateway injection, which is performed by labeling the work namespace with the matching label {istio-injection: enabled}.

### DIFF
--- a/tests/network/service_mesh/conftest.py
+++ b/tests/network/service_mesh/conftest.py
@@ -174,6 +174,13 @@ def ns_outside_of_service_mesh(admin_client):
     yield from create_ns(admin_client=admin_client, name="outside-mesh")
 
 
+@pytest.fixture(scope="module")
+def service_mesh_tests_namespace(namespace, admin_client):
+    # The namespace used for the ServiceMesh tests must be added the `istio-injection` label.
+    label_project(name=namespace.name, label={"istio-injection": "enabled"}, admin_client=admin_client)
+    return namespace
+
+
 @pytest.fixture(scope="class")
 def httpbin_deployment_service_mesh(service_mesh_tests_namespace):
     with ServiceMeshDeployments(
@@ -415,11 +422,10 @@ def peer_authentication_strict_service_mesh(service_mesh_tests_namespace):
 @pytest.fixture(scope="class")
 def peer_authentication_service_mesh_deployment(
     istio_system_namespace,
-    service_mesh_tests_namespace,
+    peer_authentication_strict_service_mesh,
     vm_fedora_with_service_mesh_annotation,
     ns_outside_of_service_mesh,
     httpbin_service_service_mesh,
-    peer_authentication_strict_service_mesh,
     service_mesh_vm_console_connection_ready,
 ):
     wait_service_mesh_components_convergence(
@@ -435,10 +441,3 @@ def vmi_http_server(vm_fedora_with_service_mesh_annotation):
         vm=vm_fedora_with_service_mesh_annotation,
         commands=[f'while true ; do  echo -e "HTTP/1.1 200 OK\n\n $(date)" | nc -l -p {SERVICE_MESH_PORT}  ; done &'],
     )
-
-
-@pytest.fixture(scope="module")
-def service_mesh_tests_namespace(namespace, admin_client):
-    # The namespace used for the ServiceMesh tests must be added the `istio-injection` label.
-    label_project(name=namespace.name, label={"istio-injection": "enabled"}, admin_client=admin_client)
-    return namespace
--- a/tests/network/service_mesh/conftest.py
+++ b/tests/network/service_mesh/conftest.py
@@ -36,7 +36,7 @@ from tests.network.utils import (
     authentication_request,
 )
 from utilities.constants import PORT_80, TIMEOUT_4MIN, TIMEOUT_10SEC
-from utilities.infra import add_scc_to_service_account, create_ns, unique_name
+from utilities.infra import add_scc_to_service_account, create_ns, label_project, unique_name
 from utilities.virt import running_vm, vm_console_run_commands, wait_for_console
 
 LOGGER = logging.getLogger(__name__)
@@ -441,3 +441,8 @@ def vmi_http_server(vm_fedora_with_service_mesh_annotation):
         vm=vm_fedora_with_service_mesh_annotation,
         commands=[f'while true ; do  echo -e "HTTP/1.1 200 OK\n\n $(date)" | nc -l -p {SERVICE_MESH_PORT}  ; done &'],
     )
+
+
+@pytest.fixture(scope="module")
+def label_istio_injection_namespace(namespace, admin_client):
+    label_project(name=namespace.name, label={"istio-injection": "enabled"}, admin_client=admin_client)
--- a/tests/network/service_mesh/conftest.py
+++ b/tests/network/service_mesh/conftest.py
@@ -32,7 +32,6 @@ from tests.network.utils import (
     FedoraVirtualMachineForServiceMesh,
     ServiceMeshDeployments,
     ServiceMeshDeploymentService,
-    ServiceMeshMemberRollForTests,
     authentication_request,
 )
 from utilities.constants import PORT_80, TIMEOUT_4MIN, TIMEOUT_10SEC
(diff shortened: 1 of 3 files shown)


sbahar619:
What about just add the namespace label at creation time?
At create_ns function as labels argument?
If the label is meant to be permanent, so just create the namespace with the label builtin.

tests/network/service_mesh/test_service_mesh.py
```diff
@@ -6,7 +6,10 @@
 )
 from tests.network.utils import assert_authentication_request
 
-pytestmark = pytest.mark.service_mesh
+pytestmark = [
+    pytest.mark.service_mesh,
+    pytest.mark.usefixtures("label_istio_injection_namespace"),
```
EdDev, line 11: Is there a difference between setting the fixture here or on definition to specify `autouse=True`?

EdDev:
> What about just add the namespace label at creation time? At create_ns function as labels argument? If the label is meant to be permanent, so just create the namespace with the label builtin.

If you check the implementation, it cannot be done on creation unless an admin client is used.
Also, the current `namespace` fixture is not supporting labels and these will have to be added. Adding that will effect all tests and is not something you want to do as a side-effect of a simple network specific test change.

tests/network/service_mesh/conftest.py
```diff
@@ -441,3 +441,8 @@ def vmi_http_server(vm_fedora_with_service_mesh_annotation):
         vm=vm_fedora_with_service_mesh_annotation,
         commands=[f'while true ; do  echo -e "HTTP/1.1 200 OK\n\n $(date)" | nc -l -p {SERVICE_MESH_PORT}  ; done &'],
     )
+
+
+@pytest.fixture(scope="module")
+def label_istio_injection_namespace(namespace, admin_client):
```
EdDev, line 447: Maybe in the spirit of what @sbahar619 suggested, you can use this one to wrap the namespace one.
I.e. rename this to `istio_namespace`, add the label like you do here and return the namespace back.

yossisegev:
> What about just add the namespace label at creation time? At create_ns function as labels argument? If the label is meant to be permanent, so just create the namespace with the label builtin.

@sbahar619 Thank you for raising it.
That is what I originally planned, but this namespace is is created using the `namespace` fixture, like any other namespace that is created per module in this repo.
Adding the label there will necessarily affect all these namespaces, but my intention is to add this label only to this specific namespace, and I am doing it by editing it after it is already created.

tests/network/service_mesh/conftest.py
yossisegev: Sounds like an overkill to me.
Going this way will require changing all the namespace references in this module from `namespace` to `istio_namespace`, which contradicts the behavior in all the other tests in this repo.
Can you please explain the rational for doing so?

yossisegev: OK, I see the conceptual point you are raising. I'll change it.
BTW - `namespace` fixture may not be used or referenced in the test module, but it is in the package's conftest. Nevertheless I'll apply this change.

tests/network/service_mesh/test_service_mesh.py
yossisegev: Accepted.

sbahar619 (approved):
lgtm.

Anatw (approved):

tests/network/service_mesh/conftest.py
```diff
@@ -441,3 +444,10 @@ def vmi_http_server(vm_fedora_with_service_mesh_annotation):
         vm=vm_fedora_with_service_mesh_annotation,
         commands=[f'while true ; do  echo -e "HTTP/1.1 200 OK\n\n $(date)" | nc -l -p {SERVICE_MESH_PORT}  ; done &'],
     )
+
+
+@pytest.fixture(scope="module", autouse=True)
```
rnetser, line 449: 1. why not create the ns with this label in the first place? 
2. why are you using `autouse`; `autouse` usage should be minimal (if at all)

yossisegev: > why not create the ns with this label in the first place?

This was already discussed [here](https://github.com/RedHatQE/openshift-virtualization-tests/pull/711#issuecomment-2789243763) and [here](https://github.com/RedHatQE/openshift-virtualization-tests/pull/711#issuecomment-2789237204).
tl;dr: Doing so means changing the repo-wide `namespace` fixture and necessarily affect all the tests. It must remain un-touched as much as possible.
In addition - editing an existing namespace is valid (like `oc create ns new-ns` and then `oc edit ns new-ns`).

> why are you using autouse; autouse usage should be minimal (if at all)

Originally I called it with `usefixtures`, but [Edy asked to autouse it](https://github.com/RedHatQE/openshift-virtualization-tests/pull/711#discussion_r2035090945).

EdDev: What is the problem with `autouse`?
In this case, it says something like: In case any test is collected in this package, a special namespace is needed and created implicitly for all tests (and fixtures here) to use.

If the fixture is already explicitly dependent on by tests in an indirect manner (through other fixtures along the way), then it can and should be dropped.
Is this the case here?

cnv-qe-bot1:
yossisegev is not part of the approver, only approvers can mark pull request with hold

tests/network/service_mesh/conftest.py
yossisegev: autouse/usefixtures are now entirely removed.

tests/network/service_mesh/conftest.py
```diff
@@ -441,3 +435,10 @@ def vmi_http_server(vm_fedora_with_service_mesh_annotation):
         vm=vm_fedora_with_service_mesh_annotation,
         commands=[f'while true ; do  echo -e "HTTP/1.1 200 OK\n\n $(date)" | nc -l -p {SERVICE_MESH_PORT}  ; done &'],
     )
+
+
+@pytest.fixture(scope="module")
+def service_mesh_tests_namespace(namespace, admin_client):
+    # The namespace used for the ServiceMesh tests must be added the `istio-injection` label.
+    label_project(name=namespace.name, label={"istio-injection": "enabled"}, admin_client=admin_client)
```
sbahar619, line 443:
```suggestion
def service_mesh_tests_namespace():
    # The namespace used for the ServiceMesh tests must be added the `istio-injection` label.
    yield from create_ns(
        name="test-service-mesh-ns",
        labels={"istio-injection": "enabled"},
    )
```
Will that work?
Its more straight forward.

tests/network/service_mesh/conftest.py
```diff
@@ -441,3 +435,10 @@ def vmi_http_server(vm_fedora_with_service_mesh_annotation):
         vm=vm_fedora_with_service_mesh_annotation,
         commands=[f'while true ; do  echo -e "HTTP/1.1 200 OK\n\n $(date)" | nc -l -p {SERVICE_MESH_PORT}  ; done &'],
     )
+
+
+@pytest.fixture(scope="module")
+def service_mesh_tests_namespace(namespace, admin_client):
```
sbahar619, line 441: I think its best to move this fixture to a more top level place in the file based in its fixture scope.

tests/network/service_mesh/conftest.py
```diff
@@ -412,16 +404,18 @@ def change_routing_to_v2(
...
+    ) as pa:
         yield pa
 
 
 @pytest.fixture(scope="class")
 def peer_authentication_service_mesh_deployment(
     istio_system_namespace,
-    namespace,
-    service_mesh_member_roll,
+    service_mesh_tests_namespace,
```
sbahar619, line 418: Redundant, the service_mesh_tests_namespace fixture is call by peer_authentication_strict_service_mesh fixture, so you can just move it up and adjust the order execution.

tests/network/service_mesh/conftest.py
yossisegev: It should work, but I prefer not to do it this way.
Why? Because I prefer being consistent with the common convention in cnv-tests, where each test module has its dedicated namespace which is created by the `namespace` fixture, and if necessary - apply any needed configuration (in this case - this additional label) on that namespace.
Does that make sense?

tests/network/service_mesh/conftest.py
yossisegev: Done

yossisegev: Done

sbahar619 (approved):
lgtm.

dshchedr (approved):

EdDev (approved):
Thank you!

I prefer PRs with proper commits, each treating a specific logic change that can be explored in the future (and less work progress). I know the style used in the project by other SIGs is to keep the work history on the commits, but I find that harder to review. It is not a problem with small changes like this one, but it is very hard with larger ones (for me).

Lets try to do that in the future please.

vsibirsk (approved):

redhat-cnv-qe-bot1:

PR #724, opened 2025-04-10
Mark network tests that don't require node's additional interfaces

yossisegev:
Acceptance tests for new architectures, platforms etc. shouldn't depend on having extra NICs on the cluster nodes, except for the primary interface. In order to have satisfactory coveage in these tests, we mark all the these tests with the a new marker, which can and should be used for these types of tests.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -36,6 +36,7 @@ markers =
     hugepages: tests that require nodes with hugepages
     service_mesh: tests that require the service mesh operator to be installed
     jumbo_frame: tests that require network configurations supporting jumbo frames
+    single_nic: tests that dont require mutli-nic nodes, required for conformance tests for new archs, platforms etc.
     # CI
     smoke: Mark tests as smoke tests
     ci: Mark tests as CI tests
--- a/tests/network/conftest.py
+++ b/tests/network/conftest.py
@@ -218,7 +218,11 @@ def network_sanity(
     failure_msgs = []
     collected_tests = request.session.items
 
-    def _verify_multi_nic():
+    def _verify_multi_nic(request=request):
+        marker_args = request.config.getoption("-m")
+        if marker_args and "single_nic" in marker_args and "not single_nic" not in marker_args:
+            LOGGER.info("Running only single-NIC network cases, no need to verify multi NIC support")
+            return
         LOGGER.info("Verifying if the cluster has multiple NICs for network tests")
         if len(hosts_common_available_ports) <= 1:
             failure_msgs.append(
@@ -290,7 +294,7 @@ def network_sanity(
             else:
                 LOGGER.info("Validated network lane is running against an IPV4 supported cluster")
 
-    _verify_multi_nic()
+    _verify_multi_nic(request=request)
     _verify_dpdk()
     _verify_service_mesh()
     _verify_jumbo_frame()
--- a/tests/network/cluster_addons_operator/test_network_addons_operator.py
+++ b/tests/network/cluster_addons_operator/test_network_addons_operator.py
@@ -215,6 +215,7 @@ def net_add_op_bridge_attached_vm(namespace, net_add_op_br1test_nad):
 
 @pytest.mark.gating
 @pytest.mark.post_upgrade
+@pytest.mark.single_nic
 @pytest.mark.polarion("CNV-2520")
 def test_component_installed_by_operator(network_addons_config_scope_session):
     """
@@ -246,6 +247,7 @@ def test_linux_bridge_functionality(net_add_op_bridge_attached_vm):
 
 
 @pytest.mark.polarion("CNV-6754")
+@pytest.mark.single_nic
 def test_cnao_labels(
     admin_client,
     network_addons_config_scope_session,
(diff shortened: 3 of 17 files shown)


EdDev (commented):
@yossisegev , I think there is a need to update the condition in the [network sanity check](https://github.com/EdDev/openshift-virtualization-tests/blob/ad4f79b33e9800e441fc8c9f77727550d71c3383/tests/network/conftest.py#L223) to allow such tests to run.

yossisegev:
> @yossisegev , I think there is a need to update the condition in the [network sanity check](https://github.com/EdDev/openshift-virtualization-tests/blob/ad4f79b33e9800e441fc8c9f77727550d71c3383/tests/network/conftest.py#L223) to allow such tests to run.

@EdDev I prefer handling it in another PR because:
1.  The tests that are marked in this PR don't rely on the multi NIC verification.
2. Changing this condition affects the bond scenarios as well. While 1 extra NIC is still enough for most of the L2 cases, bond cases still require 2 extra NICs. So my concern is that changing this condition might break some of the bond cases, requiring to specially handling them, which is out of the scope of this PR.

EdDev:
> > @yossisegev , I think there is a need to update the condition in the [network sanity check](https://github.com/EdDev/openshift-virtualization-tests/blob/ad4f79b33e9800e441fc8c9f77727550d71c3383/tests/network/conftest.py#L223) to allow such tests to run.
> 
> @EdDev I prefer handling it in another PR because:
> 
>     1. The tests that are marked in this PR don't rely on the multi NIC verification.
> 
>     2. Changing this condition affects the bond scenarios as well. While 1 extra NIC is still enough for most of the L2 cases, bond cases still require 2 extra NICs. So my concern is that changing this condition might break some of the bond cases, requiring to specially handling them, which is out of the scope of this PR.

The sanity check is triggered if any test in the network package is collected.
Therefore, if you try to run any network test, including the ones who are ok with a single nic, on a cluster that has nodes with a single nic, the execution will fail before any test starts.

At least, this is my understanding of that sanity check does.
@sbahar619 , can you confirm?

dshchedr (approved):

EdDev:
> OK, now I think I got your intention.
> So I need to add a skip of the `_verify_multi_nic` check if the `single_nic` marker is sent in the pytest args, right?

Not a skip, an "exception". If that arg is there, that sanity check should not be executed.

tests/network/conftest.py
```diff
@@ -218,7 +218,11 @@ def network_sanity(
     failure_msgs = []
     collected_tests = request.session.items
 
-    def _verify_multi_nic():
+    def _verify_multi_nic(request=request):
+        marker_args = request.config.getoption("-m")
+        if marker_args and "single_nic" in marker_args and "not single_nic" not in marker_args:
```
EdDev, line 223: I do not like this condition, it blows my mind, however, I do not have a nicer solution for now.
I hope we can find something nicer in a follow up.

EdDev (commented):
Thanks!

/approve

EdDev:
> Thanks!
> 
> /approve

/approved

EdDev (approved):

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-724 published

Anatw (approved):
Thank you

tests/network/conftest.py
yossisegev: Adding the answer I gave in our discussion in slack:
If the `-m` was not sent, then `getoption` doesn't return an empty string, but rather `False`.
See the example below:
```
ipdb> request.config.getoption("-m")
'single_nic'
ipdb> request.config.getoption("-l")
False
```
So if I start by checking "single_nic" in `marker_args` when `marker_args` is False - the whole run will fail.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-724 published

EdDev (approved):
/approve
/lgtm

dshchedr (approved):

pytest.ini
```diff
@@ -36,6 +36,7 @@ markers =
     hugepages: tests that require nodes with hugepages
     service_mesh: tests that require the service mesh operator to be installed
     jumbo_frame: tests that require network configurations supporting jumbo frames
+    single_nic: tests that dont require mutli-nic nodes, required for conformance tests for new archs, platforms etc.
```
sbahar619, line 39:
```suggestion
    single_nic: tests that don’t require multi-NIC nodes, required for conformance tests on new architectures, platforms, etc.

sbahar619, line 39: The single_nic marker name is a bit misleading. All components (tests, nodes, pods) inherently require at least one NIC, so labeling common test scenarios as single_nic adds unnecessary noise. It would be clearer to flip the logic and instead mark the non-standard cases — those requiring multiple NICs — with a multi_nic marker. That way, we only tag the exception cases, which improves readability and avoids confusion.

EdDev: Please send a PR over this to mark all other tests as multi-nic automatically. I.e.anything that is not marked single.

Your suggestion to explicitly mark all as multi-link is impractical IMO, only a few are single.

PR #782, opened 2025-04-21
tests, net, bond: Drop the usage of running_vm

EdDev:
##### What this PR does / why we need it:

`running_vm` is performing multiple checks to assure correct VM start. However, these checks are not required for the network tests.

For the network tests, a successful VM start is one that:
- Reaches ready status (i.e. a VMI in running phase).
- Reaches `AgentConnected` condition status (i.e. cloud-init stage succeeded based on the Fedora image setup).

This is an optimization to reduce observed flakiness on VM startup and simplify the overall maintenance (e.g. the use of a common helper that covers many aspects of a VM startup cycle forces all its users to the same logic, even though that logic is not of interest).

A few redundant fixtures have been removed.

##### Special notes for reviewer:
Verified by running locally with:
`uv run pytest -x --jira --junitxml=xunit_results.xml --pytest-log-file=pytest-tests.log --leftovers-collector --tb=native -rs -s -o log_clie=true --skip-deprecated-api-test --cluster-sanity-skip-storage-check tests/network/bond`

PR #793, opened 2025-04-22
Introduce `wait_for_agent_connected` method

EdDev:

Replace the `wait_for_guest_agent` function in `utilities/virt` module
with a direct call to the wrapper implementation and a method in the
`VirtualMachineForTests` class.

Add the same method to the newer `VirtualMachineBase` under `libs/vm`.

Switch all usages of the underlying general `wait_for_condition` with
the `AGENT_CONNECTED` type, to the new method.

### DIFF
--- a/libs/vm/vm.py
+++ b/libs/vm/vm.py
@@ -6,7 +6,7 @@ from typing import Any
 
 from kubernetes.dynamic import DynamicClient
 from ocp_resources.node import Node
-from ocp_resources.virtual_machine import VirtualMachine
+from ocp_resources.virtual_machine import VirtualMachine, VirtualMachineInstance
 from pytest_testconfig import config as py_config
 
 from libs.vm.spec import (
@@ -71,6 +71,12 @@ class BaseVirtualMachine(VirtualMachine):
             timeout=timeout,
             verify_commands_output=True,
             command_output=True,
+        )
+
+    def wait_for_agent_connected(self) -> None:
+        self.vmi.wait_for_condition(
+            condition=VirtualMachineInstance.Condition.Type.AGENT_CONNECTED,
+            status=VirtualMachineInstance.Condition.Status.TRUE,
         )
 
 
--- a/utilities/virt.py
+++ b/utilities/virt.py
@@ -98,25 +98,6 @@ VM_ERROR_STATUSES = [
 ]
 
 
-def wait_for_guest_agent(vmi: VirtualMachineInstance, timeout: int = TIMEOUT_12MIN) -> None:
-    LOGGER.info(f"Wait until guest agent is active on {vmi.name}")
-
-    sampler = TimeoutSampler(wait_timeout=timeout, sleep=1, func=lambda: vmi.instance)
-    try:
-        for sample in sampler:
-            agent_status = [
-                condition
-                for condition in sample.get("status", {}).get("conditions", {})
-                if condition.get("type") == "AgentConnected" and condition.get("status") == "True"
-            ]
-            if agent_status:
-                return
-
-    except TimeoutExpiredError:
-        LOGGER.error(f"Guest agent is not installed or not active on {vmi.name}")
-        raise
-
-
 def wait_for_vm_interfaces(vmi: VirtualMachineInstance, timeout: int = TIMEOUT_12MIN) -> bool:
     """
     Wait until guest agent report VMI network interfaces.
@@ -132,7 +113,12 @@ def wait_for_vm_interfaces(vmi: VirtualMachineInstance, timeout: int = TIMEOUT_1
         TimeoutExpiredError: After timeout reached.
     """
     # Waiting for guest agent connection before checking guest agent interfaces report
-    wait_for_guest_agent(vmi=vmi, timeout=timeout)
+    LOGGER.info(f"Wait until guest agent is active on {vmi.name}")
+    vmi.wait_for_condition(
+        condition=VirtualMachineInstance.Condition.Type.AGENT_CONNECTED,
+        status=VirtualMachineInstance.Condition.Status.TRUE,
+        timeout=timeout,
+    )
     LOGGER.info(f"Wait for {vmi.name} network interfaces")
     sampler = TimeoutSampler(wait_timeout=timeout, sleep=1, func=lambda: vmi.instance)
     for sample in sampler:
@@ -1062,6 +1048,13 @@ class VirtualMachineForTests(VirtualMachine):
     @property
     def privileged_vmi(self):
         return VirtualMachineInstance(client=get_client(), name=self.name, namespace=self.namespace)
+
+    def wait_for_agent_connected(self, timeout: int = TIMEOUT_5MIN):
+        self.vmi.wait_for_condition(
+            condition=VirtualMachineInstance.Condition.Type.AGENT_CONNECTED,
+            status=VirtualMachineInstance.Condition.Status.TRUE,
+            timeout=timeout,
+        )
 
 
 class VirtualMachineForTestsFromTemplate(VirtualMachineForTests):
--- a/tests/network/bond/test_l2_bridge_over_bond.py
+++ b/tests/network/bond/test_l2_bridge_over_bond.py
@@ -5,7 +5,6 @@ Connectivity over bond bridge on secondary interface
 from collections import OrderedDict
 
 import pytest
-from ocp_resources.resource import Resource
 
 import utilities.network
 from tests.network.libs import cloudinit as netcloud
@@ -177,9 +176,7 @@ def ovs_linux_bond_bridge_attached_vms(ovs_linux_bond_bridge_attached_vma, ovs_l
     vms = (ovs_linux_bond_bridge_attached_vma, ovs_linux_bond_bridge_attached_vmb)
     for vm in vms:
         vm.wait_for_ready_status(status=True)
-        vm.vmi.wait_for_condition(
-            condition=Resource.Condition.Type.AGENT_CONNECTED, status=Resource.Condition.Status.TRUE
-        )
+        vm.wait_for_agent_connected()
     yield vms
 
 
--- a/tests/network/connectivity/test_pod_network.py
+++ b/tests/network/connectivity/test_pod_network.py
@@ -61,17 +61,13 @@ def pod_net_vmb(
 
 @pytest.fixture()
 def pod_net_running_vma(pod_net_vma):
-    pod_net_vma.vmi.wait_for_condition(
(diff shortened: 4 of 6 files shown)


EdDev:
bond tests verified locally.
result: `14 passed, 1 deselected, 79 warnings`

EdDev:
> bond tests verified locally.
> result: `14 passed, 1 deselected, 79 warnings`

UDN tests verified locally.
results: `5 passed, 1 deselected, 26 warnings`

I think these represent well the changes done in the PR.

/verify

EdDev:
@vsibirsk , this is the follow up change per the feedback you [raised](https://github.com/RedHatQE/openshift-virtualization-tests/pull/782#discussion_r2053909193).

EdDev:
> > bond tests verified locally.
> > result: `14 passed, 1 deselected, 79 warnings`
> 
> UDN tests verified locally. results: `5 passed, 1 deselected, 26 warnings`
> 
> I think these represent well the changes done in the PR.
> 
> /verified

/verified

yossisegev (approved):

vsibirsk (commented):
thanks for the pr!

why not do this change directly in wrapper? instead of adding same method to all child classes?

also since you're touching `running_vm()` part could you please also run smoke tests as part of verification (to run smoke tests add these params when running pytest `-m smoke --latest-rhel`)

EdDev:
> thanks for the pr!
> 
> why not do this change directly in wrapper? instead of adding same method to all child classes?

It was already here, I'm not sure why would the wrapper project accept it there.
The wrapper has already the API to check any condition type and state, which sounds the correct generic API for it to have. This one is specific to these tests and its Fedora images which allows to use it frequently to check the guest has successfully booted.

> also since you're touching `running_vm()` part could you please also run smoke tests as part of verification (to run smoke tests add these params when running pytest `-m smoke --latest-rhel`)

Sure, will run and add the results.

EdDev:
> > also since you're touching `running_vm()` part could you please also run smoke tests as part of verification (to run smoke tests add these params when running pytest `-m smoke --latest-rhel`)
> 
> Sure, will run and add the results.

Running: `uv run pytest -m smoke --latest-rhel --jira --junitxml=xunit_results.xml --pytest-log-file=pytest-tests.log --leftovers-collector --tb=native -rs -s -o log_clie=true --skip-deprecated-api-test`

Gave these results: `errors="5" failures="2" skipped="26" tests="54"` (21 passed)

The two failures were on: `"timeout_sampler.TimeoutExpiredError: Timed Out: 1800 Function: ocp_resources.datavolume.wait_for_dv_success.lambda: self.exists Last exception: N/A."`

The 5 errors were on: `"failed on setup with "AssertionError: Unable to connect to test image server:"` ... `error: { "errors" : [ { "status" : 401, "message" : "Token failed verification: signature" } ] }`.

I do not think these errors/failures are related to this change, but please suggest if something further is needed to verify this.

utilities/virt.py
```diff
@@ -1063,6 +1049,13 @@ def wait_for_specific_status(self, status, timeout=TIMEOUT_3MIN, sleep=TIMEOUT_5
     def privileged_vmi(self):
         return VirtualMachineInstance(client=get_client(), name=self.name, namespace=self.namespace)
 
+    def wait_for_agent_connected(self, timeout: int = TIMEOUT_5MIN):
```
Anatw, line 1052: In wait_for_guest_agent the timeout was TIMEOUT_12MIN. Are you sure 5 minutes is enough?

EdDev: That default was never used because the only usage was from `wait_for_vm_interface` which explicitly called it with the same insane timeout of 12min.

The 5min is the default seen on the wrapper.

yossisegev: IIRC, the 12 minutes was originally introduced as a W/A for some infrastructure slowness, but I agree it's not an acceptable timeout to wait for a VM to come up.
It will be interesting, however, to see is this new timeout introduces new failures.

EdDev: It will show up on new tests, not old/existing ones. So there is no regression in this regard.

EdDev:
change: Rebase

utilities/virt.py
EdDev: The function was internal and never used from outside, so I do not understand this long thread about this topic.

Do you have a specific argument why we need a 12min default timeout for waiting on a VM guest to boot?

dshchedr: since the original timeout is not changed inside `wait_for_vm_interfaces` function - I'm ok to have 5 min here. 

12 min timeout was implemented so long ago.. Mostly for slow infra and Windows on PSI clusters, and I'm pretty sure the real reason for it may not even exist anymore, but since `wait_for_vm_interfaces` is used in many places - it would be better to keep original timeout there.

for `wait_for_agent_connected` - 5min sounds ok.

libs/vm/vm.py
```diff
@@ -6,7 +6,7 @@
```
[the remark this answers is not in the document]
dshchedr: +1

EdDev:
@vsibirsk , @dshchedr is the execution of `openshift-virtualization-tests-runner/2177` enough to verify this change?

dshchedr:
> @vsibirsk , @dshchedr is the execution of `openshift-virtualization-tests-runner/2177` enough to verify this change?

As Vasiliy mentioned above - running only network tests sounds not enough. Since the common part is modified - I would prefer to have at least "smoke" tests executed


and also, the comment about moving `libs/vm/vm.py` should also be addressed

vsibirsk:
/hold

@EdDev please address comments and run verification we're asking of

EdDev:
> /hold
> 
> @EdDev please address comments and run verification we're asking of

@vsibirsk , I have addressed the request to run sanity tests [here](https://github.com/RedHatQE/openshift-virtualization-tests/pull/793#issuecomment-2826275194) and asked for help regarding the failures related to storage.

What other comments do I need to address exactly?
If you refer to [this one](https://github.com/RedHatQE/openshift-virtualization-tests/pull/793#discussion_r2058476473), I think it is out of scope of this PR.

I guess I can split this PR into two, one for the libs/vm and the other for the *ForTest object. Do you think it will help?

EdDev:
> > /hold
> > @EdDev please address comments and run verification we're asking of
> 
> @vsibirsk , I have addressed the request to run sanity tests [here](https://github.com/RedHatQE/openshift-virtualization-tests/pull/793#issuecomment-2826275194) and asked for help regarding the failures related to storage.
> 
> What other comments do I need to address exactly? If you refer to [this one](https://github.com/RedHatQE/openshift-virtualization-tests/pull/793#discussion_r2058476473), I think it is out of scope of this PR.
> 
> I guess I can split this PR into two, one for the libs/vm and the other for the *ForTest object. Do you think it will help?

@vsibirsk , @dshchedr , here is a smoke run that passed all except 2 tests: openshift-virtualization-tests-runner/2180/
The failures are related to some DV not being ready,I do not think these are related to the change here.

@jpeimer can you help and have a look and confirm this is not related? Is it something known?

jpeimer:
> @vsibirsk , @dshchedr , here is a smoke run that passed all except 2 tests: openshift-virtualization-tests-runner/2180/ The failures are related to some DV not being ready,I do not think these are related to the change here.
> 
> @jpeimer can you help and have a look and confirm this is not related? Is it something known?

Yes, they are not related
(A small tip: we can run smoke with `--storage-class-matrix=ocs-storagecluster-ceph-rbd-virtualization` so that other storage classes will not be picked up, we don't need them for smoke)

utilities/virt.py
```diff
@@ -132,7 +113,12 @@ def wait_for_vm_interfaces(vmi: VirtualMachineInstance, timeout: int = TIMEOUT_1
         TimeoutExpiredError: After timeout reached.
     """
     # Waiting for guest agent connection before checking guest agent interfaces report
-    wait_for_guest_agent(vmi=vmi, timeout=timeout)
+    LOGGER.info(f"Wait until guest agent is active on {vmi.name}")
+    vmi.wait_for_condition(
```
yossisegev, line 117: If you already implemented `wait_for_agent_connected` below - why not calling it instead of explicitly checking the condition?

yossisegev (changes requested):

utilities/virt.py
EdDev: That one is a method (of the VM class) while this one is a function that works with a VMI.

yossisegev (approved):

vsibirsk:
> @vsibirsk , @dshchedr , here is a smoke run that passed all except 2 tests: openshift-virtualization-tests-runner/2180/

thanks. smoke look ok for me (dv failure doesn't seem related to this pr)

> @vsibirsk , I have addressed the request to run sanity tests https://github.com/RedHatQE/openshift-virtualization-tests/pull/793#issuecomment-2826275194 and asked for help regarding the failures related to storage.

sorry, missed that one and saw the comment with another job run with only network tests there

> What other comments do I need to address exactly?
> If you refer to https://github.com/RedHatQE/openshift-virtualization-tests/pull/793#discussion_r2058476473, I think it is out of scope of this PR.
> 
> I guess I can split this PR into two, one for the libs/vm and the other for the *ForTest object. Do you think it will help?

yes, also agree it should go under `tests/network`. I'm ok for it to be done in follow up PR

libs/vm/vm.py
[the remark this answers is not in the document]
dshchedr: any changes outside of tests/network folder will require approve from room maintainers, so moving all that stuff inside network area will help you to merge PRs faster and independently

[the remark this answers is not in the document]
EdDev: At the moment this is not the case. I.e. the sig-network maintainer cannot approve without a root approver and cannot merge.

But as I mentioned previously, I am fine with a decision to move it. Just prefer not to mix it with this change.

EdDev:
> > @vsibirsk , @dshchedr , here is a smoke run that passed all except 2 tests: openshift-virtualization-tests-runner/2180/ The failures are related to some DV not being ready,I do not think these are related to the change here.
> > @jpeimer can you help and have a look and confirm this is not related? Is it something known?
> 
> Yes, they are not related (A small tip: we can run smoke with `--storage-class-matrix=ocs-storagecluster-ceph-rbd-virtualization` so that other storage classes will not be picked up, we don't need them for smoke)

/verified

cnv-qe-bot1:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-793.

cnv-qe-bot1:

PR #815, opened 2025-04-25
update uv.lock packages

akri3i:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,7 @@ dependencies = [
   "urllib3>=2.2.3",
   "xmltodict>=0.14.2",
   "openshift-python-wrapper~=11.0.10",
+  "marshmallow~=3.26.1",
 ]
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -75,6 +75,7 @@ dependencies = [
   "urllib3>=2.2.3",
   "xmltodict>=0.14.2",
   "openshift-python-wrapper~=11.0.10",
+  "marshmallow~=3.26.1",
```
dshchedr, line 78: why we need it?

dshchedr (changes requested):

pyproject.toml
vsibirsk: it's one of the dependencies of `pytest-jira` and if updated to latest (4.0.0) it breaks pytest
```
TypeError: Field.__init__() got an unexpected keyword argument 'default'
```

PR #820, opened 2025-04-27
tests, net, kmp: Drop the usage of 'running_vm'

EdDev:

Running a VM in the KMP dry-run test is redundant as the test is just checking that the MAC address is assigned by KMP. Running the VM is not really required as it is not checking any KMP functionality.

## Summary by CodeRabbit

- **Tests**
	- Simplified the test for verifying MAC address behavior by removing an unused fixture and retrieving necessary data directly within the test. No changes to test outcomes or user-facing functionality.

### DIFF
--- a/tests/network/dry_run/test_dry_run_kubemacpool.py
+++ b/tests/network/dry_run/test_dry_run_kubemacpool.py
@@ -5,7 +5,7 @@ import pytest
 from kubernetes.client import ApiException
 
 from utilities.network import LINUX_BRIDGE, network_device, network_nad
-from utilities.virt import VirtualMachineForTests, fedora_vm_body, running_vm
+from utilities.virt import VirtualMachineForTests, fedora_vm_body
 
 LOGGER = logging.getLogger(__name__)
 
@@ -95,19 +95,6 @@ def vm_with_mac_address(
 
 
 @pytest.fixture()
-def running_vm_mac_addresses(vm_with_mac_address):
-    running_vm(
-        vm=vm_with_mac_address,
-        check_ssh_connectivity=False,
-        wait_for_interfaces=False,
-    )
-    vm_interfaces = vm_with_mac_address.get_interfaces()
-    for interface in vm_interfaces:
-        LOGGER.info(f"interface_name {interface['name']} was allocated with MAC address {interface[MAC_ADDRESS]}")
-    return vm_interfaces[1][MAC_ADDRESS]
-
-
-@pytest.fixture()
 def dry_run_vm_with_mac_address(
     unprivileged_client,
     namespace,
@@ -130,11 +117,12 @@ def dry_run_vm_with_mac_address(
 def test_dry_run_mac_not_saved(
     dry_run_vma_mac_address,
     vm_with_mac_address,
-    running_vm_mac_addresses,
 ):
     """Assure that the MAC address given to a VM created in dry-run mode is still available in the MAC-pool."""
-    assert dry_run_vma_mac_address == running_vm_mac_addresses, (
-        f"{vm_with_mac_address.name} MAC address is {running_vm_mac_addresses}"
+    vm_interfaces = vm_with_mac_address.get_interfaces()
+    secondary_net_mac = vm_interfaces[1][MAC_ADDRESS]
+    assert dry_run_vma_mac_address == secondary_net_mac, (
+        f"{vm_with_mac_address.name} MAC address is {secondary_net_mac}"
         f" and not {dry_run_vma_mac_address}, as it should be."
     )
 


tests/network/dry_run/test_dry_run_kubemacpool.py
```diff
@@ -130,11 +117,12 @@ def dry_run_vm_with_mac_address(
 def test_dry_run_mac_not_saved(
     dry_run_vma_mac_address,
     vm_with_mac_address,
-    running_vm_mac_addresses,
 ):
     """Assure that the MAC address given to a VM created in dry-run mode is still available in the MAC-pool."""
-    assert dry_run_vma_mac_address == running_vm_mac_addresses, (
-        f"{vm_with_mac_address.name} MAC address is {running_vm_mac_addresses}"
+    vm_interfaces = vm_with_mac_address.get_interfaces()
+    secondary_net_mac = vm_interfaces[1][MAC_ADDRESS]
```
yossisegev, line 123: Are you sure kubevirt always places the default pod interface as the first one in the list, thus any secondary interface will necessarily start from index 1?
I really don't like the approach of relying on presumed indices. Maybe it would be better to iterate through the interfaces and get the one that its corresponding entry in `networks` is `multus`:
```
            networks:
            - name: default
              pod: {}
            - multus:
                networkName: lnx-br-nad
              name: br1
```
WDYT?

yossisegev (changes requested):

tests/network/dry_run/test_dry_run_kubemacpool.py
yossisegev: > Generally speaking, if the default is not specified, nothing will add it implicitly to the VM spec template.

Which is one more reason to explicitly search for the `multus` interface. If the default is indeed not added, the the mutlus interface will occupy `vm_interfaces[0]`, and trying to access `vm_interfaces[1]` will end with a failure.

EdDev: You are right, but why do you request this on this change? This should be done all over.
And I prefer this is done by replacing all VMs to use the ones currently in libs/vm.

tests/network/dry_run/test_dry_run_kubemacpool.py
```diff
@@ -130,11 +117,12 @@ def dry_run_vm_with_mac_address(
 def test_dry_run_mac_not_saved(
     dry_run_vma_mac_address,
     vm_with_mac_address,
-    running_vm_mac_addresses,
 ):
     """Assure that the MAC address given to a VM created in dry-run mode is still available in the MAC-pool."""
-    assert dry_run_vma_mac_address == running_vm_mac_addresses, (
-        f"{vm_with_mac_address.name} MAC address is {running_vm_mac_addresses}"
+    vm_interfaces = vm_with_mac_address.get_interfaces()
```
Anatw, line 122: I don't think lines 122+123 are part of the test. They are what the test require as part of it - part of the test's setup - and therefor should be in a fixture.

Anatw (commented):
Thank you, I added one comment regarding fixture|test distinction.

tests/network/dry_run/test_dry_run_kubemacpool.py
yossisegev: Yes, you are actually right.
Now I see that referencing the specific index (1) in the interfaces list, instead of looking up the interface we are interested in (either by name or by type) is a legacy from before this change.
Very well, I will approve, and we will have this handling in all the places where an interface lookup is required, as we spoke.

yossisegev (approved):

tests/network/dry_run/test_dry_run_kubemacpool.py
EdDev: Reading the interfaces is not part of the test?
We are back to claiming that the test body is just the assertion, which I strongly disagree.

In this case, even the creation of the VM as dry-run should be part if the test, but it is not in the scope of this change.

EdDev:
Verified on `openshift-virtualization-tests-runner/2206/`.

/verified

EdDev:
change: Rebase

EdDev:
Nothing changed.

/verified

yossisegev (approved):

tests/network/dry_run/test_dry_run_kubemacpool.py
Anatw: It was in a fixture before your change, so it make sense it'll stay, there. But as we discussed before - both approaches have their pros and cons, you can continue with this direction, I won't force mine. You can mark this thread as resolved.

Anatw, line 122: I don't see that we start the VM, was this step removed on purpose?

EdDev: OK, thanks.

EdDev: I expressed the change in the commit message:
> Running a VM in the KMP dry-run test is redundant as the test is just
checking that the MAC address is assigned by KMP. Running the VM is not
really required as it is not checking any KMP functionality.

tests/network/dry_run/test_dry_run_kubemacpool.py
rnetser, line 123: Is there no need to check if `vm_interfaces` is not None / empty?

tests/network/dry_run/test_dry_run_kubemacpool.py
EdDev: @Anatw ?

tests/network/dry_run/test_dry_run_kubemacpool.py
EdDev: It must be there because that was the spec with which the VM was created.
If it is not there, we wrote the test or/and there is a bug. Either way it will blow up in the same way access to the API will blow up if something is wrong.

rnetser (approved):
/approve

tests/network/dry_run/test_dry_run_kubemacpool.py
Anatw: If you don't start it, how can you know that the address is available? If the VM isn't started, we can't be sure the assigned MAC address is actually usable. While KMP allocates a MAC when the VMI is defined, I don’t believe it performs full conflict checks or enforcement until the VM is launched. Starting the VM ensures that the MAC address survives the full validation path.

EdDev: This is not how KMP works and it has nothing to do with a running VM.
KMP works on the VM and not on the VMI, therefore there is no meaning to start a VM.
Therefore, the fact a VM was deployed is enough to check KMP added the mac address, and this is exactly what is done here.

dshchedr (approved):

sbahar619 (approved):
lgtm.

cnv-qe-bot1:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-820.

PR #834, opened 2025-04-27
  Fix wait_for_running_pods flakyness

hmeir:
##### Short description:
pod respinned and its name changes, so the pod's list contains an already deleted pod name, which may lead to a failure.
The reason is that pod.instance being called for deleted pod, and the sampler getting stuck.

The solution - wrap it with the try-except block, and get the pod list with each func call

Gating tests flaky across multiple versions.

## Summary by CodeRabbit

- **Refactor**
	- Improved error handling and logic flow in pod status checks for more robust monitoring.
	- Enhanced the process for detecting non-running pods by dynamically retrieving the pod list during status checks.
	- Updated logging to provide clearer information about pod statuses and warnings when issues are detected.

### DIFF
--- a/utilities/infra.py
+++ b/utilities/infra.py
@@ -287,23 +287,24 @@ def get_pod_container_error_status(pod):
 def get_not_running_pods(pods, filter_pods_by_name=None):
     pods_not_running = []
     for pod in pods:
-        pod_instance = pod.instance
-        if filter_pods_by_name and filter_pods_by_name in pod.name:
-            LOGGER.warning(f"Ignoring pod: {pod.name} for pod state validations.")
-            continue
-        container_status_error = get_pod_container_error_status(pod=pod)
-        if container_status_error:
-            pods_not_running.append({pod.name: container_status_error})
         try:
-            # Waits for all pods in a given namespace to be in final healthy state(running/completed).
-            # We also need to keep track of pods marked for deletion as not running. This would ensure any
-            # pod that was spinned up in place of pod marked for deletion, reaches healthy state before end
-            # of this check
-            if pod_instance.metadata.get("deletionTimestamp") or pod_instance.status.phase not in (
-                pod.Status.RUNNING,
-                pod.Status.SUCCEEDED,
-            ):
-                pods_not_running.append({pod.name: pod.status})
+            pod_instance = pod.instance
+            if filter_pods_by_name and filter_pods_by_name in pod.name:
+                LOGGER.warning(f"Ignoring pod: {pod.name} for pod state validations.")
+                continue
+            container_status_error = get_pod_container_error_status(pod=pod)
+            if container_status_error:
+                pods_not_running.append({pod.name: container_status_error})
+
+                # Waits for all pods in a given namespace to be in final healthy state(running/completed).
+                # We also need to keep track of pods marked for deletion as not running. This would ensure any
+                # pod that was spinned up in place of pod marked for deletion, reaches healthy state before end
+                # of this check
+                if pod_instance.metadata.get("deletionTimestamp") or pod_instance.status.phase not in (
+                    pod.Status.RUNNING,
+                    pod.Status.SUCCEEDED,
+                ):
+                    pods_not_running.append({pod.name: pod.status})
         except (ResourceNotFoundError, NotFoundError):
             LOGGER.warning(f"Ignoring pod {pod.name} that disappeared during cluster sanity check")
             pods_not_running.append({pod.name: "Deleted"})
@@ -331,10 +332,10 @@ def wait_for_pods_running(
     samples = TimeoutSampler(
         wait_timeout=TIMEOUT_2MIN,
         sleep=TIMEOUT_5SEC,
-        func=get_not_running_pods,
-        pods=list(Pod.get(dyn_client=admin_client, namespace=namespace.name)),
-        filter_pods_by_name=filter_pods_by_name,
-        exceptions_dict={NotFoundError: []},
+        func=lambda: get_not_running_pods(
+            pods=list(Pod.get(dyn_client=admin_client, namespace=namespace.name)),
+            filter_pods_by_name=filter_pods_by_name,
+        ),
     )
     sample = None
     try:
--- a/utilities/infra.py
+++ b/utilities/infra.py
@@ -284,8 +284,10 @@ def get_pod_container_error_status(pod):
             return waiting_container["reason"] if waiting_container.get("reason") else waiting_container
 
 
-def get_not_running_pods(pods, filter_pods_by_name=None):
+def get_not_running_pods(pods=None, dyn_client=None, namespace=None, filter_pods_by_name=None):
+    pods = pods if pods else list(Pod.get(dyn_client=dyn_client, namespace=namespace.name))
     pods_not_running = []
+
     for pod in pods:
         try:
             pod_instance = pod.instance
@@ -332,10 +334,10 @@ def wait_for_pods_running(
     samples = TimeoutSampler(
         wait_timeout=TIMEOUT_2MIN,
         sleep=TIMEOUT_5SEC,
-        func=lambda: get_not_running_pods(
-            pods=list(Pod.get(dyn_client=admin_client, namespace=namespace.name)),
-            filter_pods_by_name=filter_pods_by_name,
-        ),
+        func=get_not_running_pods,
+        dyn_client=admin_client,
+        namespace=namespace,
+        filter_pods_by_name=filter_pods_by_name,
     )
     sample = None
     try:
--- a/utilities/infra.py
+++ b/utilities/infra.py
@@ -274,14 +274,17 @@ def wait_for_pods_deletion(pods):
         pod.wait_deleted()
 
 
-def get_pod_container_error_status(pod):
-    pod_instance_status = pod.instance.status
-
-    # Check the containerStatuses and if any containers is in waiting state, return that information:
-    for container_status in pod_instance_status.get("containerStatuses", []):
-        waiting_container = container_status.get("state", {}).get("waiting")
-        if waiting_container:
-            return waiting_container["reason"] if waiting_container.get("reason") else waiting_container
+def get_pod_container_error_status(pod: Pod) -> str | None:
+    try:
+        pod_instance_status = pod.instance.status
+        # Check the containerStatuses and if any container is in waiting state, return that information:
+        for container_status in pod_instance_status.get("containerStatuses", []):
+            if waiting_container := container_status.get("state", {}).get("waiting"):
+                return waiting_container["reason"] if waiting_container.get("reason") else waiting_container
+        return None
+    except NotFoundError:
+        LOGGER.error(f"Pod {pod.name} was not found")
+        raise
 
 
 def get_not_running_pods(pods: list[Pod], filter_pods_by_name: str = "") -> list[dict[str, str]]:
--- a/utilities/infra.py
+++ b/utilities/infra.py
@@ -259,7 +259,7 @@ def get_jira_status(jira):
     return jira_connection.issue(id=jira).fields.status.name.lower()
 
 
-def get_pods(dyn_client, namespace, label=None):
+def get_pods(dyn_client: DynamicClient, namespace: Namespace, label: str = "") -> list[Pod]:
     return list(
         Pod.get(
             dyn_client=dyn_client,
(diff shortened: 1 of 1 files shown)


redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-834 published

utilities/hco.py
```diff
@@ -236,6 +236,7 @@ def wait_for_hco_post_update_stable_state(admin_client, hco_namespace):
         namespace=hco_namespace.name,
     ):
         wait_for_dp(dp=deployment)
+        deployment.wait_for_replicas()
```
vsibirsk, line 239: isn't it the same action? to me it seems both `wait_for_dp` func and `wait_for_replicas` method duplicate each other

utilities/infra.py
```diff
@@ -333,7 +333,6 @@ def wait_for_pods_running(
         func=get_not_running_pods,
         pods=list(Pod.get(dyn_client=admin_client, namespace=namespace.name)),
         filter_pods_by_name=filter_pods_by_name,
-        exceptions_dict={NotFoundError: []},
```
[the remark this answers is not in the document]
vsibirsk: +1, i don't see how this can help reducing the flake.

vsibirsk (changes requested):

utilities/infra.py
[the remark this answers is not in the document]
hmeir: I think that the issue is pod.instance being called when a pod is deleted, and the pod list being static.

I handled it wrong, it either should move inside the "try" block or call the list of pods each time(for new/deleted ones)
I'll try that and update the PR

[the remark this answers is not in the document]
hmeir: Now we dont really need that, because we have`except (ResourceNotFoundError, NotFoundError):` at line 307

dbasunag (approved):

utilities/infra.py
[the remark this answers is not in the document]
dbasunag: Interesting. In that case, you can try what CodeRabbit suggested:
```
-        func=get_not_running_pods,
-        pods=list(Pod.get(dyn_client=admin_client, namespace=namespace_name)),
+        func=lambda: get_not_running_pods(
+            pods=list(Pod.get(dyn_client=admin_client, namespace=namespace_name))
+        ),
```

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-834 published

dbasunag (approved):

utilities/infra.py
```diff
@@ -330,10 +331,10 @@ def wait_for_pods_running(
     samples = TimeoutSampler(
         wait_timeout=TIMEOUT_2MIN,
         sleep=TIMEOUT_5SEC,
-        func=get_not_running_pods,
-        pods=list(Pod.get(dyn_client=admin_client, namespace=namespace.name)),
-        filter_pods_by_name=filter_pods_by_name,
-        exceptions_dict={NotFoundError: []},
+        func=lambda: get_not_running_pods(
```
myakove, line 334: I don't understand why you need to use `lambda` here?
It will work the same without it (as it was before).

dshchedr (approved):

utilities/infra.py
myakove: It's seems odd to me to use lambda on a function, lambda is for convert something to function.
It's better to send namespace to `get_not_running_pods` and get the pods there.

myakove (changes requested):

utilities/infra.py
dshchedr: not sure I understand why you need to execute Pod.get() each time.. I think it should work when you pass 

`exceptions_dict={NotFoundError: []}` , just add `ResourceNotFoundError` there and put `pod_instance = pod.instance` inside the try

dbasunag: If the pods are not pulled every time you call get_not_running_pods() , then you are not considering new pods that are spinned up or the ones that are gone. The list must not be static. However, my understanding is, with timeout sampler we don't need lambda and I did not see any reason to believe it is not working either so far. The lambda function was suggested by code rabbit and it claimed `wait_for_pods_running samples a static pod list – pods created later are ignored.` https://github.com/opendatahub-io/opendatahub-tests/pull/270#discussion_r2064175865

Harel mentioned he is observing the same. So I suggested he gives this a try. My observation is not this though. I have not see a reason to believe we are looking at static list of pod (in or outside CNV).

hmeir: > not sure I understand why you need to execute Pod.get() each time.. I think it should work when you pass
> 
> exceptions_dict={NotFoundError: []} , just add ResourceNotFoundError there and put pod_instance = pod.instance inside the try

It won't help in a case that you'll actually find an non-running pod.
In that case, you will hit this block:

```
        except (ResourceNotFoundError, NotFoundError):
            LOGGER.warning(f"Ignoring pod {pod.name} that disappeared during cluster sanity check")
            pods_not_running.append({pod.name: "Deleted"})
```

Then the sampler will call the function again with the **same arguments** as the first call. 
So the `pods` arg is already determined, and will always hit the except condition since the non-running pod will always be present, and get to TImeExpiredError eventually.

There are multiple ways to handle that, like lambda on the func/pod arg, update the list of pods with every non successful check, or what @myakove  suggested.

I tried to implement the last one, let me know what you think

utilities/infra.py
```diff
@@ -283,26 +283,29 @@ def get_pod_container_error_status(pod):
             return waiting_container["reason"] if waiting_container.get("reason") else waiting_container
 
 
-def get_not_running_pods(pods, filter_pods_by_name=None):
+def get_not_running_pods(pods=None, dyn_client=None, namespace=None, filter_pods_by_name=None):
+    pods = pods if pods else list(Pod.get(dyn_client=dyn_client, namespace=namespace.name))
```
dbasunag, line 287: I would remove pod as argument. If you need to pull it everytime, then there is no point.

utilities/infra.py
dbasunag: I prefer the lamda function, as it is cleaner. But I will not hold this PR due to that. I however, am not seeing what you described, with this code. In my case, new pods are getting spinned up and I see not running pods too, but the code worked as the list was never stale. I believe we have same code in many places at least in this repo.

utilities/infra.py
hmeir: The issue is `wait_for_pod_running_by_prefix` is using this method.

I tried to perform similar check at my new patch:
calling `get_pods` from the sampler, then checking the non_running pods against the pod list if sample isn't non, 
and applying the consecutive check on there.

Let me know what you think

utilities/infra.py
```diff
@@ -330,21 +331,26 @@ def wait_for_pods_running(
...
         current_check = 0
         for sample in samples:
-            if not sample:
-                current_check += 1
-                if current_check >= number_of_consecutive_checks:
-                    return True
-            else:
-                current_check = 0
+            if sample:
+                not_running_pods = get_not_running_pods(pods=sample, filter_pods_by_name=filter_pods_by_name)
```
dbasunag, line 344: I think you should leave the wait_for_pods_running() with the changes you made in get_not_running_pods() and handle what you need for your other call separately. May be do a sampler call there for get_not_running_pods() or redesign that code? I am not seeing any reason to believe it is working with stale pods so far.

utilities/infra.py
```diff
@@ -286,23 +286,24 @@ def get_pod_container_error_status(pod):
...
-                pod.Status.SUCCEEDED,
-            ):
-                pods_not_running.append({pod.name: pod.status})
+            pod_instance = pod.instance
+            if filter_pods_by_name and filter_pods_by_name in pod.name:
+                LOGGER.warning(f"Ignoring pod: {pod.name} for pod state validations.")
+                continue
+            container_status_error = get_pod_container_error_status(pod=pod)
+            if container_status_error:
+                pods_not_running.append({pod.name: container_status_error})
```
rnetser, line 296: Why was this part moved under the try clause?
the check if a pod exists should be in `get_pod_container_error_status`

utilities/infra.py
```diff
@@ -286,23 +286,24 @@ def get_pod_container_error_status(pod):
...
-                pod.Status.RUNNING,
-                pod.Status.SUCCEEDED,
-            ):
-                pods_not_running.append({pod.name: pod.status})
+            pod_instance = pod.instance
+            if filter_pods_by_name and filter_pods_by_name in pod.name:
+                LOGGER.warning(f"Ignoring pod: {pod.name} for pod state validations.")
+                continue
+            container_status_error = get_pod_container_error_status(pod=pod)
+            if container_status_error:
```
rnetser, line 295: can be re-written using walrus:
```suggestion

            if container_status_error := get_pod_container_error_status(pod=pod):
```

utilities/infra.py
```diff
@@ -330,21 +331,26 @@ def wait_for_pods_running(
     samples = TimeoutSampler(
         wait_timeout=TIMEOUT_2MIN,
         sleep=TIMEOUT_5SEC,
-        func=get_not_running_pods,
-        pods=list(Pod.get(dyn_client=admin_client, namespace=namespace.name)),
-        filter_pods_by_name=filter_pods_by_name,
-        exceptions_dict={NotFoundError: []},
+        func=get_pods,
```
rnetser, line 334: you can use lambda: list.... directly here

utilities/infra.py
rnetser: I see that `get_pod_container_error_status` only checks for `.state.waiting` and not `state.terminated`.
Coulnd't this also lead to flakiness?

utilities/infra.py
```diff
@@ -286,23 +286,24 @@ def get_pod_container_error_status(pod):
 def get_not_running_pods(pods, filter_pods_by_name=None):
```
OhadRevah, line 286: Add typing

utilities/infra.py
hmeir: This is the flakyness reason: https://github.com/RedHatQE/openshift-virtualization-tests/pull/834#discussion_r2079030748

If you'd like, I can share my custom test that demonstrates this specific scenario failing.

utilities/infra.py
hmeir: Its possible,  but isn't the current setup more informative?

utilities/infra.py
rnetser: so every call to this function should send the right pods

hmeir: done

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-834 published

dshchedr (dismissed):

utilities/infra.py
```diff
@@ -275,30 +275,30 @@ def wait_for_pods_deletion(pods):
 
 def get_pod_container_error_status(pod):
     pod_instance_status = pod.instance.status
```
rnetser, line 277: should be wrapped with try-execpt. if the pod does not exit, this will fail.

utilities/infra.py
```diff
@@ -275,30 +275,30 @@ def wait_for_pods_deletion(pods):
...
+            pod_instance = pod.instance
+            if container_status_error := get_pod_container_error_status(pod=pod):
+                pods_not_running.append({pod.name: container_status_error})
+
             # Waits for all pods in a given namespace to be in final healthy state(running/completed).
             # We also need to keep track of pods marked for deletion as not running. This would ensure any
             # pod that was spinned up in place of pod marked for deletion, reaches healthy state before end
             # of this check
-            if pod_instance.metadata.get("deletionTimestamp") or pod_instance.status.phase not in (
+            elif pod_instance.metadata.get("deletionTimestamp") or pod_instance.status.phase not in (
```
rnetser, line 301: i think it would make sense to reverse the conditions - i.e first check pod's metadata and phase and only if does not meet that condition iterate over the containers.

utilities/infra.py
```diff
@@ -310,10 +310,10 @@ def get_not_running_pods(pods, filter_pods_by_name=None):
...
 
 def wait_for_pods_running(
-    admin_client,
-    namespace,
-    number_of_consecutive_checks=1,
-    filter_pods_by_name=None,
+    admin_client: DynamicClient,
+    namespace: Namespace,
+    number_of_consecutive_checks: int = 1,
+    filter_pods_by_name: str = "",
```
rnetser, line 316: missing return typing

utilities/infra.py
```diff
@@ -330,26 +330,31 @@ def wait_for_pods_running(
...
         current_check = 0
         for sample in samples:
-            if not sample:
-                current_check += 1
-                if current_check >= number_of_consecutive_checks:
-                    return True
-            else:
-                current_check = 0
+            if sample:
+                LOGGER.info(f"All pods: {[pod.name for pod in sample]}")
```
rnetser, line 345: does this log add extra valut -i.e pod names only?

rnetser (changes requested):

utilities/infra.py
hmeir: Its good for debugging.
But yeah I guess we dont need that for 99% of the cases.

Deleting

utilities/infra.py
hmeir: If the pod doesn't exit we will hit NotFoundError in `get_not_running_pods`

hmeir: If the pod doesn't exit we will hit NotFoundError in `get_not_running_pods`

utilities/infra.py
```diff
@@ -275,26 +275,26 @@ def wait_for_pods_deletion(pods):
...
-            return waiting_container["reason"] if waiting_container.get("reason") else waiting_container
+    try:
+        pod_instance_status = pod.instance.status
+        # Check the containerStatuses and if any containers is in waiting state, return that information:
+        for container_status in pod_instance_status.get("containerStatuses", []):
+            waiting_container = container_status.get("state", {}).get("waiting")
+            if waiting_container:
+                return waiting_container["reason"] if waiting_container.get("reason") else waiting_container
+    except NotFoundError:
+        LOGGER.error(f"Pod {pod.name} was deleted")
```
rnetser, line 286: nitpick (as a user can send a pod object that never existed:
```suggestion
        LOGGER.error(f"Pod {pod.name} was not found")
```

rnetser (commented):
/approve

vsibirsk (approved):

jpeimer (approved):

cnv-qe-bot1:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-834.

PR #866, opened 2025-04-28
Mark additional network tests for gating

yossisegev:
In addition to marking new gating tests, I also un-marked tests that are not relevant anymore for gating, based on the discussion in https://issues.redhat.com/browse/CNV-52356

## Summary by CodeRabbit

- **Tests**
	- Updated test markers to adjust which tests are included in the gating suite.
	- Removed gating marker from several tests related to network bonding, OVS bridge connectivity, and bridge marker device existence.
	- Added gating marker to the entire service configuration via manifest test class and to the connectivity preservation after live migration test.

### DIFF
--- a/tests/network/migration/test_migration.py
+++ b/tests/network/migration/test_migration.py
@@ -385,7 +385,6 @@ def test_connectivity_after_migration_and_restart(
 
 
 @pytest.mark.polarion("CNV-2061")
-@pytest.mark.gating
 def test_migration_with_masquerade(
     ip_stack_version_matrix__module__,
     admin_client,
--- a/tests/network/migration/test_migration.py
+++ b/tests/network/migration/test_migration.py
@@ -385,6 +385,7 @@ def test_connectivity_after_migration_and_restart(
 
 
 @pytest.mark.polarion("CNV-2061")
+@pytest.mark.gating
 def test_migration_with_masquerade(
     ip_stack_version_matrix__module__,
     admin_client,
--- a/tests/network/bond/test_l2_bridge_over_bond.py
+++ b/tests/network/bond/test_l2_bridge_over_bond.py
@@ -182,7 +182,6 @@ def ovs_linux_bond_bridge_attached_vms(ovs_linux_bond_bridge_attached_vma, ovs_l
 
 class TestBondConnectivity:
     @pytest.mark.ipv4
-    @pytest.mark.gating
     @pytest.mark.polarion("CNV-3366")
     def test_bond(
         self,
--- a/tests/network/connectivity/test_ovs_linux_bridge.py
+++ b/tests/network/connectivity/test_ovs_linux_bridge.py
@@ -106,7 +106,6 @@ class TestConnectivityLinuxBridge:
 
 @pytest.mark.usefixtures("hyperconverged_ovs_annotations_enabled_scope_session")
 class TestConnectivityOVSBridge:
-    @pytest.mark.gating
     @pytest.mark.post_upgrade
     @pytest.mark.parametrize(
         "use_default_bridge",
--- a/tests/network/general/test_bridge_marker.py
+++ b/tests/network/general/test_bridge_marker.py
@@ -148,7 +148,6 @@ def test_bridge_marker_no_device(bridge_marker_bridge_network, bridge_attached_v
 # note: the order of fixtures is important because we should first create the
 # device before attaching a VMI to it
 @pytest.mark.sno
-@pytest.mark.gating
 @pytest.mark.polarion("CNV-2235")
 def test_bridge_marker_device_exists(bridge_device_on_all_nodes, bridge_attached_vmi_for_bridge_marker_device_exists):
     """Check that VMI successfully starts when bridge device is present."""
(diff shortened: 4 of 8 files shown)


tests/network/migration/test_migration.py
```diff
@@ -385,6 +385,7 @@ def test_connectivity_after_migration_and_restart(
 
 
 @pytest.mark.polarion("CNV-2061")
+@pytest.mark.gating
```
EdDev, line 388: This test will fail if executed on PSI because those clusters do not support IPv6.
Are you sure we can mark this as gating at the moment?

EdDev (commented):
Thanks. I had a question inline.

tests/network/migration/test_migration.py
yossisegev: I am sure [you are right](https://jenkins-csb-cnvqe-main.dno.corp.redhat.com/job/openshift-virtualization-tests-runner/2201/testReport/tests.network.migration/test_migration/test_migration_with_masquerade__ipv6__/).
I will remove the marker from this test.

EdDev (approved):
Thanks!

/approve

tests/network/general/test_network_naming.py
```diff
@@ -12,6 +12,7 @@ def invalid_network_names():
 
 
 @pytest.mark.polarion("CNV-8304")
+@pytest.mark.gating
 def test_vm_with_illegal_network_name(namespace, unprivileged_client, invalid_network_names):
```
Anatw, line 16: Why do we want this scenario as a gating?

Anatw (changes requested):
test_connectivity_over_pod_network is missing - we should also remove the gating marker from it.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-866 published

dshchedr (approved):

tests/network/general/test_network_naming.py
EdDev: We probably need to set a proper criteria for gating. I tend to agree that this test is not covering an important represebtative feature. Lets follow up on this.

EdDev:
> test_connectivity_over_pod_network is missing - we should also remove the gating marker from it.

Please provide reasoning.

EdDev (commented):
/approve
/lgtm

Anatw:
> > test_connectivity_over_pod_network is missing - we should also remove the gating marker from it.
> 
> Please provide reasoning.

This scenario is marked as a one that should be removed in the Jira task (https://issues.redhat.com/browse/CNV-52356)
But is missing from the PR.

Anatw (changes requested):
Added a comment

sbahar619:
Thanks @yossisegev 
For all the gating removal, it is justified as it was discussed in the Jira ticket.
Do we have similar justification for the gating you added?
Only UDN currently seems to me as a good candidate without further justification as it was the main feature of 4.18.

tests/network/general/test_network_naming.py
yossisegev: If there are already 2 team members that are reluctant about  having this test as a gating, then I am good with removing it.

yossisegev:
> Thanks @yossisegev For all the gating removal, it is justified as it was discussed in the Jira ticket. Do we have similar justification for the gating you added? Only UDN currently seems to me as a good candidate without further justification as it was the main feature of 4.18.

Good point @sbahar619 , so I'll answer about the additional 2 tests:
`test_service_with_configured_ip_families`: As network QE, I think we must have at least one test that covers exposing VMs via services.
`test_veth_removed_from_host_after_vm_deleted`: If this one fails, then we have a serious problem that affects cluster resources, and might eventually exhaust them, if nodes continue having orphan VETHs. We would like to be aware of it on gating time and handle it, don't you think?

yossisegev:
> > > test_connectivity_over_pod_network is missing - we should also remove the gating marker from it.
> > 
> > 
> > Please provide reasoning.
> 
> This scenario is marked as a one that should be removed in the Jira task (https://issues.redhat.com/browse/CNV-52356) But is missing from the PR.

@Anatw This is not what is said in the issue. It says that this test can settle for covering virtio and leave out e1000e.
However -
1.  Un-matrixing this test and splitting it to 2 separate tests is a change that is too complex for this PR and out of its scope, IMO.
2. There's no harm in covering e1000e type for VM's pod network.
Anyway I think we all agree that this scenario is a must for gating. If pod networking is broken in our VMs - we definitely want to know about it on gating time.

sbahar619:
> > Thanks @yossisegev For all the gating removal, it is justified as it was discussed in the Jira ticket. Do we have similar justification for the gating you added? Only UDN currently seems to me as a good candidate without further justification as it was the main feature of 4.18.
> 
> Good point @sbahar619 , so I'll answer about the additional 2 tests: `test_service_with_configured_ip_families`: As network QE, I think we must have at least one test that covers exposing VMs via services. `test_veth_removed_from_host_after_vm_deleted`: If this one fails, then we have a serious problem that affects cluster resources, and might eventually exhaust them, if nodes continue having orphan VETHs. We would like to be aware of it on gating time and handle it, don't you think?

Thanks @yossisegev , ack.

sbahar619 (approved):
lgtm.

PR #900, opened 2025-05-04
    Localnet scenario - connectivity over migration on additional bridge

Anatw:
This commit updates the localnet tests to use the new
'create_traffic_server' and 'create_traffic_client' functions from the
'liblocalnet' module. The existing Server and Client fixtures have been
replaced with these new functions to improve test readability and
maintainability.
    
Additionally, the directory now contain 2 modules: one for testing the
default (br-ex) bridge and another to test the additional ovs bridge
scenarios. Each module contain a test for verifying connectivity between
localnet VMs before and after migration.
    
The existing localnet scenario
(test_connectivity_over_migration_between_localnet_vms - CNV-11775) was
updated to use the new functions, while the new
test_connectivity_over_migration_between_ovs_bridge_localnet_vms
scenario was added to demonstrate connectivity between localnet VMs
using an additional bridge.

## Summary by CodeRabbit

- **New Features**
  - Introduced support for testing connectivity between VMs on both default and OVS bridge networks.
  - Added automated tests to verify TCP connectivity after VM migration on default and OVS bridge networks.
  - Added fixtures and utilities to support an OVS bridge network on a secondary node NIC with associated VMs and traffic generators.

- **Refactor**
  - Unified and modularized VM startup and IP assignment for improved flexibility.
  - Replaced manual traffic generator setup with reusable helper functions.
  - Parameterized network naming in VM and network attachment definitions for enhanced configurability.

- **Tests**
  - Expanded test coverage for network configurations and VM migration scenarios.


yossisegev:
> This commit updates the localnet tests to use the new start_server_vm and start_client_vm fixtures

These are functions, not fixtures.

yossisegev (approved):
A small nit in the PR message (referring to functions as fixtures), but nothing worth not approving this PR.
Thank you Anat

tests/network/localnet/conftest.py
```diff
@@ -2,22 +2,29 @@
...
-from tests.network.localnet.liblocalnet import NETWORK_NAME, localnet_nad, localnet_vm
+from tests.network.localnet.liblocalnet import NETWORK_NAME, localnet_nad, localnet_vm, start_client_vm, start_server_vm
 from utilities.constants import (
+    OVS_BRIDGE,
     WORKER_NODE_LABEL_KEY,
 )
 from utilities.infra import create_ns
 
-_IPERF_SERVER_PORT = 5201
+DHCP_IP_PREFIX = "10.0.0"
```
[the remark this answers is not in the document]
Anatw: Ack.

Anatw:
Change: rebase on main.

tests/network/localnet/conftest.py
```diff
@@ -2,22 +2,29 @@
 
 import pytest
 from ocp_resources.namespace import Namespace
+from ocp_resources.node import Node
+from ocp_resources.resource import Resource
```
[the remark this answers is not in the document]
Anatw: Ack.

tests/network/localnet/conftest.py
```diff
@@ -101,18 +103,115 @@ def vms_localnet(
...
-        assert client.is_running()
+    with start_client_vm(vms=vms_localnet) as client:
+        yield client
+
+
+@pytest.fixture(scope="class")
+def localnet_additional_bridge(
+    worker_node1: Node, nodes_available_nics: dict[str, list[str]]
+) -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
+    bridge_name = "localnet-additional-bridge"
```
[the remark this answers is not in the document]
sbahar619: ovs naming...

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/301095d50f4d28366d1cfc8cea7da147fc232b45..a980c7254ca1b12f3bcb72df8b6ab4500b5b8e2a):
Change resource import.
Add 'ovs' to additional-bridge naming.
Rename DHCP_IP_PREFIX to IP_NETWORK_PREFIX.
Add a utility function to create the Interfaces section when creating an nncp bridge.
Refactor the commit message.

tests/network/localnet/conftest.py
```diff
@@ -101,18 +103,115 @@ def vms_localnet(
...
+@pytest.fixture(scope="class")
+def nad_localnet_additional_bridge(
+    namespace: Namespace, localnet_additional_bridge: Generator[libnncp.NodeNetworkConfigurationPolicy], vlan_id: int
+) -> Generator[netattachdef.NetworkAttachmentDefinition]:
+    with localnet_nad(namespace=namespace.name, name=f"{NETWORK_NAME}-nad", vlan_id=vlan_id) as nad:
+        yield nad
+
+
+@pytest.fixture(scope="class")
+def additional_bridged_localnet_vma(
```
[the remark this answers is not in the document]
sbahar619: I vote for numeric naming, plus the name should start with hint about what it is, in this case it is a vm, so place it at the start not in the end.

tests/network/localnet/conftest.py
```diff
@@ -101,18 +114,105 @@ def vms_localnet(
...
-        server_ip=lookup_iface_status(vm=vm1, iface_name=NETWORK_NAME)[IP_ADDRESS],
-        server_port=_IPERF_SERVER_PORT,
-    ) as client:
-        assert client.is_running()
+    with start_client_vm(vms=vms_localnet) as client:
+        yield client
+
+
+@pytest.fixture(scope="class")
+def localnet_additional_ovs_bridge(
```
sbahar619, line 128: Please add nncp at the start of the fixture name, so it will be clear what it represent.

tests/network/localnet/test_localnet.py
```diff
@@ -4,8 +4,22 @@
...
+    @pytest.mark.single_nic
+    @pytest.mark.polarion("CNV-11775")
+    def test_connectivity_over_migration_between_localnet_vms(self, localnet_server, localnet_client):
+        migrate_vm_and_verify(vm=localnet_client.vm)
+        assert is_tcp_connection(server=localnet_server, client=localnet_client)
+
+
+class TestLocalnetAdditionalBridge:
+    @pytest.mark.ipv4
+    @pytest.mark.polarion("CNV-11905")
```
sbahar619, line 18: The test case and the test automation implementation are not sync, example:
The test case instruct to define 2 different namespaces and deploy in each the VMs.

tests/network/localnet/liblocalnet.py
```diff
@@ -1,12 +1,58 @@
...
+    with Client(
+        vm=vm_client,
+        server_ip=lookup_iface_status(vm=vm_server, iface_name=NETWORK_NAME)[IP_ADDRESS],
+        server_port=_IPERF_SERVER_PORT,
+    ) as client:
+        assert client.is_running()
+        yield client
+
+
+def additional_ovs_bridge_interfaces(bridge_name: str, worker: str) -> libnncp.Interface:
```
sbahar619, line 40: Please use singular name as this function is returning Interface object.

sbahar619, line 40: Please use more accurate name for the parameter "worker".
This parameter value is used to assign the interface port, so some port name variation is more accurate.

sbahar619 (changes requested):

tests/network/localnet/test_localnet.py
```diff
@@ -4,8 +4,22 @@
...
+    def test_connectivity_between_localnet_vms(
```
sbahar619, line 19: The test name can be the same name as for the above class:
test_connectivity_over_migration_between_localnet_vms
They both perform the same workflows, the class give for each test the correct context.

tests/network/localnet/conftest.py
Anatw: Ack.

tests/network/localnet/liblocalnet.py
Anatw: Ack.

Anatw: Ack, I'll change it to worker_port_name.

tests/network/localnet/test_localnet.py
Anatw: I rechecked the scenario and only a single namespace is mentioned there. But there is a node labeling step that is redundant, I'll remove it.

tests/network/localnet/test_localnet.py
Anatw: Correct, I forgot to change it according to the new name in the first test case. Thanks!

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/a980c7254ca1b12f3bcb72df8b6ab4500b5b8e2a..a4ebc8806a216edbf362d02316e7d3aa36780e13):
Mainly naming changes according to the code review.

tests/network/localnet/conftest.py
```diff
@@ -101,18 +114,105 @@ def vms_localnet(
...
-        assert client.is_running()
+    with start_client_vm(vms=vms_localnet) as client:
+        yield client
+
+
+@pytest.fixture(scope="class")
+def nncp_localnet_additional_ovs_bridge(
+    worker_node1: Node, nodes_available_nics: dict[str, list[str]]
+) -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
+    bridge_name = "localnet-additional-ovs-bridge"
```
yossisegev, line 131: This is not applicable, the interface name is too long and it will fail the NNCP creation.
IIRC the limitation is 15 chars.

yossisegev (changes requested):

tests/network/localnet/conftest.py
```diff
@@ -38,54 +50,55 @@ def nncp_localnet() -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
...
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="class")
 def vm_localnet_1(nad_localnet_1: netattachdef.NetworkAttachmentDefinition) -> Generator[BaseVirtualMachine]:
     with localnet_vm(
-        namespace=nad_localnet_1.namespace, name="test-vm1", network=nad_localnet_1.name, cidr="10.0.0.1/24"
+        namespace=nad_localnet_1.namespace,
+        name="test-vm1",
+        network=nad_localnet_1.name,
+        cidr=f"{IP_NETWORK_PREFIX}.1/24",
```
EdDev, line 85: What meaning does `IP_NETWORK_PREFIX` has exactly?
Even if we change the name, then we are still left to define the host value and the mask in a free stile, so the improvement is not that great.

If you want to make sure we get an available address from the same subnet, perhaps something like this will help:
```python
@pytest.fixture
def ipv4_address_pool():
  net_prefix = "10.0.0"
  return (f"{net_prefix}.{host_value}/24" for host_val in range(1, 254))
```

In the future, we can move this to the common lib and consider `net_prefix` and the subnet as optional arguments. But this is not needed now.

If you have this, for each test we will have a new IP pool to work with, looking like this:
`cidr=next(ipv4_address_pool)`

tests/network/localnet/liblocalnet.py
```diff
@@ -1,12 +1,57 @@
...
-from libs.net.vmspec import add_network_interface, add_volume_disk
+from libs.net.traffic_generator import Client, Server
+from libs.net.vmspec import IP_ADDRESS, add_network_interface, add_volume_disk, lookup_iface_status
 from libs.vm.affinity import new_pod_anti_affinity
 from libs.vm.factory import base_vmspec, fedora_vm
 from libs.vm.spec import CloudInitNoCloud, Interface, Metadata, Multus, Network
 from libs.vm.vm import BaseVirtualMachine, cloudinitdisk_storage
 from tests.network.libs import cloudinit
+from tests.network.libs.nodenetworkconfigurationpolicy import Resource
+from utilities.constants import OVS_BRIDGE
```
EdDev, line 14: Please create something new, this has no meaning when reading it.
The constant name should represent what it is not to reflect the value itself.

I'm not even sure that in all places it means the same.

I won't insist on this for this PR, but it needs to come as a follow up.

tests/network/localnet/liblocalnet.py
```diff
@@ -1,12 +1,57 @@
...
+    with Client(
+        vm=vm_client,
+        server_ip=lookup_iface_status(vm=vm_server, iface_name=NETWORK_NAME)[IP_ADDRESS],
+        server_port=_IPERF_SERVER_PORT,
+    ) as client:
+        assert client.is_running()
+        yield client
+
+
+def additional_ovs_bridge_interface(bridge_name: str, worker_port_name: str) -> libnncp.Interface:
```
EdDev, line 39: Why this needs to appear in the libs? I mean, why it is not inline in the configuration? To me, seeing the whole configuration/spec in one place is more clear on what is going on.

The name also is odd... this returns an ovs-bridge interface. If it is in addition to some other interface or not, is not known to this thing. I mean, you do not know how it will be used.

tests/network/localnet/conftest.py
EdDev: I wander how this works then.

tests/network/localnet/conftest.py
```diff
@@ -101,18 +114,105 @@ def vms_localnet(
...
-        server_ip=lookup_iface_status(vm=vm1, iface_name=NETWORK_NAME)[IP_ADDRESS],
-        server_port=_IPERF_SERVER_PORT,
-    ) as client:
-        assert client.is_running()
+    with start_client_vm(vms=vms_localnet) as client:
+        yield client
+
+
+@pytest.fixture(scope="class")
+def nncp_localnet_additional_ovs_bridge(
```
EdDev, line 128: How about `nncp_localnet_with_secondary_node_nic` ?

EdDev (changes requested):
> Add new localnet scenario - connectivity over migration on additional bridge

The line is too long, please follow the [recommended commit message](https://cbea.ms/git-commit/) format, especially the max line length of the subject/body (no more than 72 chars to avoid wrapping).

In general, when the commit mentions several changes, it is a hint that multiple commits are in order. Mixing the changes of the existing code with the addition of a new test makes it harder for the reviewer to go over the changes.

---

I have not finished the review, but I decided to post the findings so far to expedite things. I will continue the rest soon.

tests/network/localnet/conftest.py
```diff
@@ -101,18 +114,105 @@ def vms_localnet(
...
+    bridge_name = "localnet-additional-ovs-bridge"
+    desired_state = libnncp.DesiredState(
+        interfaces=[
+            additional_ovs_bridge_interface(
+                bridge_name=bridge_name, worker_port_name=nodes_available_nics[worker_node1.name][-1]
+            )
+        ],
+        ovn=libnncp.OVN([
+            libnncp.BridgeMappings(
+                localnet=NETWORK_NAME,
```
sbahar619, line 140: Maybe worth using a unique network name per test to keep things isolated and avoid test interference.

tests/network/localnet/conftest.py
```diff
@@ -101,18 +114,105 @@ def vms_localnet(
...
+@pytest.fixture(scope="class")
+def additional_ovs_bridge_localnet_vms(
+    additional_ovs_bridge_localnet_vma: BaseVirtualMachine, additional_ovs_bridge_localnet_vmb: BaseVirtualMachine
+) -> Generator[tuple[BaseVirtualMachine, BaseVirtualMachine]]:
+    vms = (additional_ovs_bridge_localnet_vma, additional_ovs_bridge_localnet_vmb)
+    for vm in vms:
+        vm.start()
+    for vm in vms:
+        vm.wait_for_ready_status(status=True)
+        vm.vmi.wait_for_condition(condition=vm.Condition.Type.AGENT_CONNECTED, status=vm.Condition.Status.TRUE)
```
sbahar619, line 200: The following code is repeated in two places.
It may be worth refactoring this into a utility function to avoid repetition and improve maintainability.

EdDev (commented):
Still partial review.

tests/network/localnet/conftest.py
sbahar619: I believe @yossisegev talked about the linux kernel limitation for network device names, which is indeed 15 characters.

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/pull/900/commits/8f3621574637e99a0e2f26949a28a6ffb4170df6) - answer review comments:
Change start_client_vm and start_server_vm context manager into factories and rename them to create_traffic_client and create_traffic_server.
Unite starting of localnet VMs in a ocntextmanager - running_localnet_vms.
Replace the IP_NETWORK_PREFIX constant with an IP generator.
Add unique network names.
Rename some fixtures.

tests/network/localnet/test_localnet.py
```diff
@@ -4,8 +4,22 @@
...
 
 
-@pytest.mark.ipv4
-@pytest.mark.polarion("CNV-11775")
-def test_connectivity_over_migration_between_localnet_vms(nncp_localnet, localnet_server, localnet_client):
-    migrate_vm_and_verify(vm=localnet_client.vm)
-    assert is_tcp_connection(server=localnet_server, client=localnet_client)
+class TestLocalnetDefaultBridge:
+    @pytest.mark.ipv4
+    @pytest.mark.single_nic
```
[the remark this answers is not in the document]
yossisegev: You already submitted the [PR that adds this marker](https://github.com/RedHatQE/openshift-virtualization-tests/pull/909/files), so please remove it form here.
Thamk you

yossisegev (changes requested):

tests/network/localnet/test_localnet.py
```diff
@@ -4,8 +4,22 @@
 from utilities.virt import migrate_vm_and_verify
 
 
-@pytest.mark.ipv4
-@pytest.mark.polarion("CNV-11775")
-def test_connectivity_over_migration_between_localnet_vms(nncp_localnet, localnet_server, localnet_client):
-    migrate_vm_and_verify(vm=localnet_client.vm)
-    assert is_tcp_connection(server=localnet_server, client=localnet_client)
+class TestLocalnetDefaultBridge:
```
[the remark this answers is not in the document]
EdDev: Touching existing tests is one of the reasons for slowness in reaching PR completion.

So lets see what other options we could explore to avoid it:
1. Create a new module.
2. Create a new class and do not change the existing.

I prefer 1, it has the minimum required changes.

tests/network/localnet/conftest.py
```diff
@@ -2,22 +2,34 @@
...
     WORKER_NODE_LABEL_KEY,
 )
 from utilities.infra import create_ns
 
-_IPERF_SERVER_PORT = 5201
+IP_NETWORK_PREFIX = "10.0.0"
 
 
 @pytest.fixture(scope="module")
+def vlan_id(vlan_index_number: Generator[int]) -> int:
```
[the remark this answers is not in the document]
EdDev: A commit that changes the existing test to a class may have made it easier to follow. I hope the class is not preserved, but if it will, then this is how I would prefer to review this.
Looking how an existing test was converted and then seeing the new test being added.

tests/network/localnet/conftest.py
EdDev: Fine by me.

tests/network/localnet/conftest.py
```diff
@@ -38,81 +57,185 @@ def nncp_localnet() -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
...
+        namespace=namespace_localnet_1.name, name="test-localnet-nad1", vlan_id=vlan_id, network_name=BR_EX_NETWORK_NAME
+    ) as nad:
         yield nad
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="class")
 def nad_localnet_2(
-    namespace_localnet_2: Namespace, vlan_id: int
+    nncp_localnet: Generator[libnncp.NodeNetworkConfigurationPolicy], namespace_localnet_2: Namespace, vlan_id: int
```
EdDev, line 82: `nncp_localnet` is ref but its value not used. Better ref it with a decorator marker.

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,65 @@
...
 from libs.vm.vm import BaseVirtualMachine, cloudinitdisk_storage
 from tests.network.libs import cloudinit
+from tests.network.libs.nodenetworkconfigurationpolicy import Resource
+
+_IPERF_SERVER_PORT = 5201
+NNCP_INTERFACE_TYPE_OVS_BRIDGE = "ovs-bridge"
+
+
+@contextmanager
+def running_localnet_vms(
```
EdDev, line 20: I do not understand this new contextmanager. What resource is it managing exactly?

It could be a simple helper that starts and waits for Fedora VMs to run the guest OS in parallel. (not a contextmanager)
The name should reflect that as well, e.g. `run_vms()` (it has nothing to do with the fact they are localnet ones).

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,65 @@
...
+    vms: tuple[BaseVirtualMachine, BaseVirtualMachine],
+) -> Generator[tuple[BaseVirtualMachine, BaseVirtualMachine]]:
+    for vm in vms:
+        vm.start()
+    for vm in vms:
+        vm.wait_for_ready_status(status=True)
+        vm.vmi.wait_for_condition(condition=vm.Condition.Type.AGENT_CONNECTED, status=vm.Condition.Status.TRUE)
```
EdDev, line 27: Asking again to use the new helper.

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,65 @@
...
 from libs.net import netattachdef
-from libs.net.vmspec import add_network_interface, add_volume_disk
+from libs.net.traffic_generator import Client, Server
+from libs.net.vmspec import IP_ADDRESS, add_network_interface, add_volume_disk, lookup_iface_status
 from libs.vm.affinity import new_pod_anti_affinity
 from libs.vm.factory import base_vmspec, fedora_vm
 from libs.vm.spec import CloudInitNoCloud, Interface, Metadata, Multus, Network
 from libs.vm.vm import BaseVirtualMachine, cloudinitdisk_storage
 from tests.network.libs import cloudinit
+from tests.network.libs.nodenetworkconfigurationpolicy import Resource
```
EdDev, line 13: I have no idea what you did here.

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,65 @@
...
+    )
+
+
+def additional_ovs_bridge_interface(bridge_name: str, worker_port_name: str) -> libnncp.Interface:
+    return libnncp.Interface(
+        name=bridge_name,
+        type=NNCP_INTERFACE_TYPE_OVS_BRIDGE,
+        ipv4=libnncp.IPv4(enabled=False),
+        ipv6=libnncp.IPv6(enabled=False),
+        state=Resource.Interface.State.UP,
```
EdDev, line 50: You better use `libnncp.NodeNetworkConfigurationPolicy.Interface.State.UP`, which has a more accurate context.

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,65 @@
...
+        bridge=libnncp.Bridge(
+            options=libnncp.BridgeOptions(libnncp.STP(enabled=False)),
+            port=[
+                libnncp.Port(
+                    name=worker_port_name,
+                )
+            ],
```
EdDev, line 57: Any reason this is on a one-liner?

EdDev (changes requested):

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/pull/900/commits/9f8ee67775763024d1543ec206bdf872e28ab2fd) - answer review comments:
 - Rename constant network names.
 - Change the scope of all fixtures to class scope.
 - Re-order fixture location in file for better readability.
 - Refactor running_localnet_vms name (run_vms) and functionality (no longer a contextmanager).

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-900 published

Anatw:
/verified

openshift-virtualization-tests-runner/2252
cluster: net-awax-419
iib: 960513

tests/network/localnet/conftest.py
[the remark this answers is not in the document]
EdDev: No, we are not.
We are trying to keep things intact as much as possible and do our best to make our reviewers easy to review our changes.
I have provided suggestions [here](https://github.com/RedHatQE/openshift-virtualization-tests/pull/900#discussion_r2075818030) for this specific PR changes.

As a reviewer, I can focus on one thing at a time. E.g. one changing an existing test, on adding a new test, etc.
Therefore, I would either try to limit the existing test changes or split it into separate commits.

tests/network/localnet/conftest.py
```diff
@@ -2,27 +2,46 @@
...
+BR_EX_NETWORK_NAME = "localnet-br-ex-network"
+ADDITIONAL_BRIDGE_NETWORK_NAME = "localnet-ovs-network"
 
 
 @pytest.fixture(scope="module")
+def vlan_id(vlan_index_number: Generator[int]) -> int:
+    return next(vlan_index_number)
+
+
+@pytest.fixture(scope="module")
```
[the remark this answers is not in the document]
EdDev: As the IPs are used on the VMs, it should be at the scope of the VM.

tests/network/localnet/conftest.py
```diff
@@ -2,27 +2,46 @@
...
+    running_localnet_vms,
+)
 from utilities.constants import (
     WORKER_NODE_LABEL_KEY,
 )
 from utilities.infra import create_ns
 
-_IPERF_SERVER_PORT = 5201
+BR_EX_NETWORK_NAME = "localnet-br-ex-network"
+ADDITIONAL_BRIDGE_NETWORK_NAME = "localnet-ovs-network"
```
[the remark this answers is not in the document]
EdDev: AFAIK, for localnet, we explicitly want to wire everything to the outside switch port and that happens through a NIC (or VLAN, or Bond). The bridge is just a way to wire things to it.
Therefore, calling these constants by some random name provides no meaning or no hint to what they represent.
You could name them primary and secondary, you just need to explain somehow it related to the node and not to the pod/vm.

tests/network/localnet/conftest.py
```diff
@@ -38,81 +57,185 @@ def nncp_localnet() -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
         yield nncp
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="class")
```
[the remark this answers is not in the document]
EdDev: 1. Yes, it can be a module scope.
2. It can still be class scope, it just going to be re-created.

The 1st has the advantage to re-use the same resource and make the test faster, the 2nd only has the advantage to reduce the codebase/dependencies of yet another thing.

tests/network/localnet/liblocalnet.py
EdDev: Ok, sad but nothing we can do about it.

tests/network/localnet/test_localnet.py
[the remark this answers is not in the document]
yossisegev: The marker is still here.

tests/network/localnet/conftest.py
```diff
@@ -38,81 +46,198 @@ def nncp_localnet() -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
...
+        vlan_id=vlan_id,
+        network_name=LOCALNET_BR_EX_NETWORK,
+    ) as nad:
         yield nad
 
 
-@pytest.fixture(scope="module")
-def vm_localnet_1(nad_localnet_1: netattachdef.NetworkAttachmentDefinition) -> Generator[BaseVirtualMachine]:
+@pytest.fixture(scope="class")
+def ipv4_localnet_address_pool() -> Generator[str]:
```
rnetser, line 91:
```suggestion
def ipv4_localnet_address_pool() -> str:
```
`return`  is used

tests/network/localnet/conftest.py
```diff
@@ -38,81 +46,198 @@ def nncp_localnet() -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
...
+    net_prefix = "10.0.0"
```
rnetser, line 92: can be used directly, no need to save

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,60 @@
...
 from tests.network.libs import cloudinit
 
-NETWORK_NAME = "localnet-network"
+_IPERF_SERVER_PORT = 5201
+NNCP_INTERFACE_TYPE_OVS_BRIDGE = "ovs-bridge"
 
 
-def localnet_vm(namespace: str, name: str, network: str, cidr: str) -> BaseVirtualMachine:
+def run_vms(
+    vms: tuple[BaseVirtualMachine, BaseVirtualMachine],
```
rnetser, line 16:
```suggestion
    vms: tuple[BaseVirtualMachine],
```
can be any number of VMs

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,60 @@
...
+
+
+def create_traffic_server(vm: BaseVirtualMachine) -> Server:
+    return Server(vm=vm, port=_IPERF_SERVER_PORT)
+
+
+def create_traffic_client(vms: tuple[BaseVirtualMachine, BaseVirtualMachine], network_name: str) -> Client:
+    vm_server, vm_client = vms
+    return Client(
+        vm=vm_client,
```
rnetser, line 33: can use `vms` directly (vms[0]...)

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,60 @@
...
+        vm.wait_for_ready_status(status=True)
+        vm.wait_for_agent_connected()
+    return vms
+
+
+def create_traffic_server(vm: BaseVirtualMachine) -> Server:
+    return Server(vm=vm, port=_IPERF_SERVER_PORT)
+
+
+def create_traffic_client(vms: tuple[BaseVirtualMachine, BaseVirtualMachine], network_name: str) -> Client:
```
rnetser, line 30: this is a function (not a fixture) so it should get client and server explicitly and not a tuple

tests/network/localnet/liblocalnet.py
rnetser: I agree that this does not need to be in a separate function; it just wraps `libnncp.Interface`

tests/network/localnet/test_localnet.py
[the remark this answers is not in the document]
rnetser: i agree with option 1, having the tests splits between modules with meaningful filenames makes the files cleaner and easier to find tests

rnetser (changes requested):

tests/network/localnet/liblocalnet.py
EdDev: Or even a more elegant variadic: `*vms: BaseVirtualMachine`.
The return will be then `tuple[BaseVirtualMachine]`.

tests/network/localnet/conftest.py
EdDev: @rnetser , it does return a generator. See usage.

Anatw:
Change: squashed.

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/2ba005d00950c82b420d1730c7f031a309682c8a..9449fa9a232059819a2163fff45f992c7280bf7f) - answer review comments:
 - Move interfaces back to NNCP creation instead of wrapper function.
 - Refactor typing in run_vms to accept any number of VMs. This required adding case in the 2 fixtures using this function so that mypy will be happy.
 - localnet_additional_ovs_bridge_client now get specific client and server parameters instead of general VMs.

redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-900 published

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,39 @@
...
 from tests.network.libs import cloudinit
 
-NETWORK_NAME = "localnet-network"
+_IPERF_SERVER_PORT = 5201
 
 
-def localnet_vm(namespace: str, name: str, network: str, cidr: str) -> BaseVirtualMachine:
+def run_vms(
+    *vms: BaseVirtualMachine,
+) -> tuple[BaseVirtualMachine, ...]:
```
EdDev, line 15: one-liner please

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,39 @@
...
+
+def create_traffic_client(vm_server: BaseVirtualMachine, vm_client: BaseVirtualMachine, network_name: str) -> Client:
+    return Client(
+        vm=vm_client,
+        server_ip=lookup_iface_status(vm=vm_server, iface_name=network_name)[IP_ADDRESS],
+        server_port=_IPERF_SERVER_PORT,
+    )
+
+
+def localnet_vm(namespace: str, name: str, network: str, cidr: str, network_name: str) -> BaseVirtualMachine:
```
EdDev, line 36: Please change existing code in a separate commit so it is easy to review. I wont insist on this specific additional argument now, but it is not convenient for me to review the PR like this.

- Maybe it is only me, but as a caller to this function, how do I know what is the difference between `network` and `network_name`? You need a better name, possibly for both in order to be clear.
- You have not updated the docstring.

tests/network/localnet/test_localnet.py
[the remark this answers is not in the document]
EdDev: The current format has been suggested by @rnetser , i.e. to place all fixtures in `conftest` and all helpers in one or more dedicated modules (e.g. `lib*`).
That was suggested and applied when there was only one test module, so I do not understand how a second module changes anything.
Unless we decide to change the format in the future, test modules contain only tests.

tests/network/localnet/liblocalnet.py
EdDev: Unfortunately this caused a problem with flake8.
I opened a new thread about it on the new code.

EdDev (commented):
Thanks, I finished another review cycle.

General notes:
- It is better for the author and reviewers to split existing code changes from new code, i.e. adjusting current and then adding new. I wont force require it on this PR, but on future ones I will. Up to you if you want to apply this on the current PR as well.
- Please peek at coderabbit notes, some may make sense.

tests/network/localnet/conftest.py
rnetser: missed that

tests/network/localnet/conftest.py
```diff
@@ -38,81 +47,215 @@ def nncp_localnet() -> Generator[libnncp.NodeNetworkConfigurationPolicy]:
...
 
 @pytest.fixture()
 def localnet_client(vms_localnet: tuple[BaseVirtualMachine, BaseVirtualMachine]) -> Generator[Client]:
-    vm1, vm2 = vms_localnet
-    with Client(
-        vm=vm2,
-        server_ip=lookup_iface_status(vm=vm1, iface_name=NETWORK_NAME)[IP_ADDRESS],
-        server_port=_IPERF_SERVER_PORT,
+    with create_traffic_client(
+        vm_server=vms_localnet[0], vm_client=vms_localnet[1], network_name=LOCALNET_BR_EX_NETWORK
```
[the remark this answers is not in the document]
rnetser: Following @EdDev comment about meaningful names, why not have `vms_localnet` return a dict with meaningful keys, e.g:
{"server": ..., "client":...}?

tests/network/localnet/liblocalnet.py
rnetser: ys, this can be resolved with the change to the arg names

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,39 @@
...
+        vm.wait_for_ready_status(status=True)
+        vm.wait_for_agent_connected()
+    return vms
+
+
+def create_traffic_server(vm: BaseVirtualMachine) -> Server:
+    return Server(vm=vm, port=_IPERF_SERVER_PORT)
+
+
+def create_traffic_client(vm_server: BaseVirtualMachine, vm_client: BaseVirtualMachine, network_name: str) -> Client:
```
rnetser, line 28: nitpick but wouldn't naming `server_vm` and `client_vm` will be more accurate?

tests/network/localnet/test_localnet.py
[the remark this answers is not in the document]
rnetser: just so we are on the same page:
```
tests/network/localnet
   - conftest.py -> only fixtures
   - liblocalnet.py -> only functions related to localnet
   - test_default_bridge.py (you do not have to include `localnet` in the name because it is under localnet
   - test_additional_bridge.py (btw, as mentioned, `additional` is vague)
```

rnetser (changes requested):

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/9449fa9a232059819a2163fff45f992c7280bf7f..0bfb7a230b1c6741b8ad38b7205d249ac3e1b762) - answer review comments:
 - Split the br-ex test and the ovs test to seperate modules.
 - Revert all fixtures to module scope.
 - Refactor run_vms to receive and return a list of VMs.
 - Refactor all "additional" in names (regarding additional bridge -> ovs bridge).

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/pull/900/commits/ff722413a0f0237cdc255d32274a89f4945bf418) - Rename network parameters in localnet_vm:
    
 - Rename to network nad_name and network_name to spec_logical_network.
 - Add spec_logical_network to the docstring.

tests/network/localnet/liblocalnet.py
Anatw: https://github.com/RedHatQE/openshift-virtualization-tests/pull/900/commits/ff722413a0f0237cdc255d32274a89f4945bf418

Anatw:
Change: rebase on main.

tests/network/localnet/conftest.py
```diff
@@ -57,62 +70,199 @@ def vlan_id(vlan_index_number: Generator[int]) -> int:
...
-        vm.start()
-    for vm in vms:
-        vm.wait_for_ready_status(status=True)
-        vm.vmi.wait_for_condition(condition=vm.Condition.Type.AGENT_CONNECTED, status=vm.Condition.Status.TRUE)
-    yield vms
+) -> Generator[dict[str, BaseVirtualMachine]]:
+    vms = [vm_localnet_1, vm_localnet_2]
+    running_vms = run_vms(vms=vms)
+    yield {SERVER_VM: running_vms[0], CLIENT_VM: running_vms[1]}
+
```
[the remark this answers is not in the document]
Anatw: Ack.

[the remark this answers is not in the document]
EdDev: Wrong, please do not do this!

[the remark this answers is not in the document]
EdDev: > Wrong, please do not do this!

I meant, converting to a dict is wrong and should not be done: `{SERVER_VM: running_vms[0], CLIENT_VM: running_vms[1]}`

tests/network/localnet/liblocalnet.py
EdDev: What exactly you changed?

tests/network/localnet/conftest.py
```diff
@@ -57,62 +70,197 @@ def vlan_id(vlan_index_number: Generator[int]) -> int:
...
-    vms = (vm_localnet_1, vm_localnet_2)
-    for vm in vms:
-        vm.start()
-    for vm in vms:
-        vm.wait_for_ready_status(status=True)
-        vm.vmi.wait_for_condition(condition=vm.Condition.Type.AGENT_CONNECTED, status=vm.Condition.Status.TRUE)
-    yield vms
+) -> Generator[dict[str, BaseVirtualMachine]]:
+    server_vm, client_vm = run_vms(vms=[vm_localnet_1, vm_localnet_2])
+    yield {SERVER_VM: server_vm, CLIENT_VM: client_vm}
```
EdDev, line 133: Please return the tuple, i.e. `yield server_vm, client_vm`.
You can rename the fixture to `server_client_localnet_vms` if it helps..

tests/network/localnet/conftest.py
```diff
@@ -57,62 +70,197 @@ def vlan_id(vlan_index_number: Generator[int]) -> int:
...
+    ) as vm:
+        yield vm
+
+
+@pytest.fixture(scope="module")
+def ovs_bridge_localnet_vms(
+    vm1_ovs_bridge_localnet: BaseVirtualMachine, vm2_ovs_bridge_localnet: BaseVirtualMachine
+) -> Generator[dict[str, BaseVirtualMachine]]:
+    server_vm, client_vm = run_vms(vms=[vm1_ovs_bridge_localnet, vm2_ovs_bridge_localnet])
+    yield {SERVER_VM: server_vm, CLIENT_VM: client_vm}
```
EdDev, line 244: See prev comment about this.
Also, please keep the name format consistent.

tests/network/localnet/liblocalnet.py
```diff
@@ -1,15 +1,37 @@
...
 from libs.vm.spec import CloudInitNoCloud, Interface, Metadata, Multus, Network
 from libs.vm.vm import BaseVirtualMachine, cloudinitdisk_storage
 from tests.network.libs import cloudinit
 
-NETWORK_NAME = "localnet-network"
+_IPERF_SERVER_PORT = 5201
 
 
-def localnet_vm(namespace: str, name: str, network: str, cidr: str) -> BaseVirtualMachine:
+def run_vms(vms: list[BaseVirtualMachine]) -> list[BaseVirtualMachine]:
```
EdDev, line 13: What made you change from tuple to list?

EdDev (commented):
Thanks.
A few general comments:
- The second commit does not help the review, so I could not use it. At this stage, it would be beneficial to just squash it.
- The differentiation in naming between the prev and new test is `default_bridge` vs `ovs_bridge`. But that sounds odd to me, maybe consider `default bridge` vs `secondary bridge`.

Anatw:
/verified

openshift-virtualization-tests-runner/2278
net-awax-419
iib: 960513

tests/network/localnet/conftest.py
```diff
@@ -2,27 +2,40 @@
...
 
-_IPERF_SERVER_PORT = 5201
+LOCALNET_BR_EX_NETWORK = "localnet-br-ex-network"
+LOCALNET_OVS_BRIDGE_NETWORK = "localnet-ovs-network"
+NNCP_INTERFACE_TYPE_OVS_BRIDGE = "ovs-bridge"
+SERVER_VM = "server"
+CLIENT_VM = "client"
+
+
+# ------------------ Default bridge (BR-EX) fixtures: ---------------------
```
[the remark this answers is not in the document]
Anatw: Ack. Can you please elaborate why? I find it easier to see the different fixture usage.

Anatw:
Change: rebase on main.

tests/network/localnet/conftest.py
Anatw: Ack, for both.

tests/network/localnet/conftest.py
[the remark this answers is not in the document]
EdDev: In a code source file I expect to see code and in some cases comments that are a must (provide context) to a specific entity (e.g. function).

If a "splitter" is needed, then the names should be better or the file structure should reflect the separation.
I guess this reflects one of the cons to having all fixtures in one place.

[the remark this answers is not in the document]
rnetser: All fixtures related to the feature are placed together, how is this a con? fixtre names need to be descriptive

Anatw (commented):
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/cc9768b12be8722842db0a4ae7adfd942660cf9a..e7c45b1578b8d8b02cf98ad55e7b18e6ef92d6e2):
 - Rename vms_localnet and ovs_bridge_localnet_vms to reflect server and client order.
 - Change the return type from dictionary to a tuple.

tests/network/localnet/conftest.py
```diff
@@ -57,62 +65,196 @@ def vlan_id(vlan_index_number: Generator[int]) -> int:
...
     vm_localnet_1: BaseVirtualMachine, vm_localnet_2: BaseVirtualMachine
 ) -> Generator[tuple[BaseVirtualMachine, BaseVirtualMachine]]:
-    vms = (vm_localnet_1, vm_localnet_2)
-    for vm in vms:
-        vm.start()
-    for vm in vms:
-        vm.wait_for_ready_status(status=True)
-        vm.vmi.wait_for_condition(condition=vm.Condition.Type.AGENT_CONNECTED, status=vm.Condition.Status.TRUE)
-    yield vms
+    server_vm, client_vm = run_vms(vms=[vm_localnet_1, vm_localnet_2])
```
rnetser, line 127: any reason why to save `server_vm, client_vm`? call `run_vms` and then yield `vm_localnet_1, vm_localnet_2`

tests/network/localnet/conftest.py
```diff
@@ -57,62 +65,196 @@ def vlan_id(vlan_index_number: Generator[int]) -> int:
...
+        spec_logical_network=LOCALNET_OVS_BRIDGE_NETWORK,
+    ) as vm:
+        yield vm
+
+
+@pytest.fixture(scope="module")
+def ovs_bridge_server_client_localnet_vms(
+    vm1_ovs_bridge_localnet: BaseVirtualMachine, vm2_ovs_bridge_localnet: BaseVirtualMachine
+) -> Generator[tuple[BaseVirtualMachine, BaseVirtualMachine]]:
+    server_vm, client_vm = run_vms(vms=[vm1_ovs_bridge_localnet, vm2_ovs_bridge_localnet])
```
rnetser, line 237: ditto

tests/network/localnet/liblocalnet.py
rnetser: Not sure I understand what's changed. I was referring to the arg names.
So I think that `server_vm` is more clear than `vm_server` because the function gets a VM and not a server.
(as mentioned above, this is a nitpick, please decide)

tests/network/localnet/test_ovs_bridge.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from libs.net.traffic_generator import is_tcp_connection
+from utilities.virt import migrate_vm_and_verify
+
+
+@pytest.mark.ipv4
+@pytest.mark.usefixtures("nncp_localnet_on_secondary_node_nic")
+@pytest.mark.polarion("CNV-11905")
+def test_connectivity_over_migration_between_ovs_bridge_localnet_vms(
```
rnetser, line 10: test name is a bit confusing as `migrate_vm_and_verify` waits for migration to end. connectivity check is done after the migration and not during

tests/network/localnet/conftest.py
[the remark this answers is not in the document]
EdDev: One feature may have multiple scenarios and subsets of them, like in this case. I.e. one package with multiple modules. It does make sense (to me) to have fixtures specific to a specific module in that module and common ones in conftest. Each option has pros and cons.

tests/network/localnet/conftest.py
Anatw: I think it is more readable this way - it is clear what is the server and what is the client.

tests/network/localnet/conftest.py
Anatw: Let's discuss this in the first comment, and I'll change it here too if needed.

tests/network/localnet/liblocalnet.py
Anatw: Neither do I, we went through several iterations on this one. I'll change it to a tuple.

tests/network/localnet/liblocalnet.py
Anatw: This was the original signature:
`def create_traffic_client(vms: tuple[BaseVirtualMachine, BaseVirtualMachine], network_name: str) -> Client:`
I'll rename to start with the function and end with vm.

tests/network/localnet/conftest.py
EdDev: Originally, `run_vms` returned a variable length tuple, so it was not possible to return it as is due to the typing limitations.
If the fixture will return a list, that will allow direct return.

However, I prefer we have the function helper return a tuple (feels more correct to me) and the fixture returning only two members as it shows up now.
This is not converging, so I guess we can handle it as a follow up.

tests/network/localnet/test_ovs_bridge.py
Anatw: Connectivity is done over migration - the connection is initiated before the migration, and it should stay up until after the migration.
We will have another scenario to test new connections after migration, but this is not the current scenario.

EdDev: Connectivity is checked during migration in the sense it is checked before it starts [1] and confirmed it did not break during the migration (i.e. the connection is still there).

[1] Fixture `localnet_ovs_bridge_client` does a `assert client.is_running()`.

Anatw (commented):
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/e7c45b1578b8d8b02cf98ad55e7b18e6ef92d6e2..8f9d1b87ca9872064e961db5f9a7d83fcdb5a64f):
 - run_vms accept and return tuple and not list.
 - Refactor parameters in create_traffic_client.

Anatw:
/verified

openshift-virtualization-tests-runner/2280
net-awax-419
iib: 960513

Anatw:
> The PR content is great, thanks.
> 
> Please fix the commit message and I will ack.
> 
> Note, this PR will have to wait for #932 to get merged first, as it has higher priority. Hope the rebase will not be too hard.

Thank you.
Why should it wait for #932? We will need to backport this scenario in any case before changing it to fit C-UDN.

EdDev:
> Thank you.
> Why should it wait for #932? We will need to backport this scenario in any case before changing it to fit C-UDN.

If we merge this one first, it will delay the other one due to the need to rebase and resolve conflicts.
The CUDN API is a blocker to 4.19 release at a higher priority than this one. This change is localnet specific and the CUDN is the API we actually are suppose to confirm working with.

EdDev:
@Anatw , please also resolve all the possible comments that have been agreed on. Any other comments need your attention, including the AI ones.

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/8f9d1b87ca9872064e961db5f9a7d83fcdb5a64f..0e978b9fe73f43fc3491c7b94160613c389f4bba):
 - Add docstring to localnet_nad (per coderabit suggestion).

Anatw:
Change: rebase on main.

cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-900 published

tests/network/localnet/conftest.py
```diff
@@ -2,27 +2,36 @@
...
+)
 from utilities.constants import (
     WORKER_NODE_LABEL_KEY,
 )
 from utilities.infra import create_ns
 
-_IPERF_SERVER_PORT = 5201
+LOCALNET_BR_EX_NETWORK = "localnet-br-ex-network"
+LOCALNET_OVS_BRIDGE_NETWORK = "localnet-ovs-network"
+NNCP_INTERFACE_TYPE_OVS_BRIDGE = "ovs-bridge"
```
yossisegev, line 26: 1. Why the `NNCP_` prefix?
2. There is actually no need to define this string constant, you can and should import it from [utilities/constants](https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/utilities/constants.py#L140).

Anatw:
/verified

openshift-virtualization-tests-runner/2313/
net-awax-419
iib: 968700
OCP version: 4.19.0-0.nightly-2025-04-24-005837

tests/network/localnet/conftest.py
```diff
@@ -60,60 +69,189 @@ def cudn_localnet(
...
+        vlan_id=vlan_id,
+        physical_network_name=LOCALNET_BR_EX_NETWORK,
     ) as cudn:
         cudn.wait_for_status_success()
         yield cudn
 
 
+@pytest.fixture(scope="module")
+def ipv4_localnet_address_pool() -> Generator[str]:
+    return (f"10.0.0.{host_value}/24" for host_value in range(1, 254))
```
yossisegev, line 83: It's not a must, but I think that in order to conform to the common convention, the range should start in 2 and not in 1. `x.y.z.1` is usually used for the default gateway.

yossisegev (changes requested):

tests/network/localnet/conftest.py
Anatw: The constant and it's naming are here to have a meaningful name. If you think that this name is not clear or meaningful enough, I'll be happy to refactor it.

tests/network/localnet/conftest.py
Anatw: The common convention in our tests is to use host ID 1 for the first VM :) But I don't mind changing it, it is a good practice to leave 1 available for the default gateway, even though we don't use it in this network we define for our tests.

yossisegev: Yeah, you make a good point, and for some reason I haven't noticed so far that we use IP ranges starting from 1.
So again - I'll leave it to you to decide whether to start it from 2 or leave it as it is.
Anyway you can resolve this comment.

tests/network/localnet/conftest.py
yossisegev: I'm not sure how meaningful this name is, but OTOH - you (and others) might find this naming clear and meaningful, while I (and different others :) ) may not. So I'm leaving this one for you as well to decide whether you choose to change it or not.
I'm fine with whatever you decide> Don't forget to resolve this conversation too.

tests/network/localnet/conftest.py
```diff
@@ -60,60 +69,189 @@ def cudn_localnet(
...
 ) -> Generator[tuple[BaseVirtualMachine, BaseVirtualMachine]]:
-    vms = (vm_localnet_1, vm_localnet_2)
-    for vm in vms:
-        vm.start()
-    for vm in vms:
-        vm.wait_for_ready_status(status=True)
-        vm.vmi.wait_for_condition(condition=vm.Condition.Type.AGENT_CONNECTED, status=vm.Condition.Status.TRUE)
-    yield vms
+    server_vm, client_vm = run_vms(vms=(vm_localnet_1, vm_localnet_2))
+    yield (server_vm, client_vm)
```
EdDev, line 123: Any reason this uses `yield` and not a regular `return`?

tests/network/localnet/conftest.py
```diff
@@ -60,60 +69,189 @@ def cudn_localnet(
...
+        yield nncp
+
+
+@pytest.fixture(scope="module")
+def cudn_localnet_ovs_bridge(
+    vlan_id: int,
+    namespace_localnet_1: Namespace,
+) -> Generator[libcudn.ClusterUserDefinedNetwork]:
+    with localnet_cudn(
+        name=f"{LOCALNET_OVS_BRIDGE_NETWORK}-cudn",
```
EdDev, line 190: Naming inconsistency from the other cUDN created already.

tests/network/localnet/conftest.py
```diff
@@ -60,60 +69,189 @@ def cudn_localnet(
...
+        match_labels=LOCALNET_TEST_LABEL,
+        vlan_id=vlan_id,
+        physical_network_name=LOCALNET_OVS_BRIDGE_NETWORK,
+    ) as cudn:
+        yield cudn
```
EdDev, line 195: Existing cUDN is waiting for the network to be deployed.
I.e. `cudn.wait_for_status_success()` is missing here.

tests/network/localnet/conftest.py
```diff
@@ -60,60 +69,189 @@ def cudn_localnet(
...
+
+
+@pytest.fixture(scope="module")
+def vm1_ovs_bridge_localnet(
```
EdDev, line 199: Please sync on the same naming convention the VMs have already.

EdDev (commented):
Thanks.
There are only a few small issues from my side.

There is also a need to wrap the commit message to 72 chars.

tests/network/localnet/test_ovs_bridge.py
rnetser: should have been igrate_vm_and_verify(vm=localnet_ovs_bridge_client.vm, wait_for_migration_success=False) but will not block the pr

EdDev: Why `wait_for_migration_success=False` ?
We need the migration to finish and see that the connection has survived during that process.
To be clear, if the connection breaks during the migration, it will not auto-recover and we will see it when the migration finishes.

Anatw:
/verified

openshift-virtualization-tests-runner/2315
net-awax-419
iib: 968700
OCP version: 4.19.0-0.nightly-2025-04-24-005837

tests/network/localnet/test_ovs_bridge.py
Anatw: Since is was approved by both Edi and Ruth after this conversation ended, I'm resolving this comment.

EdDev:
> /verified
> 
> openshift-virtualization-tests-runner/2315 net-awax-419 iib: 968700 OCP version: 4.19.0-0.nightly-2025-04-24-005837

Verifying on a nightly is not valid, please validate on the latest RC (at the moment, it is RC.2).

yossisegev (approved):

Anatw:
/verified

openshift-virtualization-tests-runner/2334/
net-awax-419-2
iib: 972571
OCP version: 4.19.0-rc.2

PR #973, opened 2025-05-15
Update internal-http container image

stesrn:
This patch updates the internal-http container image used in cnv-tests. The new version v1.1.0 of internal-http multi-arch image manifest also includes container image for s390x architecture in addition to amd64(x86_64) and arm64(aarch64)

##### jira-ticket:
NONE

## Summary by CodeRabbit

- **Chores**
  - Updated the container image version used for the internal HTTP server deployment to v1.1.0.


redhat-cnv-qe-bot1:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-973 published

jpeimer (approved):

rnetser:
@stesrn please note that the fedora image was reverted

tests/storage/conftest.py
```diff
@@ -82,7 +82,7 @@
         "containers": [
             {
                 "name": "http",
-                "image": "quay.io/openshift-cnv/qe-cnv-tests-internal-http:v1.0.0",
+                "image": "quay.io/openshift-cnv/qe-cnv-tests-internal-http:v1.1.0",
```
rnetser, line 85: @stesrn please note that the fedora image was reverted

rnetser: @stesrn i just want to make sure we do not have any surprises with the images that may affect other teams. Smoke tests may not enough (as the tests do not use all images)

jpeimer: This is only used in `images_internal_http_server`, which is only used in several storage tests, so if these tests are fine - we should be good

stesrn: @jpeimer Thank you very much. I will run those tests to make sure the new container image is good

PR #1029, opened 2025-05-20
[virt] drop rwo storage related skips

vsibirsk:
##### Short description:
  - drop storage related programmatic skips
  - added new "rwx_default_storage" marker
  - added new virt sanity verification step for default storage
 
##### jira-ticket:
https://issues.redhat.com/browse/CNV-57837

## Summary by CodeRabbit

- **Refactor**
	- Reorganized and expanded test markers to include a new "rwx_default_storage" category.
	- Simplified test signatures by removing unused parameters related to storage access mode fixtures.
	- Updated test markers and decorators to categorize tests under the new "rwx_default_storage" marker.

- **Tests**
	- Added verification to ensure default storage supports RWX access mode before running migration-related tests.
	- Improved test categorization and maintainability by consistently applying the new storage-related marker across multiple test modules.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -25,29 +25,29 @@ markers =
     node_remediation_ipmi_enabled: Destructive NodeHealthCheck with SNR/FAR on IPMI-enabled clusters
 
     # Cluster markers
-    ## Architecture
-    arm64: tests that can run on ARM-based cluster
+    ## Architecture support
+    arm64: Tests that can run on ARM-based cluster
 
-    ## Hardware
-    special_infra: tests that requires special infrastructure. e.g. sriov, gpu etc.
-    gpu: tests that require cluster with gpu
-    sriov: tests that require sriov net-cards on nodes
-    single_nic: tests that dont require mutli-nic nodes, required for conformance tests for new archs, platforms etc.
+    ## Hardware requirements
+    special_infra: Tests that requires special infrastructure. e.g. sriov, gpu etc.
+    gpu: Tests that require cluster with gpu cards
+    sriov: Tests that require sriov net-cards on nodes
+    single_nic: Tests that don`t require mutli-nic nodes, required for conformance tests for new archs, platforms etc.
 
-    ## Configuration
-    ipv4: IPv4 based test
-    ipv6: IPv6 based test
-    dpdk: Tests that requires dpdk
-    swap: tests that require SWAP active on nodes
-    cpu_manager: tests that require cpu manager on nodes
-    numa: tests that require numa configured on nodes
-    hugepages: tests that require nodes with hugepages
-    service_mesh: tests that require the service mesh operator to be installed
-    jumbo_frame: tests that require network configurations supporting jumbo frames
-    rwx_default_storage: tests that require RWX storage
+    ## Configuration requirements
+    ipv4: Tests that require IPv4
+    ipv6: Tests that require IPv6
+    dpdk: Tests that require dpdk
+    swap: Tests that require SWAP active on nodes
+    cpu_manager: Tests that require cpu manager on nodes
+    numa: Tests that require numa configured on nodes
+    hugepages: Tests that require nodes with hugepages
+    service_mesh: Tests that require the service mesh operator to be installed
+    jumbo_frame: Tests that require network configurations supporting jumbo frames
+    rwx_default_storage: Tests that require RWX storage
 
-    ## Resources
-    high_resource_vm: tests using VM requirening a lot of resources (like Windows OS VMs, hight performance VMs, etc)
+    ## Resources requirements
+    high_resource_vm: Tests that create VM using a lot of resources (like Windows OS VMs, hight performance VMs, etc)
 
     # CI
     smoke: Mark tests as smoke tests
--- a/pytest.ini
+++ b/pytest.ini
@@ -14,7 +14,6 @@ markers =
     # Data collection markers
     skip_must_gather_collection: skip must gather collection on failures
     gpu: indicates collecting relevant namespace data on failure
-    swap: indicates collecting relevant namespace data on failure
     # Test types
     destructive: Destructive tests
     sap_hana: SAP HANA tests
@@ -38,6 +37,7 @@ markers =
     jumbo_frame: tests that require network configurations supporting jumbo frames
     single_nic: tests that dont require mutli-nic nodes, required for conformance tests for new archs, platforms etc.
     arm64: tests that require ARM-based cluster supported
+    migration: tests that involve migration (e.g. live migrating VM, hotplugging CPU, etc)
     # CI
     smoke: Mark tests as smoke tests
     ci: Mark tests as CI tests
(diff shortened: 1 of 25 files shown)


rnetser (dismissed):

pytest.ini
```diff
@@ -14,7 +14,6 @@ markers =
     # Data collection markers
     skip_must_gather_collection: skip must gather collection on failures
     gpu: indicates collecting relevant namespace data on failure
-    swap: indicates collecting relevant namespace data on failure
```
jpeimer, line 17: Are you sure you want to remove this?

tests/virt/cluster/common_templates/centos/test_centos_os_support.py
```diff
@@ -146,16 +146,13 @@ def test_pause_unpause_vm(self, golden_image_vm_object_from_template_multi_cento
...
     @pytest.mark.dependency(
         name=f"{TESTS_CLASS_NAME}::migrate_vm_and_verify",
         depends=[f"{TESTS_CLASS_NAME}::vm_expose_ssh"],
     )
-    def test_migrate_vm(
-        self,
-        skip_access_mode_rwo_scope_class,
-        golden_image_vm_object_from_template_multi_centos_multi_storage_scope_class,
-    ):
+    def test_migrate_vm(self, golden_image_vm_object_from_template_multi_centos_multi_storage_scope_class):
```
jpeimer, line 155: So if we pass the `-m migration`, but this test depends on `vm_expose_ssh` which does not have the migration marker, the `test_migrate_vm` will be skipped because `vm_expose_ssh` didn't run?

pytest.ini
vsibirsk: it's a duplicate
https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/pytest.ini#L30
we have same marker at lines 17 and 30 (so removing one of them)

tests/virt/cluster/common_templates/centos/test_centos_os_support.py
vsibirsk: yes. if running with only `migration` marker, all common template tests will be skipped.

pytest.ini
vsibirsk: in fact same for `gpu` marker :) lines 16 and 32

tests/virt/cluster/common_templates/centos/test_centos_os_support.py
jpeimer: also when running with `-m 'migration and virt'` - we'll hit the same issue, no? 
then how the `migration` marked can be used? (asking for a friend, because I'm facing the same issue :) )

vsibirsk: the main usage is to trigger virt-sanity validation when any migration-marked test collected for the run

jpeimer: virt-sanity validation will exit the pytest run, so what if you have an RWO storage and you still want to run the other tests on it?

vsibirsk: if we're talking about t2 lanes, we should always have clusters with rwx storage. that's a req.
for local runs, pass `not migration` (for e.g. `-m virt and not migration`) and sanity check will skip migration validation. or pass `--skip-virt-sanity-check` since you know what you're doing with your local runs :)

jpeimer: Ok.. so for storage, this marked will not be usable, as we may run on a non-rwx storages...

vsibirsk: you can still use the marker (if you want). virt sanity validations are not triggered on storage tests

jpeimer: mm not sure how can we make it work when migration test depends on another test

pytest.ini
dshchedr: since you anyway touching it - can you please remove "gpu" as well

pytest.ini
```diff
@@ -38,6 +37,7 @@ markers =
     jumbo_frame: tests that require network configurations supporting jumbo frames
     single_nic: tests that dont require mutli-nic nodes, required for conformance tests for new archs, platforms etc.
     arm64: tests that require ARM-based cluster supported
+    migration: tests that involve migration (e.g. live migrating VM, hotplugging CPU, etc)
```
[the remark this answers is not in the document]
dshchedr: Since we are using it for checking RWX for default storage class (and theoretically, this marker can be used in other places not only for migration tests) - how about this name: `rwx_default_storage`

tests/virt/cluster/common_templates/centos/test_centos_os_support.py
vsibirsk: renamed marker. maybe now can be of use for storage tests as well

rnetser (dismissed):

pytest.ini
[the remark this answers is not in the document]
vsibirsk: did couple of changes here:
 - renamed marker
 - re-organized markers in `pytest.ini` and updated descriptions

dshchedr (approved):

tests/virt/conftest.py
```diff
@@ -97,6 +107,8 @@ def _verify_hugepages_1gi(_workers):
             _verify_sriov(_sriov_workers=sriov_workers)
         if any(item.get_closest_marker("hugepages") for item in request.session.items):
             _verify_hugepages_1gi(_workers=workers)
+        if any(item.get_closest_marker("migration") for item in request.session.items):
```
dshchedr, line 110:
```suggestion
        if any(item.get_closest_marker("rwx_default_storage") for item in request.session.items):
```

dshchedr (changes requested):

cnv-qe-bot1 (approved):

tests/virt/cluster/common_templates/centos/test_centos_os_support.py
dshchedr: It is tricky, but without changes we can't get rid of skips..

you can add `rwx_default_storage` marker only for migration test. And when you run tests, if you know that the default storage class is RWO - you can pass  `-m "..any_marker..  and not rwx_default_storage"` -- in this case migration test will be skipped.

vsibirsk: after the initial commit (and marker renaming) the current purpose of this new `rwx_default_storage` marker is to mark test with cluster config requirement and trigger appropriate sanity validation.
it's not intended to be used to run only migration tests (can be used to skip those like @dshchedr shown above)

dshchedr (approved):

pytest.ini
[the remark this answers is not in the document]
rnetser: looks good, thanks

PR #1081, opened 2025-05-27
fix: skip tablet tests on s390x architecture

nestoracunablanco:
Tablet-related tests are not required on the s390x architecture, as the tablet device is not supported on this platform.

To exclude these tests, the change proposes the use of a skip_s390x marker. For example, a test suite can be run with the following expression:

-m "tier2 and virt and not special_infra and not skip_s390x"

##### jira-ticket:

NONE

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -27,6 +27,7 @@ markers =
     # Cluster markers
     ## Architecture support
     arm64: Tests that can run on ARM-based cluster
+    skip_s390x: Tests that are disabled on s390x-based cluster
 
     ## Hardware requirements
     special_infra: Tests that requires special infrastructure. e.g. sriov, gpu etc.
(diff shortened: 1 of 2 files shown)


pytest.ini
```diff
@@ -27,6 +27,7 @@ markers =
     # Cluster markers
     ## Architecture support
     arm64: Tests that can run on ARM-based cluster
+    skip_s390x: Tests that are disabled on s390x-based cluster
```
dshchedr, line 30: instead of skipping - all s390x tests should be marked explicitly, same as arm64

dshchedr (changes requested):

pytest.ini
nestoracunablanco: Just to confirm my understanding: for the compute tests, we need to disable around 10 tests based on my estimation. Does that mean we should explicitly mark the remaining ~300 tests that should not be skipped as s390x?

PR #1095, opened 2025-05-29
tests, net, general: Drop usage of 'running_vm'

EdDev:

`running_vm` is performing multiple checks to assure correct VM start. However, these checks are not required for the network tests.

For the network tests, a successful VM start is one that:
- Reaches ready status (i.e. a VMI in running phase).
- Reaches `AgentConnected` condition status (i.e. cloud-init stage succeeded based on the Fedora image setup).

This is an optimization to reduce observed flakiness on VM startup and simplify the overall maintenance (e.g. the use of a common helper that covers many aspects of a VM startup cycle forces all its users to the same logic, even though that logic is not of interest).


EdDev:
Verified by running locally against a PSI cluster: `uv run pytest --jira --junitxml=xunit_results.xml --pytest-log-file=pytest-tests.log --leftovers-collector --tb=native -rs -s -o log_cli=true --skip-deprecated-api-test --cluster-sanity-skip-storage-check tests/network/general/
`

Result: `12 passed, 1 skipped, 1 deselected, 49 warnings in 1145.23s (0:19:05)`

/verified

yossisegev (approved):

PR #1107, opened 2025-05-29
Quarantine RDP test until we fix the cleint side

yossisegev:
## Summary by CodeRabbit

- **Tests**
  - Disabled a specific test related to remote desktop protocol for exposed virtual machines.

### DIFF
--- a/tests/network/rdp/test_rdp_for_exposed_vm_svc.py
+++ b/tests/network/rdp/test_rdp_for_exposed_vm_svc.py
@@ -75,6 +75,9 @@ def rdp_pod(workers_utility_pods, rdp_vm):
     ],
     indirect=True,
 )
+@pytest.mark.xfail(
+    reason="RDP test case must be fixed and implement a reliable client; tracked in CNV-43997", run=False
+)
 def test_rdp_for_exposed_win_vm_as_node_port_svc(
     rdp_vm,
     rdp_pod,


tests/network/rdp/test_rdp_for_exposed_vm_svc.py
```diff
@@ -75,6 +75,7 @@ def rdp_pod(workers_utility_pods, rdp_vm):
     ],
     indirect=True,
 )
+@pytest.mark.xfail(reason="CNV-43997", run=False)
```
RoniKishner, line 78: It looks like this mark will cause the test to always fail with the reason "CNV-43997".
Did you perhaps mean the following? 
```suggestion
@pytest.mark.xfail(reason="CNV-43997", run=False)
```

yossisegev: @RoniKishner I don't see the difference between my change and your suggestion.
You probably mean `mark.skip`, right?
Anyway - you are partially correct :) : The test will indeed fail, but won't be marked as fail, but as skipped.
But anyway, skip seems like a better option, because it will save the precious time spent on attempting to execute this test and fail.

rnetser:
```
@pytest.mark.xfail(reason="CNV-43997", run=False)
```
is the right approach, the test wil not be executed thus not marked as failed.
The reason should be more explanation; e.g `CNV-43997 - client side fix....`

tests/network/rdp/test_rdp_for_exposed_vm_svc.py
```diff
@@ -75,6 +75,7 @@ def rdp_pod(workers_utility_pods, rdp_vm):
     ],
     indirect=True,
 )
+@pytest.mark.skip(reason="CNV-43997")
```
rnetser, line 78: please use `xfail(...., run=False)` and not skip
also, please provide a reason (not just the jira id); quarantined tests will not be executed but should not be skipped

rnetser (changes requested):

tests/network/rdp/test_rdp_for_exposed_vm_svc.py
```diff
@@ -75,6 +75,7 @@ def rdp_pod(workers_utility_pods, rdp_vm):
     ],
     indirect=True,
 )
+@pytest.mark.xfail(reason="RDP test case must be fixed and implement a reliable client", run=False)
```
rnetser, line 78: please add the jira id

rnetser (changes requested):

PR #1110, opened 2025-05-30
 Add test to validate openshift.io/required-scc annotation for all HCO deployments

geetikakay:
##### Short description:
Refer : https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3284

##### jira-ticket:
https://issues.redhat.com/browse/CNV-60087

## Summary by CodeRabbit

- **Tests**
  - Added automated tests to ensure all HyperConverged Operator deployments have the correct security context constraint annotation, improving security compliance and reporting any missing or incorrect annotations.

### DIFF
--- a/tests/install_upgrade_operators/security/scc/test_cnv_deployment_required_scc.py
+++ b/tests/install_upgrade_operators/security/scc/test_cnv_deployment_required_scc.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+"""
+Test to verify all HCO deployments have 'openshift.io/required-scc' annotation.
+"""
+
+import pytest
+from ocp_resources.deployment import Deployment
+
+from utilities.constants import ALL_CNV_DEPLOYMENTS_NO_HPP_POOL
+
+REQUIRED_SCC_ANNOTATION = "openshift.io/required-scc"
+REQUIRED_SCC_VALUE = "restricted-v2"
+
+
+@pytest.fixture(scope="module")
+def required_scc_deployment_check(admin_client, hco_namespace):
+    missing_required_scc_annotation = []
+    incorrect_required_scc_annotation_value = {}
+
+    for dp in (
+        Deployment(client=admin_client, name=name, namespace=hco_namespace.name)
+        for name in ALL_CNV_DEPLOYMENTS_NO_HPP_POOL
+    ):
+        scc = dp.instance.spec.template.metadata.annotations.get(REQUIRED_SCC_ANNOTATION)
+
+        if scc is None:
+            missing_required_scc_annotation.append(dp.name)
+        elif scc != REQUIRED_SCC_VALUE:
+            incorrect_required_scc_annotation_value[dp.name] = scc
+
+    return {
+        "missing_required_scc_annotation": missing_required_scc_annotation,
+        "incorrect_required_scc_annotation_value": incorrect_required_scc_annotation_value,
+    }
+
+
+@pytest.mark.polarion("CNV-11964")
+def test_deployments_missing_required_scc_annotation(required_scc_deployment_check):
+    assert not required_scc_deployment_check["missing_required_scc_annotation"], (
+        f"Deployments missing {REQUIRED_SCC_ANNOTATION} annotation: "
+        f"{required_scc_deployment_check['missing_required_scc_annotation']}"
+    )
+
+
+@pytest.mark.polarion("CNV-11965")
+def test_deployments_with_incorrect_required_scc(required_scc_deployment_check):
+    assert not required_scc_deployment_check["incorrect_required_scc_annotation_value"], (
+        f"Deployments incorrect {REQUIRED_SCC_ANNOTATION} annotation : "
+        f"{required_scc_deployment_check['incorrect_required_scc_annotation_value']}"
+    )


tests/install_upgrade_operators/security/scc/test_cnv_deployment_required_scc.py
```diff
@@ -0,0 +1,56 @@
...
+        f"{required_scc_deployment_check['missing_required_scc_annotation']}"
+    )
+
+
+@pytest.mark.polarion("CNV-11965")
+def test_deployments_with_incorrect_required_scc(required_scc_deployment_check):
+    assert not required_scc_deployment_check["incorrect_required_scc_annotation_value"], (
+        f"Deployments incorrect {REQUIRED_SCC_ANNOTATION} annotation: "
+        f"{required_scc_deployment_check['incorrect_required_scc_annotation_value']}"
+    )
```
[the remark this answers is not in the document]
dshchedr: so it is a possible that we may have annotation `openshift.io/required-scc` with some another value (other than `restricted-v2`) - and the first tests should pass?

if we always need to have `restricted-v2` on a cluster then I still don't understand the point of having two separate tests. 
if `restricted-v2` is optional - then ok, two tests may sound reasonable (but I guess they should not be run together)

tests/install_upgrade_operators/security/scc/test_cnv_deployment_required_scc.py
```diff
@@ -0,0 +1,56 @@
...
+
+        scc = dp.instance.spec.template.metadata.annotations.get(REQUIRED_SCC_ANNOTATION)
+
+        if scc != REQUIRED_SCC_VALUE:
+            if scc is None:
+                missing_required_scc_annotation.append(dp.name)
+            incorrect_required_scc_annotation_value.append(dp.name)
+
+    return {
+        "missing_required_scc_annotation": sorted(missing_required_scc_annotation),
```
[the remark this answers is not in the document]
rnetser: `sorted` retunr a new list, so if this action is not needed (and it is not in this case), please remove. when someone looks at the code, they'd assume that this is needed.

rnetser (changes requested):

tests/install_upgrade_operators/security/scc/test_cnv_deployment_required_scc.py
```diff
@@ -0,0 +1,56 @@
...
+    return [Deployment(name=dep_name, namespace=hco_namespace.name) for dep_name in ALL_CNV_DEPLOYMENTS_NO_HPP_POOL]
+
+
+@pytest.fixture(scope="module")
+def required_scc_deployment_check(all_hco_deployments):
+    missing_required_scc_annotation = []
+    incorrect_required_scc_annotation_value = []
+
+    for dp in all_hco_deployments:
+        if is_jira_open(jira_id="CNV-62807") and dp.name in VIRT_EXPORTPROXY:
```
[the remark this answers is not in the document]
geetikakay: @rnetser can i resolve this, we will have fix soon

tests/install_upgrade_operators/security/scc/test_cnv_deployment_required_scc.py
```diff
@@ -0,0 +1,54 @@
...
+    ):
+        if is_jira_open(jira_id="CNV-62807") and dp.name in VIRT_EXPORTPROXY:
+            continue
+
+        scc = dp.instance.spec.template.metadata.annotations.get(REQUIRED_SCC_ANNOTATION)
+
+        if scc is None:
+            missing_required_scc_annotation.append(dp.name)
+        elif scc != REQUIRED_SCC_VALUE:
+            incorrect_required_scc_annotation_value[dp.name] = scc
```
hmeir, line 33: Tnx for this PR!
So according to https://issues.redhat.com/browse/CNV-52649, we expect the deployments podTemplate to have `restricted-v2` annotation.
Is it always the case?

For example if we set on the csv different value, it may be different.
Shouldn't we decide about the value dynamically?

And what about the pods?
For example, `bridge-marke`r pods are getting the value from their deamonset.
Should we check that as well?

geetikakay: @hmeir if response looks satisfactory can i resolve this conversation.

tests/install_upgrade_operators/security/scc/test_cnv_deployment_required_scc.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@pytest.fixture(scope="module")
+def required_scc_deployment_check(admin_client, hco_namespace):
+    missing_required_scc_annotation = []
+    incorrect_required_scc_annotation_value = {}
+
+    for dp in (
+        Deployment(client=admin_client, name=name, namespace=hco_namespace.name)
+        for name in ALL_CNV_DEPLOYMENTS_NO_HPP_POOL
```
hmeir, line 22: Why skip the HPP one?
Shouldn't we verify its with expected value?

geetikakay: HPP doesn't use restricted-v2 and is not enforced to use it

dshchedr (approved):

openshift-virtualization-qe-bot-2:

PR #1126, opened 2025-06-03
Update hyperv features test to use instance types

RoniKishner:
##### Short description:
`hyperv_vm` will now use preferences instead of templates

##### More details:
The tests in `test_hyperv_features_in_vm.py` currently use templates (windows-2019 and fedora) to create VMs. This PR updates them to use common-instance types instead, as that is the recommended approach.

Change tests from using templates to common-instance types

##### jira-ticket:

https://issues.redhat.com/browse/CNV-62919

## Summary by CodeRabbit

- **Tests**
  - Refactored Hyper-V feature support tests to enhance VM creation and configuration using new resource-based fixtures.
  - Improved test parameterization for more flexible and maintainable validation of Hyper-V features.

### DIFF
--- a/tests/virt/node/hyperv_support/test_hyperv_features_in_vm.py
+++ b/tests/virt/node/hyperv_support/test_hyperv_features_in_vm.py
@@ -1,18 +1,19 @@
 import logging
 
 import pytest
+from ocp_resources.virtual_machine_cluster_instancetype import VirtualMachineClusterInstancetype
+from ocp_resources.virtual_machine_cluster_preference import VirtualMachineClusterPreference
 from pytest_testconfig import py_config
 
 from tests.os_params import (
     FEDORA_LATEST,
-    FEDORA_LATEST_LABELS,
     FEDORA_LATEST_OS,
     WINDOWS_LATEST,
-    WINDOWS_LATEST_LABELS,
     WINDOWS_LATEST_OS,
 )
-from utilities.constants import VIRT_LAUNCHER
-from utilities.virt import vm_instance_from_template
+from utilities.constants import OS_FLAVOR_FEDORA, OS_FLAVOR_WINDOWS, VIRT_LAUNCHER
+from utilities.storage import data_volume_template_with_source_ref_dict
+from utilities.virt import VirtualMachineForTests, running_vm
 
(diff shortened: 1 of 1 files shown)


rnetser:
@RoniKishner @kbidarkar I think that this PR should be marked as `hold` until the teams decide on its contents

tests/virt/node/hyperv_support/test_hyperv_features_in_vm.py
```diff
@@ -23,19 +24,47 @@ def hyperv_vm(
...
-        namespace=namespace,
-        data_source=golden_image_data_source_scope_class,
+    with VirtualMachineForTests(
+        name=request.param.get("vm_name"),
+        namespace=namespace.name,
+        client=unprivileged_client,
+        data_volume_template=data_volume_template_with_source_ref_dict(
+            data_source=golden_image_data_source_scope_class
+        ),
+        disk_type=None,
```
geetikakay, line 37: do we need disk_type?

dshchedr:
> @RoniKishner Infra team should work only on infra tests, they should not update other component teams tests. This is the main reason why the tests are split as per the component teams.
> 
> I discussed this with both Dan and Dominik during the Ready-Ready meeting and it was mentioned and agreed that moving to instancetypes and preferences Epic is all about only introducing the fixtures and functions ( supporting stuff for the tests for the infra team itself. ) The Epic name was also updated due to this.
> 
> @dshchedr @rnetser cc

Until the Templates officially deprecated - we have to keep the coverage. So, we can't simply switch to using IT/Perf

If Infra team is interested in covering same scenarios but with it/perf - they can clone existing tests into their component folder and update tests according to needs.

dshchedr:
> What aspect of coverage is this test addressing? It appears to be focused on Hyper-V rather than templates, as it uses a specific Windows.server.2019 template. Additionally, there's already a dedicated Hyper-V test under the `os_support` folder. If this test does, in fact, cover something template-related, should we consider moving it under `common_templates` for better organization?

This test is about Hyper-V features, not about templates.  But it was working with templates for a long time and we didn't have any issues with that. Templates are still supported, so from test point of view it still ok.

PR #1203, opened 2025-06-12
[cnv-4.18] Mark network tests that don't require node's additional interfaces

yossisegev:
Acceptance tests for new architectures, platforms etc. shouldn't depend on having extra NICs on the cluster nodes, except for the primary interface. In order to have satisfactory coveage in these tests, we mark all the these tests with the a new marker, which can and should be used for these types of tests.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -32,6 +32,7 @@ markers =
     hugepage: tests that require nodes with hugepages
     service_mesh: tests that require the service mesh operator to be installed
     jumbo_frame: tests that require network configurations supporting jumbo frames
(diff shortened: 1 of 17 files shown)


openshift-virtualization-qe-bot-2:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-1203 published

yossisegev:
Verified by running https://jenkins-csb-cnvqe-main.dno.corp.redhat.com/job/openshift-virtualization-tests-runner/2453/consoleFull (--collect-only)

PR #1242, opened 2025-06-19
Update Cirros image to Fedora into test_clone.py

josemacassan:
##### Short description:
Change image use from Cirros to Fedora into test `tests/storage/cdi_clone/test_clone.py`
It makes use of Fedora image in order to create VMs into test_clone.py tests as well as all the changes related to it. 
## Summary by CodeRabbit

- **Tests**
	- Added new test fixtures for Fedora-based DataVolumes with both filesystem and block volume modes.
	- Updated test cases to use Fedora images, data volumes, and memory sizes instead of Cirros.
	- Introduced a utility for creating Fedora DataVolumes in test setups.

### DIFF
--- a/tests/storage/cdi_clone/conftest.py
+++ b/tests/storage/cdi_clone/conftest.py
@@ -1,34 +1,8 @@
 import pytest
 from ocp_resources.datavolume import DataVolume
 
-from tests.storage.utils import create_cirros_dv, create_fedora_dv
+from tests.storage.utils import create_fedora_dv
 from utilities.storage import data_volume
-
-
-@pytest.fixture(scope="module")
-def cirros_dv_with_filesystem_volume_mode(
-    namespace,
-    storage_class_with_filesystem_volume_mode,
-):
-    yield from create_cirros_dv(
-        namespace=namespace.name,
-        name="cirros-fs",
-        storage_class=storage_class_with_filesystem_volume_mode,
-        volume_mode=DataVolume.VolumeMode.FILE,
-    )
-
-
-@pytest.fixture(scope="module")
-def cirros_dv_with_block_volume_mode(
-    namespace,
-    storage_class_with_block_volume_mode,
-):
-    yield from create_cirros_dv(
-        namespace=namespace.name,
-        name="cirros-block",
-        storage_class=storage_class_with_block_volume_mode,
-        volume_mode=DataVolume.VolumeMode.BLOCK,
-    )
 
 
 @pytest.fixture(scope="module")
--- a/tests/storage/cdi_clone/conftest.py
+++ b/tests/storage/cdi_clone/conftest.py
@@ -1,7 +1,7 @@
 import pytest
 from ocp_resources.datavolume import DataVolume
 
-from tests.storage.utils import create_cirros_dv
+from tests.storage.utils import create_cirros_dv, create_fedora_dv
 from utilities.storage import data_volume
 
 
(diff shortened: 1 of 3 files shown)


tests/storage/cdi_clone/test_clone.py
```diff
@@ -114,8 +114,8 @@ def test_successful_clone_of_large_image(
         pytest.param(
             {
                 "dv_name": "dv-source",
-                "image": f"{Images.Cirros.DIR}/{Images.Cirros.QCOW2_IMG}",
-                "dv_size": Images.Cirros.DEFAULT_DV_SIZE,
+                "image": f"{Images.Fedora.DIR}/{Images.Fedora.FEDORA41_IMG}",
```
jpeimer, line 117: It would be even better to use `FEDORA_LATEST["image_path"]` like you did for RHEL images, so we wouldn't need to update the tests when a new OS version comes up.

Please use it here and in all other places in this PR

tests/storage/cdi_clone/test_clone.py
```diff
@@ -51,9 +51,9 @@ def create_vm_from_clone_dv_template(
     with VirtualMachineForTests(
         name=vm_name,
         namespace=namespace_name,
-        os_flavor=OS_FLAVOR_CIRROS,
+        os_flavor=OS_FLAVOR_FEDORA,
         client=client,
-        memory_requests=Images.Cirros.DEFAULT_MEMORY_SIZE,
+        memory_requests=Images.Fedora.DEFAULT_MEMORY_SIZE,
```
jpeimer, line 56: `memory_requests` field is deprecated, please use `memory_guest` instead

```suggestion
        memory_guest=Images.Fedora.DEFAULT_MEMORY_SIZE,
```

tests/storage/cdi_clone/conftest.py
```diff
@@ -31,6 +31,32 @@ def cirros_dv_with_block_volume_mode(
     )
```
jpeimer, line 32: tox is failing in CI, most likely because `cirros_dv_with_filesystem_volume_mode` and `cirros_dv_with_block_volume_mode` fixtures are not used anymore, which means you can delete them

jpeimer (changes requested):
Thanks for the PR! We've been wanting to replace these images for a long time

tests/storage/utils.py
```diff
@@ -414,6 +414,29 @@ def create_cirros_dv(
...
+    storage_class,
+    access_modes=None,
+    volume_mode=None,
+    client=None,
+    dv_size=Images.Fedora.DEFAULT_DV_SIZE,
+):
+    with create_dv(
+        dv_name=f"dv-{name}",
+        namespace=namespace,
+        url=get_http_image_url(image_directory=Images.Fedora.DIR, image_name=Images.Fedora.FEDORA41_IMG),
```
jpeimer, line 429: Can also use latest here 
```suggestion
        url=f"{get_test_artifact_server_url()}{FEDORA_LATEST['image_path']}",
```

jpeimer (changes requested):

tests/storage/utils.py
josemacassan: Done! Thanks for the suggestion

tests/storage/cdi_clone/test_clone.py
```diff
@@ -114,8 +114,8 @@ def test_successful_clone_of_large_image(
         pytest.param(
             {
                 "dv_name": "dv-source",
-                "image": f"{Images.Cirros.DIR}/{Images.Cirros.QCOW2_IMG}",
-                "dv_size": Images.Cirros.DEFAULT_DV_SIZE,
+                "image": FEDORA_LATEST["image_path"],
```
rnetser, line 117:
```suggestion
                "image": FEDORA_LATEST.get("image_path"),
```
In cases where we do not have a fedora image ready for a new arch
this is needed for the collection part (the test will not be executed if the image is missing as there will be no arch marker on it)

Applies to all places where accessing such keys in pytest.param

rnetser (changes requested):

openshift-virtualization-qe-bot-2:

PR #1247, opened 2025-06-19
net tests, hotplug: use unprivileged client for migration

servolkov:
##### Short description:
Don't use cluster admin for migration in hot-plug tests. Use namespace admin + granted role instead. 

Starting with release 4.19, VM migration can be performed by either a cluster admin or a namespace admin with an assigned role - kubevirt.io:migrate. In hot-plug tests, VM migration is performed by the admin cluster, which is not entirely correct from the point of view of the user scenario. It is more correct to grant the necessary rights to the namespace admin and migrate from the namespace admin.

Since 4.20 live migration will be performed automatically for hot-plug, and there will be no need to trigger VM migration manually at all. Nevertheless, current PR makes sense for 4.19.

## Summary by CodeRabbit

- **Tests**
	- Enhanced tests to support running VM migration operations as an unprivileged user by granting necessary permissions.
	- Updated test fixtures and functions to use the unprivileged client context for migration-related actions.

### DIFF
--- a/tests/network/l2_bridge/test_bridge_nic_hot_plug.py
+++ b/tests/network/l2_bridge/test_bridge_nic_hot_plug.py
@@ -1,6 +1,8 @@
 import time
 
 import pytest
+from ocp_resources.cluster_role import ClusterRole
+from ocp_resources.role_binding import RoleBinding
 
 from libs.net import netattachdef
 from tests.network.constants import IPV4_ADDRESS_SUBNET_PREFIX
@@ -19,7 +21,7 @@ from tests.network.l2_bridge.utils import (
     set_secondary_static_ip_address,
     wait_for_interface_hot_plug_completion,
 )
-from utilities.constants import FLAT_OVERLAY_STR, SRIOV
+from utilities.constants import FLAT_OVERLAY_STR, SRIOV, UNPRIVILEGED_USER
 from utilities.network import (
     IfaceNotFound,
     assert_ping_successful,
@@ -28,11 +30,30 @@ from utilities.network import (
 )
 from utilities.virt import migrate_vm_and_verify
 
-pytestmark = pytest.mark.usefixtures("label_schedulable_nodes")
+pytestmark = pytest.mark.usefixtures(
+    "label_schedulable_nodes",
+    "unprivileged_client_migration_permission_role_binding",
+)
 
 HOT_PLUG_STR = "hot-plug"
 TEST_BASIC_HOT_PLUGGED_INTERFACE_CONNECTIVITY = "test_basic_connectivity_of_hot_plugged_interface"
 SECONDARY_SETUP_INTERFACE_NAME = "eth1"
+
+
+@pytest.fixture(scope="module")
+def unprivileged_client_migration_permission_role_binding(admin_client, namespace):
+    migration_role = ClusterRole(name="kubevirt.io:migrate", client=admin_client, ensure_exists=True)
+
+    with RoleBinding(
+        name="migrate-role-binding",
+        namespace=namespace.name,
+        client=admin_client,
+        subjects_kind="User",
+        subjects_name=UNPRIVILEGED_USER,
+        role_ref_kind=migration_role.kind,
+        role_ref_name=migration_role.name,
+    ):
+        yield
 
 
 @pytest.fixture(scope="class")
@@ -148,6 +169,7 @@ def running_vm_with_secondary_and_hot_plugged_interfaces(
 def hot_plugged_interface_name_on_vm_created_with_secondary_interface(
     running_vm_with_secondary_and_hot_plugged_interfaces,
     network_attachment_definition_for_hot_plug,
+    unprivileged_client,
 ):
     hot_plugged_interface_name = f"{HOT_PLUG_STR}-additional-iface"
     hot_plug_interface(
@@ -157,7 +179,7 @@ def hot_plugged_interface_name_on_vm_created_with_secondary_interface(
     )
     # In order to complete the interface hot-plug, and have the interface available in the guest VM -
     # the VM must be migrated.
-    migrate_vm_and_verify(vm=running_vm_with_secondary_and_hot_plugged_interfaces)
+    migrate_vm_and_verify(vm=running_vm_with_secondary_and_hot_plugged_interfaces, client=unprivileged_client)
 
     return hot_plugged_interface_name
 
@@ -176,10 +198,10 @@ def hot_plugged_second_interface_with_address(
 
 
 @pytest.fixture()
-def migrated_vm_with_hot_plugged_interface_attached(running_vm_for_nic_hot_plug):
+def migrated_vm_with_hot_plugged_interface_attached(running_vm_for_nic_hot_plug, unprivileged_client):
     # In order to complete the interface hot-plug, and have the interface available in the guest VM -
     # the VM must be migrated.
-    migrate_vm_and_verify(vm=running_vm_for_nic_hot_plug)
+    migrate_vm_and_verify(vm=running_vm_for_nic_hot_plug, client=unprivileged_client)
     return running_vm_for_nic_hot_plug
 
 
@@ -353,12 +375,13 @@ def hot_unplugged_additional_interface(
     namespace,
     running_vm_with_secondary_and_hot_plugged_interfaces,
     hot_plugged_interface_name_on_vm_created_with_secondary_interface,
+    unprivileged_client,
 ):
     hot_unplug_interface(
         vm=running_vm_with_secondary_and_hot_plugged_interfaces,
         hot_plugged_interface_name=hot_plugged_interface_name_on_vm_created_with_secondary_interface,
     )
-    migrate_vm_and_verify(vm=running_vm_with_secondary_and_hot_plugged_interfaces)
+    migrate_vm_and_verify(vm=running_vm_with_secondary_and_hot_plugged_interfaces, client=unprivileged_client)
     return hot_plugged_interface_name_on_vm_created_with_secondary_interface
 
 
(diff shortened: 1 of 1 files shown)


openshift-virtualization-qe-bot-2:

## Welcome! 🎉

This pull request will be automatically processed with the following features:

### 🔄 Automatic Actions
* **Reviewer Assignment**: Reviewers are automatically assigned based on the OWNERS file in the repository root
* **Size Labeling**: PR size labels (XS, S, M, L, XL, XXL) are automatically applied based on changes
* **Issue Creation**: A tracking issue is created for this PR and will be closed when the PR is merged or closed
* **Pre-commit Checks**: [pre-commit](https://pre-commit.ci/) runs automatically if `.pre-commit-config.yaml` exists
* **Branch Labeling**: Branch-specific labels are applied to track the target branch
* **Auto-verification**: Auto-verified users have their PRs automatically marked as verified

### 📋 Available Commands

#### PR Status Management
* `/wip` - Mark PR as work in progress (adds WIP: prefix to title)
* `/wip cancel` - Remove work in progress status
* `/hold` - Block PR merging (approvers only)
* `/hold cancel` - Unblock PR merging
* `/verified` - Mark PR as verified
* `/verified cancel` - Remove verification status

#### Review & Approval
* `/lgtm` - Approve changes (looks good to me)
* `/approve` - Approve PR (approvers only)
* `/assign-reviewers` - Assign reviewers based on OWNERS file
* `/assign-reviewer @username` - Assign specific reviewer
* `/check-can-merge` - Check if PR meets merge requirements

#### Testing & Validation
 * `/retest tox` - Run Python test suite with tox
 * `/retest build-container` - Rebuild and test container image
 * `/retest all` - Run all available tests


#### Container Operations
* `/build-and-push-container` - Build and push container image (tagged with PR number)
  * Supports additional build arguments: `/build-and-push-container --build-arg KEY=value`

#### Cherry-pick Operations
* `/cherry-pick <branch>` - Schedule cherry-pick to target branch when PR is merged
  * Multiple branches: `/cherry-pick branch1 branch2 branch3`

#### Label Management
* `/<label-name>` - Add a label to the PR
* `/<label-name> cancel` - Remove a label from the PR

### ✅ Merge Requirements

This PR will be automatically approved when the following conditions are met:

1. **Approval**: `/approve` from at least one approver
2. **LGTM Count**: Minimum 2 `/lgtm` from reviewers
3. **Status Checks**: All required status checks must pass
4. **No Blockers**: No WIP, hold, or conflict labels
5. **Verified**: PR must be marked as verified (if verification is enabled)

### 📊 Review Process

<details>
<summary><strong>Approvers and Reviewers</strong></summary>


**Approvers:**
 * @EdDev


**Reviewers:**
 * @Anatw
 * @EdDev
 * @servolkov
 * @yossisegev


</details>

<details>
<summary><strong>Available Labels</strong></summary>

 * hold
 * verified
 * wip
 * lgtm
 * approve

</details>

### 💡 Tips

* **WIP Status**: Use `/wip` when your PR is not ready for review
* **Verification**: The verified label is automatically removed on each new commit
* **Cherry-picking**: Cherry-pick labels are processed when the PR is merged
* **Container Builds**: Container images are automatically tagged with the PR number
* **Permission Levels**: Some commands require approver permissions
* **Auto-verified Users**: Certain users have automatic verification and merge privileges

For more information, please refer to the project documentation or contact the maintainers.

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
```diff
@@ -28,13 +30,34 @@
...
+        name="role-bind-kubevirt-migrate",
+        namespace=namespace.name,
+        client=admin_client,
+        subjects_kind="User",
+        subjects_name=UNPRIVILEGED_USER,
+        subjects_namespace=namespace.name,
+        role_ref_kind=migration_role.kind,
+        role_ref_name=migration_role.name,
+    ):
+        yield
```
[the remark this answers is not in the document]
servolkov: Indeed. Fixed. Thank you, AI!

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
```diff
@@ -28,13 +30,33 @@
...
+    "grant_unprivileged_client_permission_for_migration",
+)
 
 HOT_PLUG_STR = "hot-plug"
 TEST_BASIC_HOT_PLUGGED_INTERFACE_CONNECTIVITY = "test_basic_connectivity_of_hot_plugged_interface"
 SECONDARY_SETUP_INTERFACE_NAME = "eth1"
 
 
+@pytest.fixture(scope="module")
+def grant_unprivileged_client_permission_for_migration(admin_client, namespace):
```
rnetser, line 44: please rename to match what is returned by the fixture, e.g `unprivileged_client_migration_ permission_role_binding`

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
```diff
@@ -28,13 +30,33 @@
...
+    """Grant unprivileged_client permission for VM migration"""
```
rnetser, line 45: once the fixture name is meaningful, this is not needed.

rnetser (changes requested):

yossisegev (changes requested):

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
```diff
@@ -148,6 +170,7 @@ def running_vm_with_secondary_and_hot_plugged_interfaces(
 def hot_plugged_interface_name_on_vm_created_with_secondary_interface(
     running_vm_with_secondary_and_hot_plugged_interfaces,
     network_attachment_definition_for_hot_plug,
+    unprivileged_client,
```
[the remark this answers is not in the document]
rnetser: @yossisegev if `grant_unprivileged_client_permission_for_migration` it will be hard to understand what fails if the name is not indicative or if someone would like to reuse the fixture so something like `unprivileged_client_migration_role_binding' would explain what is the fixture about

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
```diff
@@ -28,13 +30,32 @@
 )
 from utilities.virt import migrate_vm_and_verify
 
-pytestmark = pytest.mark.usefixtures("label_schedulable_nodes")
+pytestmark = pytest.mark.usefixtures(
+    "label_schedulable_nodes",
+    "unprivileged_client_migration_permission_role_binding",
```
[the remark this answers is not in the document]
rnetser: They should be moved, but should be in a separate PR.
In general, we avoid using `autouse` as it makes it harder to follow the flow.

rnetser (commented):
@servolkov is there a reason why this PR is on `cnv-4.19` and not on `main`?

servolkov:
> > /verified
> 
> @servolkov , please detail exactly what you executed and what was the result so we get some more details about what was verified.

done

servolkov:
> * In the commit message, please wrap the body to 72 chars, it is one of the best practices suggestion: https://cbea.ms/git-commit/#wrap-72
>   It helps reading it on the terminal.

@EdDev done

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
yossisegev: @servolkov You raised good points here, so I will try to clarify:
> this fixture doesn’t return anything (consciously). Its only purpose is to modify an existing object (the client) by granting it permissions needed for the test.

Even if the fixture doesn't explicitly return/yield an object, there is still something that was produced out of it - an object, or a state, that were created or changed as a result of calling this fixture.
In this specific case - the fixture produced a RoleBinding resource, so the new name you gave (`unprivileged_client_migration_permission_role_binding`) is good. Note that we could have stated that the output of this fixture is a project-admin with migration privileges, so you could also rename it to something like `migration_permitted_unprivileged_client` (and just to clarify - this is not to say that you should rename the fixture again, but rather to point what can be another valid option).

> I noticed that in other similar fixtures we often return the binding object but don’t use it — which seems like extra overhead that’s not needed.

If a fixture is called, then it is required. There are fixtures that are not meant to be referenced anywhere in the code, but they are required because they do something that is required for the flow, for example - a fixture that drains a node.

yossisegev (approved):

EdDev (approved):
Thank you.

We have a disagreement on the use of `autouse` [1] and in this case I think it is easy to see that there is no point in defining the fixture and then adding another ref to it at the same level. The arguments provided are not convincing me not to use it, however, I won't fight over this through this PR.
 
 I expect we will continue this debate offline as I like `autouse`. :)
 
 /approve
[1] https://github.com/RedHatQE/openshift-virtualization-tests/pull/1247#discussion_r2166222970

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
servolkov: Thanks @yossisegev.

>>I noticed that in other similar fixtures we often return the binding object but don’t use it — which seems like extra overhead that’s not needed.

>If a fixture is called, then it is required. There are fixtures that are not meant to be referenced anywhere in the code, but they are required because they do something that is required for the flow, for example - a fixture that drains a node.

My point was not that the fixture is unnecessary or unused. It's necessary as part of the test flow. What I meant: we sometimes return an object (like the binding) from the fixture, even though we never use that returned object in the test code.

Compare:
1) ```
    def some_fixture():
        result = some_action()
        return result
    ```

2)
      ```
      def some_fixture():
          some_action()
      ```
If we never use `result` later, returning it just keeps it in memory longer than needed - potentially holding state or resources that we’re not making use of. In such cases, not returning is a small optimization and a way to keep things more intentional. 

Hope that clears it up.

yossisegev: That's 100%.
If you need your fixture to just do stuff without returning any object - don't return any object.

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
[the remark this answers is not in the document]
yossisegev: I agree that using autouse is implicit and thus harder to follow, but I don't think that's a reason not to use it. It should be used per the need. If a fixture is a must before a module is exemight becuted, and it is needed all across the module, then autouse is probably the most accurate option.
And regarding the difficulty to follow it - I experienced the same difficulty when I was first introduced to autouse and usefixture. It took several times until I got myself used to search for these decorators (or whatever their official title is) when I see some setup which I am not sure what initiated it.
And one more thing about it - there are always cases where one would prefer one way to implement, while the other would prefer a different way - and both ways are accurate. Let's not confuse correct coding with personal preference. We have already been there with cnv-tests. let's not go back to it, and give ourselves the "artistic freedom" to write the code as each of use sees fit (as long as it is correct and doesn't contradict any sensible coding guidelines).

PR #1080, opened 2025-05-27
Remove irrelevant host VLAN tests

yossisegev:
These tests cover VLAN configurations - and their usages - in cluster nodes, without testing the usage of these configurations by virtual machines, and therefore - they don't belong in Openhsitf Virtualization tests.

## Summary by CodeRabbit

- **Tests**
  - Removed all automated tests and utilities related to VLAN network interfaces, DHCP server/client setups, and VLAN bonding in host network environments.
  - Test cases for VLAN interface presence, dynamic IP assignment, VLAN connectivity, and interface deletion have been deleted.
  - All supporting test fixtures and utility functions for managing VLAN test environments have been removed.

PR #1258, opened 2025-06-22
[Cherry-pick cnv-4.18] Remove irrelevant host VLAN tests (#1080)

yossisegev:
These tests cover VLAN configurations - and their usages - in cluster nodes, without testing the usage of these configurations by virtual machines, and therefore - they don't belong in Openhsitf Virtualization tests.


yossisegev (commented):
/lgtm

PR #1309, opened 2025-06-26
[Chaos] Refactor chaos_worker_background_process

qwang1:
##### Short description:
1. Replaced multiprocessing with threading to avoid pickling issues on macOS and Windows.
2. Increased the utility pod’s resource limits to prevent pod restarts or stress processes being killed by the OOM killer during load injection.
3. Improved the stress-ng command by using stronger stressors and redirecting output to a log file for easier analysis.

## Summary by CodeRabbit

* **New Features**
  * Enhanced chaos testing capabilities with new background thread fixtures and utilities for running and monitoring commands inside utility pods.
  * Improved process detection and asynchronous command execution within pods for more robust test scenarios.

* **Bug Fixes**
  * Updated resource limits for utility daemonset containers to prevent resource exhaustion during intensive operations.

* **Chores**
  * Changed utility container image references to a new repository path.
  * Added installation of an additional system package to the utility container for improved process management.

### DIFF
--- a/tests/chaos/standard/test_standard.py
+++ b/tests/chaos/standard/test_standard.py
@@ -3,6 +3,7 @@ from ocp_resources.deployment import Deployment
 from ocp_resources.virtual_machine import VirtualMachine
 
 from tests.chaos.constants import STRESS_NG
+from tests.chaos.utils import wait_for_process_to_start
 from utilities.constants import (
     TIMEOUT_2MIN,
     TIMEOUT_5MIN,
@@ -118,13 +119,14 @@ def test_odf_storage_outage(
 
 @pytest.mark.gpfs
 @pytest.mark.parametrize(
-    "chaos_worker_background_process",
+    "chaos_worker_background_thread",
     [
         pytest.param(
             {
-                "max_duration": TIMEOUT_2MIN,
-                "background_command": f"{STRESS_NG}  --io 5 -t 120s",
-                "process_name": STRESS_NG,
+                "timeout": TIMEOUT_2MIN,
+                "background_command": f"{STRESS_NG}  --iomix 5 -t 120s --metrics-brief 2>&1 | tee /tmp/stress-ng.log",
+                "thread_name": STRESS_NG,
+                "retries": 2,
             },
         ),
     ],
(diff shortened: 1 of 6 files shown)


dshchedr (changes requested):

dshchedr:
> 2\. Increased the utility pod’s resource limits to prevent pod restarts or stress processes being killed by the OOM killer during load injection.

I just want to highlight that you should not increase utility pods limits because it will affect the whole project and all possible test runs. Most of tests executed on PSI clusters which has very limited resources, so increasing utility limits is not an option. 

I think, the option for you can be - create a separate pod and run stress there, in this case you can set any needed amount of resources for this particular pod (it can be even calculated based on the node size). Also in this case you can use node selector to run the pod on the same node where VM is running

tests/chaos/standard/test_standard.py
```diff
@@ -135,22 +137,40 @@ def test_host_io_stress(
...
+
+    thread, result_queue = chaos_worker_background_thread
+    target_node = vm_node_with_chaos_label[0]
+
+    stress_pid = wait_for_process_to_start(
+        probe_func=stress_ng_probe,
+        node=target_node,
+        timeout=120,
+        sleep=5,
+    )
```
vsibirsk, line 156: this should be part of setup

tests/chaos/utils.py
```diff
@@ -415,3 +417,126 @@ def get_instance_type(name):
...
+            command=f"pgrep -f {process_name}",
+            chroot_host=False,
+            timeout=30,
+        )
+        pids = output.strip().split("\n")
+        if pids and pids[0].strip():
+            LOGGER.info(f"[process probe] Found '{process_name}' with PID(s): {pids}")
+            return pids[0]
+    except Exception as e:
+        LOGGER.debug(f"[process probe] Error probing '{process_name}': {e}")
```
vsibirsk, line 517:
```suggestion
        LOGGER.warning(f"[process probe] Error probing '{process_name}': {e}")
```

tests/chaos/utils.py
```diff
@@ -415,3 +417,126 @@ def get_instance_type(name):
...
+
+    Raises:
+        AssertionError: If process is not detected within timeout.
+    """
+    for pid in TimeoutSampler(wait_timeout=timeout, sleep=sleep, func=lambda: probe_func(node)):
+        if pid:
+            LOGGER.info(f"[process wait] Process started with PID: {pid}")
+            return pid
+
+    raise AssertionError(f"[process wait] Process not running on {node.name} after {timeout}s")
```
vsibirsk, line 542: should be `ValueError`

vsibirsk (changes requested):

PR #1333, opened 2025-06-30
Add test for link state in localnet VM

frenzyfriday:
##### Short description:
Adds a test that : 
- creates and run a VM with a multus interface that is connected via localnet in the node and the interface set to  `down`.
- After the VM comes up - edit the VMI  and set the interface state to `up`.
- Verify the secondary interface appears as `UP` in the guest (`ip link show dev eth0`).
- Verify it obtains connectivity.

##### What this PR does / why we need it:
From KubeVirt v1.5.0 we can specify the state of an interface when creating a VM. This PR adds one of the test scenarios for this feature.

##### Special notes for reviewer:
I am marking the PR as draft, but could you pls leave me notes on how it can be improved?

##### jira-ticket:
https://issues.redhat.com/browse/CNV-64146

PR #1346, opened 2025-07-02
net,hotplug: adjust tests to VMRolloutStrategy

servolkov:
##### Short description: 

Starting from 4.20 the migration logic is an encapsulated detail and there is no need to follow it for hot-plug\unplug. The only thing to monitor to meet success of hot-plug\unplug process is waiting for the interface is presented (or not presented if hot-unplug) in VMI status.

There  was also refactored `lookup_iface_status` to accept an optional `wait_timeout`
parameter with a default value.  `LOOKUP_IFACE_STATUS_TIMEOUT_SEC` is not suitable for the hot-plug issue. The change improves flexibility for callers who need to override the default timeout without changing internal
logic or duplicating code.

  **Depends on:**  https://github.com/RedHatQE/openshift-virtualization-tests/pull/1333

## Summary by CodeRabbit

* **Tests**
  * Made VM migration steps explicit within tests rather than implicit in fixtures.
  * Improved test stability by explicitly waiting for network interface status changes after hot-plug and hot-unplug operations.
* **New Features**
  * Added functionality to wait for network interface removal within a timeout period.
* **Bug Fixes**
  * Replaced implicit VM migration verification with explicit interface status polling to enhance reliability.

### DIFF
--- a/libs/net/vmspec.py
+++ b/libs/net/vmspec.py
@@ -1,16 +1,14 @@
-import logging
 from collections.abc import Callable
 from typing import Any, Final
 
 from kubernetes.dynamic.client import ResourceField
-from timeout_sampler import TimeoutExpiredError, TimeoutSampler
+from timeout_sampler import TimeoutExpiredError, TimeoutSampler, retry
 
 from libs.vm.spec import Devices, Interface, Network, SpecDisk, VMISpec, Volume
 from libs.vm.vm import BaseVirtualMachine
 
-LOGGER = logging.getLogger(__name__)
-
 LOOKUP_IFACE_STATUS_TIMEOUT_SEC: Final[int] = 30
+WAIT_FOR_MISSING_IFACE_STATUS_TIMEOUT_SEC: Final[int] = 120
 RETRY_INTERVAL_SEC: Final[int] = 5
 IP_ADDRESS: Final[str] = "ipAddress"
 
@@ -35,7 +33,7 @@ def lookup_iface_status(
     vm: BaseVirtualMachine,
     iface_name: str,
     predicate: Callable[[Any], bool] = _default_interface_predicate,
-    wait_timeout: int = LOOKUP_IFACE_STATUS_TIMEOUT_SEC,
+    timeout: int = LOOKUP_IFACE_STATUS_TIMEOUT_SEC,
 ) -> ResourceField:
     """
     Awaits and returns the network interface status requested if found and the predicate function,
@@ -47,7 +45,7 @@ def lookup_iface_status(
         predicate (Callable[[dict[str, Any]], bool]): A function that takes a network interface as an argument
             and returns a boolean value. this function should define the condition that
             the interface needs to meet.
-        wait_timeout (int): lookup operation timeout
+        timeout (int): lookup operation timeout
 
     Returns:
         iface (ResourceField): The requested interface.
@@ -56,7 +54,7 @@ def lookup_iface_status(
         VMInterfaceStatusNotFoundError: If the requested interface was not found in the vmi status.
     """
     sampler = TimeoutSampler(
-        wait_timeout=wait_timeout,
+        wait_timeout=timeout,
         sleep=RETRY_INTERVAL_SEC,
         func=_lookup_iface_status,
         vm=vm,
@@ -92,6 +90,31 @@ def _lookup_iface_status(
         if iface.name == iface_name and predicate(iface):
             return iface
     return None
+
+
+@retry(
+    wait_timeout=WAIT_FOR_MISSING_IFACE_STATUS_TIMEOUT_SEC,
+    sleep=RETRY_INTERVAL_SEC,
+    exceptions_dict={VMInterfaceStatusStillExistsError: []},
+)
+def wait_for_missing_iface_status(vm: BaseVirtualMachine, iface_name: str) -> bool:
+    """
+        Waits for a network interface to be deleted from the virtual machine's interface status.
+
+        Args:
+            vm (BaseVirtualMachine): The virtual machine to check for the interface status.
+            iface_name: (str): The name of the network interface to wait for deletion.
+
+        Raises:
+            VMInterfaceStatusStillExistsError: If the interface still exists after the timeout period.
+    there was a usage of it
+        Returns:
+            bool: True if the interface is missing otherwise raises an exception.
+    """
+    if _lookup_iface_status(vm=vm, iface_name=iface_name, predicate=lambda _: True) is not None:
+        raise VMInterfaceStatusStillExistsError(f"Interface {iface_name} still exists in {vm.name}")
+
+    return True
 
 
 def lookup_primary_network(vm: BaseVirtualMachine) -> Network:
--- a/libs/net/vmspec.py
+++ b/libs/net/vmspec.py
@@ -1,11 +1,14 @@
+import logging
 from collections.abc import Callable
 from typing import Any, Final
 
 from kubernetes.dynamic.client import ResourceField
-from timeout_sampler import TimeoutExpiredError, retry
+from timeout_sampler import TimeoutExpiredError, TimeoutSampler
 
 from libs.vm.spec import Devices, Interface, Network, SpecDisk, VMISpec, Volume
 from libs.vm.vm import BaseVirtualMachine
+
+LOGGER = logging.getLogger(__name__)
 
 LOOKUP_IFACE_STATUS_TIMEOUT_SEC: Final[int] = 30
 RETRY_INTERVAL_SEC: Final[int] = 5
@@ -20,24 +23,31 @@ class VMInterfaceStatusNotFoundError(Exception):
     pass
 
 
+class VMInterfaceStatusStillExistsError(Exception):
+    pass
+
+
 def _default_interface_predicate(interface: ResourceField) -> bool:
     return "guest-agent" in interface["infoSource"] and interface[IP_ADDRESS]
 
 
 def lookup_iface_status(
-    vm: BaseVirtualMachine, iface_name: str, predicate: Callable[[Any], bool] = _default_interface_predicate
+    vm: BaseVirtualMachine,
+    iface_name: str,
+    predicate: Callable[[Any], bool] = _default_interface_predicate,
+    wait_timeout: int = LOOKUP_IFACE_STATUS_TIMEOUT_SEC,
 ) -> ResourceField:
     """
-    Returns the network interface status requested if found, otherwise raises VMInterfaceStatusNotFoundError.
-    The interface status information is expected to be sourced from the guest-agent with an IP address.
+    Awaits and returns the network interface status requested if found and the predicate function,
+    otherwise raises VMInterfaceStatusNotFoundError.
 
     Args:
         vm (BaseVirtualMachine): VM in which to search for the network interface.
         iface_name (str): The name of the requested interface.
         predicate (Callable[[dict[str, Any]], bool]): A function that takes a network interface as an argument
-            and returns a boolean value. This function should define the condition that
-            the interface needs to meet. By default, this looks for condition
-            "guest-agent" in interface["infoSource"] and interface[IP_ADDRESS]
+            and returns a boolean value. this function should define the condition that
+            the interface needs to meet.
+        wait_timeout (int): lookup operation timeout
 
     Returns:
         iface (ResourceField): The requested interface.
@@ -45,25 +55,28 @@ def lookup_iface_status(
     Raises:
         VMInterfaceStatusNotFoundError: If the requested interface was not found in the vmi status.
     """
+    sampler = TimeoutSampler(
+        wait_timeout=wait_timeout,
+        sleep=RETRY_INTERVAL_SEC,
+        func=_lookup_iface_status,
+        vm=vm,
+        iface_name=iface_name,
+        predicate=predicate,
+    )
     try:
-        return _lookup_iface_status(
-            vm=vm,
-            iface_name=iface_name,
-            predicate=predicate,
-        )
+        for iface in sampler:
+            if iface:
+                return iface
     except TimeoutExpiredError:
         raise VMInterfaceStatusNotFoundError(f"Network interface named {iface_name} was not found in VM {vm.name}.")
 
 
-@retry(
-    wait_timeout=LOOKUP_IFACE_STATUS_TIMEOUT_SEC,
-    sleep=RETRY_INTERVAL_SEC,
-    exceptions_dict={VMInterfaceStatusNotFoundError: []},
-)
-def _lookup_iface_status(vm: BaseVirtualMachine, iface_name: str, predicate: Callable[[Any], bool]) -> ResourceField:
+def _lookup_iface_status(
+    vm: BaseVirtualMachine, iface_name: str, predicate: Callable[[Any], bool]
+) -> ResourceField | None:
     """
     Returns the interface requested if found and the predicate function (to which the interface is
-    sent) returns True. Else, raise VMInterfaceStatusNotFoundError.
+    sent) Else, returns None.
 
     Args:
         vm (BaseVirtualMachine): VM in which to search for the network interface.
@@ -73,15 +86,12 @@ def _lookup_iface_status(vm: BaseVirtualMachine, iface_name: str, predicate: Cal
             the interface needs to meet.
 
     Returns:
-        iface (ResourceField): The requested interface.
-
-    Raises:
-        VMInterfaceStatusNotFoundError: If the requested interface was not found in the VM.
+        iface (ResourceField) | None: The requested interface or None
     """
     for iface in vm.vmi.interfaces:
         if iface.name == iface_name and predicate(iface):
             return iface
-    raise VMInterfaceStatusNotFoundError(f"Network interface named {iface_name} was not found in VM {vm.name}.")
+    return None
 
 
 def lookup_primary_network(vm: BaseVirtualMachine) -> Network:
--- a/tests/network/l2_bridge/test_bridge_nic_hot_plug.py
+++ b/tests/network/l2_bridge/test_bridge_nic_hot_plug.py
@@ -155,9 +155,6 @@ def hot_plugged_interface_name_on_vm_created_with_secondary_interface(
         hot_plugged_interface_name=hot_plugged_interface_name,
         net_attach_def_name=network_attachment_definition_for_hot_plug.name,
     )
-    # In order to complete the interface hot-plug, and have the interface available in the guest VM -
-    # the VM must be migrated.
-    migrate_vm_and_verify(vm=running_vm_with_secondary_and_hot_plugged_interfaces)
 
     return hot_plugged_interface_name
 
@@ -173,14 +170,6 @@ def hot_plugged_second_interface_with_address(
         ipv4_address=f"{IPV4_ADDRESS_SUBNET_PREFIX}.{next(index_number)}",
         vmi_interface=hot_plugged_interface_name_on_vm_created_with_secondary_interface,
     )
-
-
-@pytest.fixture()
-def migrated_vm_with_hot_plugged_interface_attached(running_vm_for_nic_hot_plug):
-    # In order to complete the interface hot-plug, and have the interface available in the guest VM -
-    # the VM must be migrated.
-    migrate_vm_and_verify(vm=running_vm_for_nic_hot_plug)
-    return running_vm_for_nic_hot_plug
 
 
 @pytest.fixture()
@@ -358,7 +347,7 @@ def hot_unplugged_additional_interface(
         vm=running_vm_with_secondary_and_hot_plugged_interfaces,
         hot_plugged_interface_name=hot_plugged_interface_name_on_vm_created_with_secondary_interface,
     )
-    migrate_vm_and_verify(vm=running_vm_with_secondary_and_hot_plugged_interfaces)
+
     return hot_plugged_interface_name_on_vm_created_with_secondary_interface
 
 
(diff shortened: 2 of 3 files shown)


tests/network/l2_bridge/utils.py
```diff
@@ -123,15 +122,60 @@ def hot_plug_interface(
...
     })
 
+    migration_id_before_hotplug = getattr(vm.vmi.instance.status.migrationState, "migrationUid", None)
+    node_before = vm.vmi.node
+
     update_hot_plug_config_in_vm(vm=vm, interfaces=interfaces, networks=networks)
 
+    if wait_for_migration:
+        wait_for_automatic_migration(
+            vm=vm, migration_id_before_hotplug=migration_id_before_hotplug, node_before=node_before
```
Anatw, line 132: Better check that migration_id_before_hotplug is not None before using it.

tests/network/l2_bridge/utils.py
```diff
@@ -123,15 +122,60 @@ def hot_plug_interface(
...
+        )
 
-def hot_unplug_interface(vm, hot_plugged_interface_name):
+
+def hot_unplug_interface(vm, hot_plugged_interface_name, wait_for_migration=True):
     interfaces = vm.get_interfaces()
     unplugged_interface = next(interface for interface in interfaces if interface["name"] == hot_plugged_interface_name)
     unplugged_interface.update(dict(state="absent"))
+
+    migration_id_before_hotplug = getattr(vm.vmi.instance.status.migrationState, "migrationUid", None)
```
Anatw, line 141: Ditto (migration_id_before_hotplug)

tests/network/l2_bridge/utils.py
```diff
@@ -123,15 +122,60 @@ def hot_plug_interface(
...
+    vm: VirtualMachineForTests,
+    migration_id_before_hotplug: str,
+    node_before: str,
+    timeout: int = TIMEOUT_2MIN,
+) -> None:
+    try:
+        for sample in TimeoutSampler(
+            wait_timeout=timeout,
+            sleep=TIMEOUT_5SEC,
+            func=lambda: getattr(vm.vmi.instance.status.migrationState, "migrationUid", None)
```
Anatw, line 162: This is the third usage of this line. How about moving it into a function? I'm referring to the line: 
getattr(vm.vmi.instance.status.migrationState, "migrationUid", None).
You can also assert in that function that it is not None.

Anatw (commented):
Thank you!

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
```diff
@@ -523,12 +531,15 @@ def test_basic_connectivity_of_hot_plugged_interface(
     )
     def test_basic_connectivity_of_hot_plugged_interface_after_second_migration(
         self,
+        running_vm_for_nic_hot_plug,
         running_utility_vm_for_connectivity_check,
         network_attachment_definition_for_hot_plug,
-        migrated_vm_with_hot_plugged_interface_attached,
+        unprivileged_client,
+        unprivileged_client_migration_permission_role_binding,
```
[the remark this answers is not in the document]
servolkov: Very good question. 

>The approach starting from 4.20 is that only cluster admins can perform migration.

Technically, manual migration can be performed by either namespace admin with permissions or cluster admin - this functionality is provided (like in 4.19). So, why not to use namespace admin for it?

EdDev (changes requested):
Thank you @servolkov .

I am trying to understand if this was not over-complicated. See inline comments.

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
```diff
@@ -502,13 +510,13 @@ def test_multiple_interfaces_hot_plugged(
     )
     def test_basic_connectivity_of_hot_plugged_interface(
         self,
-        migrated_vm_with_hot_plugged_interface_attached,
+        running_vm_for_nic_hot_plug,
```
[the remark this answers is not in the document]
yossisegev: Actually, [I had similar thing to say about this change too](https://github.com/RedHatQE/openshift-virtualization-tests/pull/1247/files#r2161069795), but I was in minority in that case :) , so I can only guess it will be the same here.
Very well, it may not be straightforward, but still the call to the hotplug fixture can imply that the VM that is involved in this flow has undergone hot-plug.
You can resolve this conversation (unless I managed to covince you after all :) ).

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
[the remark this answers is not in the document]
yossisegev: Because "_namespace admin with permissions_" is a special "mutation", which I would consider to be a workaround.
And this is exactly the case in 4.19 - a temporary W/A which we recommend to our users until they can upgrade to 4.20. So I don't think it reflects the user scenario.

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
```diff
@@ -69,7 +89,7 @@ def network_attachment_definition_for_hot_plug(
 
 
 @pytest.fixture(scope="class")
-def hot_plugged_interface_name(
+def hot_plugged_interface(
```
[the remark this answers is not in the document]
yossisegev: I'm pasting my comment about it here, so we can continue the discussion in one thread:

As this fixture actually returns something (and not only creating), its name should reflect what it returns.
Why does it matter? Because `hot_plugged_interface` gives the impression that it returns an interface object, while it only returns the name. One could refer to this fixture in the code when needing to have the object - and it will fail.
In general, you need a justification for changing code, but I didn't find one. The comment you provided in the PR description is unclear to me:
> Minor refactoring was done to align fixture naming with a consistent style (like hot_plugged_interface_name -> hot_plugged_interface).

What consistent style are you referring to?
If you end up keeping this change, then you should also fix its references in the code, where it is used for getting the interface name - instead of referring to `hot_plugged_interface`, it will now have to refer to something like `hot_plugged_interface.name`.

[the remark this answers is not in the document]
yossisegev: Very well, I copied this comment to that thread so we can continue there. Please resolve this conversation so we can avoid the duplication.

yossisegev (changes requested):

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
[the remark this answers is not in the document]
EdDev: Without entering what is better and preferred, this PR should not change the style or try to organize too much. If there is such a need, as a pre-requirement for adjusting the tests here for the rollout, then it should be done in a separate PR, dedicated for that specific thing. Sometime a different commit can also help.

As a rule, performing more than one thing in a commit will make it harder for reviewers to review the change.
Which in turn will make the review process longer, delaying the potential agreement and merge.
In order to approve your PR, I prefer to see the minimum needed so I can focus on that and not track side-changes that are not directly related.

I encourage to send PRs to solve inconsistency and improve naming. We have a lot to refactor and improve, so having small PRs towards that goal is a welcome thing.

tests/network/l2_bridge/test_bridge_nic_hot_plug.py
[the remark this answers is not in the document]
EdDev: I agree with @yossisegev , but I still do not understand what this test is checking.

@yossisegev , can you please reason why we need this test scenario?
I see no meaning to migrate a VM after a hotplug/hounplug succeeded, it is no different from migrating a VM that had no hotplug/hotunplug.

If it is not needed, we need it dropped in a separate PR.

tests/network/l2_bridge/utils.py
```diff
@@ -113,6 +111,7 @@ def hot_plug_interface(
     hot_plugged_interface_name,
     net_attach_def_name,
     sriov=False,
+    wait_for_migration=True,
```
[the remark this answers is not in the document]
EdDev: > > The only reason to get into the details of checking the migration, is to try and fail sooner
> 
> I agree that migration is an encapsulated implementation detail for the user. But I think waiting for explicitly automatic migration is a way to simplify crash investigation. If an error occurs during the migration process, you will receive a report about it. Otherwise, you will need additional manual investigation of this failure, it takes time. Why not to use it then? (this awaiting does not affect user scenarios).

There are logs and events which are expected to be examined on failure.
Based on this reasoning, you should track many other intermediate states a VM passes so you can fail fast or be able to debug. I do not think the implementation should guide the test and if there are some things to monitor, then I would only use the direct ones and not side-effects.

> Why not to use it then? (this awaiting does not affect user scenarios).

Users care about the interface being plugged or not, they do not care about the migration.
They do care if something is wrong to understand why it did not work, and there I would expect a negative test with an assertion on the failure which a user is expected to "see".

> > I realized that there must be a check anyway about the interface appearing or disappearing from the VMI status.
> 
> got it, on the way

The one that exists is also needed by @frenzyfriday but requires some generalization from its current state.
You may want to sync.

tests/network/l2_bridge/utils.py
servolkov: Not relevant anymore

tests/network/l2_bridge/utils.py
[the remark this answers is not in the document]
servolkov: done without migration status

tests/network/l2_bridge/utils.py
servolkov: not relevant anymore

tests/network/l2_bridge/utils.py
```diff
@@ -125,13 +123,23 @@ def hot_plug_interface(
 
     update_hot_plug_config_in_vm(vm=vm, interfaces=interfaces, networks=networks)
 
+    lookup_iface_status(
+        vm=vm,
+        iface_name=hot_plugged_interface_name,
+        predicate=lambda interface: "guest-agent" in interface["infoSource"] and interface["linkState"] == "up",
```
servolkov, line 129: not sure if there is a need to have `interface["linkState"] == "up"`, I added "just in case"

libs/net/vmspec.py
```diff
@@ -20,40 +23,64 @@ class VMInterfaceStatusNotFoundError(Exception):
...
         VMInterfaceStatusNotFoundError: If the requested interface was not found in the vmi status.
     """
+    sampler = TimeoutSampler(
+        wait_timeout=wait_timeout,
+        sleep=RETRY_INTERVAL_SEC,
+        func=_lookup_iface_status,
+        vm=vm,
+        iface_name=iface_name,
+        predicate=predicate,
+    )
```
EdDev, line 65: Dropping the decorator in favor of this direct usage is not trivial and requires a good reasoning.
Please reason for this by dedicating a commit for this change with a convincing commit message.

libs/net/vmspec.py
```diff
@@ -63,15 +90,38 @@ def _lookup_iface_status(vm: BaseVirtualMachine, iface_name: str, predicate: Cal
...
+        wait_timeout (int): The maximum time (in seconds) to wait for the interface to be deleted.
+
+    Raises:
+        VMInterfaceStatusStillExistsError: If the interface still exists after the timeout period.
+    """
+    sampler = TimeoutSampler(
+        wait_timeout=wait_timeout,
+        sleep=RETRY_INTERVAL_SEC,
+        func=lambda: all(iface.name != interface_name for iface in vm.vmi.interfaces),
+    )
```
EdDev, line 117: Please use the `retry` decorator.

tests/network/l2_bridge/utils.py
EdDev: The unplug should not have any relation to the link-state, therefore I would not expect it to me mentioned here unless you are attempting to have a special test with some combination in mind.

I would also expect here to use no special predicate because this should be a simple scenario.
It looks like we should drop the "address" requirement which is not a fit in many cases and use an explicit predicate for when it is a requirement. For that, we'll need yet another PR to treat the current usages.

libs/net/vmspec.py
```diff
@@ -63,15 +90,38 @@ def _lookup_iface_status(vm: BaseVirtualMachine, iface_name: str, predicate: Cal
...
+        iface (ResourceField) | None: The requested interface or None
     """
     for iface in vm.vmi.interfaces:
         if iface.name == iface_name and predicate(iface):
             return iface
-    raise VMInterfaceStatusNotFoundError(f"Network interface named {iface_name} was not found in VM {vm.name}.")
+    return None
+
+
+def wait_for_iface_status_deleted(vm: BaseVirtualMachine, interface_name: str, wait_timeout: int = 30) -> None:
```
EdDev, line 101: The name implies as if you know who called this function. The name should reflect the operation intent with no knowledge about why it was called (e.g. it may be because one unplugged the interface and it may be because one wants to confirm the specific interface is not there in the first place).

Alternative names:
- `wait_for_missing_iface_status`
- `confirm_missing_iface_status`

libs/net/vmspec.py
```diff
@@ -20,40 +23,64 @@ class VMInterfaceStatusNotFoundError(Exception):
     pass
 
 
-def lookup_iface_status(vm: BaseVirtualMachine, iface_name: str) -> ResourceField:
+class VMInterfaceStatusStillExistsError(Exception):
+    pass
+
+
+def _default_interface_predicate(interface: ResourceField) -> bool:
+    return "guest-agent" in interface["infoSource"] and interface[IP_ADDRESS]
```
yossisegev, line 31: Are you sure that having an IP address is a condition that the interface must follow?

tests/network/l2_bridge/utils.py
yossisegev: > not sure if there is a need to have interface["linkState"] == "up", I added "just in case"

Hot plugging an interface doesn't mean committing to specific initial state,but rather just to have the interface  added to the VM. In other words - I think that the linkState validation should be removed.

yossisegev (changes requested):

libs/net/vmspec.py
EdDev: The reasoning for not using `retry` in favor of the underlying function is accepted by itself, that is indeed the cons of the decorator. However, you will have to convince us that a flexible timeout is indeed required.
In most cases, it is not required and even harming. The need to maintain the timeout of many instances needs a good case, as it is expensive. E.g. if we need to increase the timeout, we may potentially need to go all over the codebase to do it. Moreover, flakes may appear in one case and not in the other because there are differences.
On the other hand, if we decide on an agreed worse case scenario in a test for waiting on something, it may be good enough, saving us from the need to choose the best/recommended time.
Many times you will be asked to provide such a recommended time in a form of a const, which is very similar to this case.

But in some cases, there are always exceptions. Does this one requires such an exception?

tests/network/l2_bridge/utils.py
```diff
@@ -14,6 +15,7 @@
     NODE_TYPE_WORKER_LABEL,
     SRIOV,
     TIMEOUT_1MIN,
+    TIMEOUT_2MIN,
```
EdDev, line 18: I usually nack usages of these useless consts, but in case the existing module already uses it, consistency wins.

libs/net/vmspec.py
servolkov: >ASAIU this does not exist yet, i.e. the edge case. If it does, please elaborate why the current default is not good enough.

Why does not exist? If I need to change lib func due to some new case occurs and the default value is not suitable - that is the case. I answered already here - https://github.com/RedHatQE/openshift-virtualization-tests/pull/1346/files#r2188045105: `I encountered with such issue when I wanted to use lookup_iface_status function - timeout was hardcoded and not applicable in my case.`  The default is 30 sec - not enough for hot-plug.

>The usage of the function is very noisy and much harder to read than the decorator, therefore, unless there is a very string convincing argument that it is required in this specific case and now, I would keep the simple and elegant current decorator.

Agree with the readability of this cycle. But I don't get this argument. We are discussing the need for flexibility in function behavior. In this context, the argument about elegance is not understandable for me. In general, I should not have to compromise on functional adaptability (if it's really needed) just to preserve a aesthetic simplicity.

> I want to use it only if the case in question really needs it... e.g. the default is 2min and we need a 6min timeout. If the difference is to increase all by 1min, I'll prefer that.

I can only repeat again - you can’t know in advance whether flexibility will be needed. What if it will be 6 min for some case in the future?

>My concern is the ability to have permutations

Flakes may appear with a global timeout as well. As a user of a function, I usually I don’t tend to change its internals. During debugging, I might hit the default internal timeout and pass, but later face sporadic failures if my case really needed more time. Developers may stop thinking critically about appropriate timing for their tests. If the timeout is a parameter with default, it's a sign that usually operation takes the default value, but it might be different in your case. 

Changing a global timeout can also be harmful: I might unknowingly "fix" someone else's performance issue by extending the timeout, masking deeper problems. As well as increasing test runtime on failures.

servolkov: a dedicated commit for refactoring was added

libs/net/vmspec.py
servolkov: same function as lookup. I believe that the code should be consistent if we agreed in this thread https://github.com/RedHatQE/openshift-virtualization-tests/pull/1346/files#r2188044042

libs/net/vmspec.py
rnetser: @servolkov if `_lookup_iface_status` requires retry (as the current code implies) then if the retry timeout needs to be controlled, the retyr mechanism should be moved inside `_lookup_iface_status` and not to function(s) which call it

servolkov: I'm probably missing something, but I still don’t understand the concern about `_lookup_iface_status`.
 
 > if the retry timeout needs to be controlled, the retyr mechanism should be moved inside `_lookup_iface_status`
 
 but why? This is already handled in`lookup_iface_status`. You don't need to think of `_lookup_iface_status` at all - it is a private helper for `lookup_iface_status`, the atomic unit.
 
`lookup_iface_status` handles the retry logic and error handling, we have all flexibility there, this is the main func. If any other place needs retry behavior, they should call `lookup_iface_status`, not `_lookup_iface_status`.

I’m simply following a common pattern: one small atomic function that just performs a single task (lookup), and another function that wraps it with retries and error handling.

EdDev (approved):
Thank you!

/approve

yossisegev (approved):

servolkov:
/verified 

PSI 4.20 cluster:
uv run  pytest  --cluster-sanity-skip-nodes-check --cluster-sanity-skip-storage-check --skip-deprecated-api-test tests/network/l2_bridge/test_bridge_nic_hot_plug.py

BM cluster with SRIOV is not available and SRIOV test was not verified.

libs/net/vmspec.py
```diff
@@ -20,50 +21,60 @@ class VMInterfaceStatusNotFoundError(Exception):
...
         raise VMInterfaceStatusNotFoundError(f"Network interface named {iface_name} was not found in VM {vm.name}.")
 
 
-@retry(
-    wait_timeout=LOOKUP_IFACE_STATUS_TIMEOUT_SEC,
-    sleep=RETRY_INTERVAL_SEC,
-    exceptions_dict={VMInterfaceStatusNotFoundError: []},
-)
-def _lookup_iface_status(vm: BaseVirtualMachine, iface_name: str, predicate: Callable[[Any], bool]) -> ResourceField:
+def _lookup_iface_status(
```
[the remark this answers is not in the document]
servolkov: it is used now not only inside `lookup_iface_status`. About the order - I just leaved it how in the order how it was before (easier to review).

libs/net/vmspec.py
```diff
@@ -73,15 +84,37 @@ def _lookup_iface_status(vm: BaseVirtualMachine, iface_name: str, predicate: Cal
...
+            vm (BaseVirtualMachine): The virtual machine to check for the interface status.
+            iface_name: (str): The name of the network interface to wait for deletion.
+
+        Raises:
+            VMInterfaceStatusStillExistsError: If the interface still exists after the timeout period.
+    there was a usage of it
+        Returns:
+            bool: True if the interface is missing otherwise raises an exception.
+    """
+    if _lookup_iface_status(vm=vm, iface_name=iface_name, predicate=lambda _: True) is not None:
```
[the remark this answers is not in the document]
servolkov: This func isn't strictly about removal - it checks for the absence of an interface, which might be useful in other scenarios not directly related to removal.

libs/net/vmspec.py
```diff
@@ -20,50 +21,60 @@ class VMInterfaceStatusNotFoundError(Exception):
...
     Args:
         vm (BaseVirtualMachine): VM in which to search for the network interface.
         iface_name (str): The name of the requested interface.
         predicate (Callable[[dict[str, Any]], bool]): A function that takes a network interface as an argument
-            and returns a boolean value. This function should define the condition that
-            the interface needs to meet. By default, this looks for condition
-            "guest-agent" in interface["infoSource"] and interface[IP_ADDRESS]
+            and returns a boolean value. this function should define the condition that
+            the interface needs to meet.
+        timeout (int): lookup operation timeout
```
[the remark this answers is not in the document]
servolkov: thx, will be fixed in the next PR

libs/net/vmspec.py
```diff
@@ -20,50 +21,60 @@ class VMInterfaceStatusNotFoundError(Exception):
...
+    otherwise raises VMInterfaceStatusNotFoundError.
 
     Args:
         vm (BaseVirtualMachine): VM in which to search for the network interface.
         iface_name (str): The name of the requested interface.
         predicate (Callable[[dict[str, Any]], bool]): A function that takes a network interface as an argument
-            and returns a boolean value. This function should define the condition that
-            the interface needs to meet. By default, this looks for condition
-            "guest-agent" in interface["infoSource"] and interface[IP_ADDRESS]
+            and returns a boolean value. this function should define the condition that
```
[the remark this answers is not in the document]
servolkov: thx, will be fixed in the next PR

openshift-virtualization-qe-bot-2:

PR #1364, opened 2025-07-03
Give grace for connectivity process to appear

yossisegev:
Tests that use this function for verifying that iperf3 is running on the VM occasionally fail, in case the process is not found yet without a real retry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of process detection by introducing a timeout-based retry mechanism, ensuring more robust handling of process status checks.

### DIFF
--- a/libs/net/traffic_generator.py
+++ b/libs/net/traffic_generator.py
@@ -107,7 +107,6 @@ def _is_process_running(vm: BaseVirtualMachine, cmd: str) -> bool:
     except TimeoutExpiredError as e:
         LOGGER.warning(f"Process is not running on VM {vm.name}. Error: {str(e.last_exp)}")
         return False
-    return False
 
 
 def is_tcp_connection(server: Server, client: Client) -> bool:
--- a/libs/net/traffic_generator.py
+++ b/libs/net/traffic_generator.py
@@ -2,6 +2,7 @@ import logging
 from typing import Final
 
 from ocp_utilities.exceptions import CommandExecFailed
+from timeout_sampler import TimeoutExpiredError, TimeoutSampler
 
 from libs.vm.vm import BaseVirtualMachine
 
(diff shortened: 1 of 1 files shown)


dshchedr (dismissed):

libs/net/traffic_generator.py
```diff
@@ -94,13 +95,17 @@ def _stop_process(vm: BaseVirtualMachine, cmd: str) -> None:
...
             timeout=_DEFAULT_CMD_TIMEOUT_SEC,
-        )
-        return True
-    except CommandExecFailed as e:
-        LOGGER.info(f"Process is not running on VM {vm.name}. Error: {str(e)}")
+        ):
+            if sample:
+                return True
+    except TimeoutExpiredError as e:
+        LOGGER.info(f"Process is not running on VM {vm.name}. Error: {str(e.last_exp)}")
```
rnetser, line 108: if you're already touching the code, i'd consider changing to `warning`

openshift-virtualization-qe-bot-2:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-1364 published

libs/net/traffic_generator.py
```diff
@@ -94,13 +95,17 @@ def _stop_process(vm: BaseVirtualMachine, cmd: str) -> None:
 
 def _is_process_running(vm: BaseVirtualMachine, cmd: str) -> bool:
     try:
-        vm.console(
-            commands=[f"pgrep -fAx '{cmd}'"],
+        for sample in TimeoutSampler(
+            wait_timeout=60,
+            sleep=5,
+            func=vm.console,
+            commands=[f"pgrep -fx '{cmd}'"],
```
frenzyfriday, line 102: [Question Non Blocking]: Comparing with the old version - we don't need the -A flag anymore?

frenzyfriday (approved):
lgtm

libs/net/traffic_generator.py
yossisegev: @frenzyfriday Indeed the coderabbit recommended removing `-A`.
It is indeed unnecessary, because the command is sent frpm withing the context of the VM console, so there is no ancestor here (like bash or a new shell) which should be ignored.
Thank you for pointing this out and raising this discussion.

libs/net/traffic_generator.py
```diff
@@ -94,13 +95,17 @@ def _stop_process(vm: BaseVirtualMachine, cmd: str) -> None:
 
 def _is_process_running(vm: BaseVirtualMachine, cmd: str) -> bool:
     try:
-        vm.console(
-            commands=[f"pgrep -fAx '{cmd}'"],
+        for sample in TimeoutSampler(
```
servolkov, line 98: you can consider using a retry decorator for such case

RoniKishner (approved):

libs/net/traffic_generator.py
```diff
@@ -94,14 +95,19 @@ def _stop_process(vm: BaseVirtualMachine, cmd: str) -> None:
...
-        return True
-    except CommandExecFailed as e:
-        LOGGER.info(f"Process is not running on VM {vm.name}. Error: {str(e)}")
+        ):
+            if sample:
+                return True
+    except TimeoutExpiredError as e:
+        LOGGER.warning(f"Process is not running on VM {vm.name}. Error: {str(e.last_exp)}")
         return False
+    return False
```
rnetser, line 110: based on slack, please add mypy ignore on the function as the sampler behaves differently and this line will not be reached. ty

rnetser (changes requested):

openshift-virtualization-qe-bot-2:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-1364.

PR #1407, opened 2025-07-08
s390x marker for network tests

HarshithaMS005:
##### Short description:
Added s390x marker for network tests

### DIFF
--- a/tests/network/bond/test_bond_modes.py
+++ b/tests/network/bond/test_bond_modes.py
@@ -195,17 +195,20 @@ def bond_resource(index_number, nodes_available_nics, worker_node1):
 
 
 @pytest.mark.polarion("CNV-4382")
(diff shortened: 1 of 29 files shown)


yossisegev (changes requested):
Please specify in the PR message what are the criteria for picking tests for s390x, so we can better understand the rationale behind why certain tests were marked when we review the PR.
Thank you

rnetser:
@HarshithaMS005 this PR should be against `main`

HarshithaMS005:
Hi @rnetser, I have rebased against main. Thank you 😊

EdDev (commented):
Can we get an explanation of what this is?
And how the sig-network is supposed to maintain this? I am mainly worried about the maintability of it.

By explanation I mean commit + PR description or a ref to something.

PR #1411, opened 2025-07-08
Add centos tests with common instance types

RoniKishner:
##### Short description:
Add centos tests with common instance types

##### More details:
 - Add `instance_type_centos_os_matrix` to global_config
 - Added `test_centos_os.py`
 - Update `generate_instance_type_rhel_os_matrix` to `generate_instance_type_os_matrix` to work for all OS names

Add tests supporting CentOs operating systems

##### jira-ticket:

https://issues.redhat.com/browse/CNV-54690

## Summary by CodeRabbit

* **New Features**
  * Added new tests for CentOS VM instance types, including creation, validation, and deletion scenarios.
  * Introduced support for generating CentOS instance type OS matrices, similar to existing RHEL support.

* **Tests**
  * Implemented comprehensive test coverage for CentOS-based VM lifecycle and guest agent validation.

* **Chores**
  * Enhanced configuration to include CentOS instance type OS matrices for improved test parameterization.
  * Generalized instance type OS matrix generation to support multiple OS types with updated version notation.

### DIFF
--- a/tests/global_config_x86_64.py
+++ b/tests/global_config_x86_64.py
@@ -1,7 +1,17 @@
 from typing import Any
 
+from utilities.constants import (
+    CENTOS_STREAM9_PREFERENCE,
+    CENTOS_STREAM10_PREFERENCE,
+    RHEL8_PREFERENCE,
+    RHEL9_PREFERENCE,
+    RHEL10_PREFERENCE,
+)
 from utilities.infra import get_latest_os_dict_list
-from utilities.os_utils import generate_instance_type_rhel_os_matrix, generate_os_matrix_dict
+from utilities.os_utils import (
+    generate_linux_instance_type_os_matrix,
+    generate_os_matrix_dict,
+)
 
 global config
 
@@ -28,7 +38,12 @@ windows_os_matrix = generate_os_matrix_dict(
 
 fedora_os_matrix = generate_os_matrix_dict(os_name="fedora", supported_operating_systems=["fedora-42"])
 centos_os_matrix = generate_os_matrix_dict(os_name="centos", supported_operating_systems=["centos-stream-9"])
-instance_type_rhel_os_matrix = generate_instance_type_rhel_os_matrix(preferences=["rhel-8", "rhel-9", "rhel-10"])
+instance_type_rhel_os_matrix = generate_linux_instance_type_os_matrix(
+    os_name="rhel", preferences=[RHEL8_PREFERENCE, RHEL9_PREFERENCE, RHEL10_PREFERENCE]
+)
+instance_type_centos_os_matrix = generate_linux_instance_type_os_matrix(
+    os_name="centos.stream", preferences=[CENTOS_STREAM9_PREFERENCE, CENTOS_STREAM10_PREFERENCE]
+)
 
 (
     latest_rhel_os_dict,
--- a/tests/infrastructure/instance_types/supported_os/conftest.py
+++ b/tests/infrastructure/instance_types/supported_os/conftest.py
@@ -6,8 +6,7 @@ from utilities.constants import DATA_SOURCE_NAME
 
 @pytest.fixture(scope="class")
 def xfail_if_rhel8(instance_type_rhel_os_matrix__module__):
-    current_rhel_name = [*instance_type_rhel_os_matrix__module__][0]
-    if current_rhel_name == "rhel-8":
+    if [*instance_type_rhel_os_matrix__module__][0] == "rhel.8":
         pytest.xfail("EFI is not enabled by default before RHEL9")
 
 
@@ -28,4 +27,24 @@ def golden_image_rhel_vm_with_instance_type(
         modern_cpu_for_migration=modern_cpu_for_migration,
         storage_class_name=[*storage_class_matrix__module__][0],
         data_source_name=instance_type_rhel_os_matrix__module__[os_name][DATA_SOURCE_NAME],
+    )
+
+
+@pytest.fixture(scope="module")
+def golden_image_centos_vm_with_instance_type(
+    unprivileged_client,
+    namespace,
+    golden_images_namespace,
+    modern_cpu_for_migration,
+    instance_type_centos_os_matrix__module__,
+    storage_class_matrix__module__,
+):
+    os_name = next(iter(instance_type_centos_os_matrix__module__))
+    return golden_image_vm_with_instance_type(
+        client=unprivileged_client,
+        namespace_name=namespace.name,
+        golden_images_namespace_name=golden_images_namespace.name,
+        modern_cpu_for_migration=modern_cpu_for_migration,
+        storage_class_name=[*storage_class_matrix__module__][0],
+        data_source_name=instance_type_centos_os_matrix__module__[os_name][DATA_SOURCE_NAME],
     )
--- a/utilities/os_utils.py
+++ b/utilities/os_utils.py
@@ -1,4 +1,6 @@
+import logging
 import os
+import re
 from typing import Any
 
 from ocp_resources.template import Template
@@ -21,6 +23,9 @@ from utilities.constants import (
     WORKLOAD_STR,
     Images,
 )
+
+LOGGER = logging.getLogger(__name__)
+
 
 RHEL_OS_MAPPING: dict[str, dict[str, Any]] = {
     WORKLOAD_STR: Template.Workload.SERVER,
@@ -218,30 +223,41 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
     return os_formatted_list
 
 
-def generate_instance_type_rhel_os_matrix(preferences: list[str]) -> list[dict[str, dict[str, Any]]]:
+def generate_linux_instance_type_os_matrix(os_name: str, preferences: list[str]) -> list[dict[str, dict[str, Any]]]:
     """
-    Generate a list of dictionaries representing the instance type matrix for RHEL OS.
-
+    Generate a list of dictionaries representing the instance type matrix for a Linux OS type.
     Each dictionary represents a specific instance type and its configuration.
 
     Args:
-        preferences (list[str]): A list of preferences for the instance types. Preference format is "rhel-<version>".
+        os_name (str): The name of the OS.
+        preferences (list[str]): A list of preferences for the instance types. Preference format is "<os>.<version>".
 
     Returns:
         list[dict[str, dict[str, Any]]]: A list of dictionaries representing the instance type matrix.
-    """
-    latest_rhel = f"rhel-{max([preference.split('-')[1] for preference in preferences], key=int)}"
 
+    """
+
+    def _extract_version(pref: str) -> Any:
+        match = re.search(r"\d+", pref)
+        return int(match.group()) if match else None
+
+    def _format_data_source_name(preference_name: str) -> str:
+        version = _extract_version(pref=preference_name)
+        if version is None:
+            return preference_name
+        return f"{os_name.replace('.', '-')}{version}"
+
+    latest_os = max(preferences, key=_extract_version)
     instance_types: list[dict[str, dict[str, Any]]] = []
 
     for preference in preferences:
         preference_config: dict[str, Any] = {
-            PREFERENCE_STR: preference.replace("-", "."),
-            DATA_SOURCE_NAME: preference.replace("-", ""),
+            PREFERENCE_STR: preference,
+            DATA_SOURCE_NAME: _format_data_source_name(preference_name=preference),
         }
-        if preference == latest_rhel:
+
+        if preference == latest_os:
             preference_config[LATEST_RELEASE_STR] = True
 
         instance_types.append({preference: preference_config})
-
     return instance_types
--- a/tests/global_config_arm64.py
+++ b/tests/global_config_arm64.py
@@ -7,12 +7,15 @@ from utilities.constants import (
     EXPECTED_CLUSTER_INSTANCE_TYPE_LABELS,
     HPP_CAPABILITIES,
     PREFERENCE_STR,
+    RHEL8_PREFERENCE,
+    RHEL9_PREFERENCE,
+    RHEL10_PREFERENCE,
     Images,
     StorageClassNames,
 )
 from utilities.infra import get_latest_os_dict_list
 from utilities.os_utils import (
-    generate_instance_type_rhel_os_matrix,
+    generate_linux_instance_type_os_matrix,
     generate_os_matrix_dict,
 )
 from utilities.storage import HppCsiStorageClass
@@ -54,7 +57,9 @@ rhel_os_matrix = generate_os_matrix_dict(os_name="rhel", supported_operating_sys
 latest_rhel_os_dict = get_latest_os_dict_list(os_list=[rhel_os_matrix])[0]
 
 # Modify instance_type_rhel_os_matrix for arm64
-instance_type_rhel_os_matrix = generate_instance_type_rhel_os_matrix(preferences=["rhel-8", "rhel-9", "rhel-10"])
+instance_type_rhel_os_matrix = generate_linux_instance_type_os_matrix(
+    os_name="rhel", preferences=[RHEL8_PREFERENCE, RHEL9_PREFERENCE, RHEL10_PREFERENCE]
+)
 for os_matrix_dict in instance_type_rhel_os_matrix:
     for os_params in os_matrix_dict.values():
         os_params[PREFERENCE_STR] += f".{ARM_64}"
--- a/tests/global_config_s390x.py
+++ b/tests/global_config_s390x.py
@@ -7,7 +7,7 @@ from utilities.constants import (
     S390X,
 )
 from utilities.infra import get_latest_os_dict_list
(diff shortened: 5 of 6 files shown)


tests/infrastructure/instance_types/supported_os/conftest.py
```diff
@@ -1,40 +1,53 @@
...
 @pytest.fixture(scope="class")
 def skip_if_rhel8(instance_type_rhel_os_matrix__module__):
-    current_rhel_name = [*instance_type_rhel_os_matrix__module__][0]
-    if current_rhel_name == "rhel-8":
+    if [*instance_type_rhel_os_matrix__module__][0] == "rhel-8":
         pytest.xfail("EFI is not enabled by default before RHEL9")
 
 
+@pytest.fixture(scope="class")
+def skip_if_centos(instance_type_centos_fedora_os_matrix__module__):
```
geetikakay, line 13: rename function to use xfail just to be in sync with other xfail fixtures in repo
ditto with skip_if_rhel8

tests/infrastructure/instance_types/supported_os/utils.py
```diff
@@ -0,0 +1,31 @@
+from ocp_resources.data_source import DataSource
+
+from utilities.constants import DATA_SOURCE_NAME
+from utilities.storage import data_volume_template_with_source_ref_dict
+from utilities.virt import VirtualMachineForTests
+
+
+def _golden_image_vm_with_instance_type(
```
geetikakay, line 8: since it is not used internally by a parent function maybe remove _. 
use hints

tests/global_config.py
```diff
@@ -436,6 +436,33 @@ def _get_default_storage_class(sc_list):
     },
 ]
 
+instance_type_centos_fedora_os_matrix = [
```
geetikakay, line 439: should we keep centos fedora seperate?

tests/global_config.py
```diff
@@ -436,6 +436,33 @@ def _get_default_storage_class(sc_list):
...
+    {
+        "fedora": {
+            DV_SIZE_STR: Images.Fedora.DEFAULT_DV_SIZE,
+            INSTANCE_TYPE_STR: "u1.medium",
+            PREFERENCE_STR: "fedora",
```
geetikakay, line 444: use constants

tests/global_config.py
```diff
@@ -436,6 +436,33 @@ def _get_default_storage_class(sc_list):
...
+            DATA_SOURCE_NAME: "fedora",
+        }
+    },
+    {
+        "centos-stream9": {
+            DV_SIZE_STR: Images.CentOS.DEFAULT_DV_SIZE,
+            INSTANCE_TYPE_STR: "u1.medium",
+            PREFERENCE_STR: "centos.stream9",
```
geetikakay, line 452: ditto

tests/global_config.py
RoniKishner: Iv'e considered this, but I feel like it's a waste of repeated code since they are very similar.

tests/infrastructure/instance_types/supported_os/utils.py
```diff
@@ -0,0 +1,31 @@
+from ocp_resources.data_source import DataSource
+
+from utilities.constants import DATA_SOURCE_NAME
+from utilities.storage import data_volume_template_with_source_ref_dict
+from utilities.virt import VirtualMachineForTests
+
+
+def golden_image_vm_with_instance_type(
```
geetikakay, line 8: use hints

tests/global_config.py
rnetser: i agree with @geetikakay

keep it simple;
you also have `@pytest.mark.usefixtures("xfail_if_centos")` wchih literally runs only on a subset of the matrix.

tests/global_config.py
```diff
@@ -397,6 +399,33 @@ def _get_default_storage_class(sc_list):
     },
 ]
 
+instance_type_centos_fedora_os_matrix = [
+    {
+        OS_FLAVOR_FEDORA: {
```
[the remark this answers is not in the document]
RoniKishner: It's using the fedora DataSource so I wouldn't recommend it, since it will be auto updated on it's on, where as `fedora_os_matrix` uses a constant file.

utilities/os_utils.py
```diff
@@ -240,3 +241,67 @@ def generate_instance_type_rhel_os_matrix(preferences: list[str]) -> list[dict[s
...
+        preference_config: dict[str, Any] = {
+            PREFERENCE_STR: preference.replace("-", "."),
+            DATA_SOURCE_NAME: preference,
+        }
+        if preference == latest_centos:
+            preference_config[LATEST_RELEASE_STR] = True
+
+        instance_types.append({preference: preference_config})
+
+    return instance_types
```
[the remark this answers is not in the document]
RoniKishner: 1. The regex fits to the naming convention of the preferences we use
2. the different format is what we expect to be used
3. The string naming convention is what we usually use in the repository

utilities/os_utils.py
```diff
@@ -240,3 +241,67 @@ def generate_instance_type_rhel_os_matrix(preferences: list[str]) -> list[dict[s
         instance_types.append({preference: preference_config})
 
     return instance_types
+
+
+def generate_instance_type_centos_os_matrix(preferences: list[str]) -> list[dict[str, dict[str, Any]]]:
```
rnetser, line 246: duplicate code, please re-use `generate_instance_type_rhel_os_matrix`; remove `rhel` from the name and add an arg with the os type

rnetser (changes requested):

utilities/os_utils.py
```diff
@@ -213,30 +214,37 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
     return os_formatted_list
 
 
-def generate_instance_type_rhel_os_matrix(preferences: list[str]) -> list[dict[str, dict[str, Any]]]:
+def generate_instance_type_os_matrix(os_name: str, preferences: list[str]) -> list[dict[str, dict[str, Any]]]:
     """
-    Generate a list of dictionaries representing the instance type matrix for RHEL OS.
+    Generate a list of dictionaries representing the instance type matrix for a given OS type.
```
geetikakay, line 219: are we making it generic?

RoniKishner: yes

utilities/os_utils.py
```diff
@@ -213,30 +214,37 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
...
+
+    latest_os = max(preferences, key=extract_version)
     instance_types: list[dict[str, dict[str, Any]]] = []
 
     for preference in preferences:
         preference_config: dict[str, Any] = {
-            PREFERENCE_STR: preference.replace("-", "."),
-            DATA_SOURCE_NAME: preference.replace("-", ""),
+            PREFERENCE_STR: preference,
+            DATA_SOURCE_NAME: preference.replace(".", "" if os_name == "rhel" else "-"),
```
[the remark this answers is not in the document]
RoniKishner: I also think in windows we will need a new function, but this isn't in this PR scope.
what do you mean about the variants? we only check the data sources, we aren't checking all preferences, the function is being called only with similar list to: ["rhel.8", "rhel.9", "rhel.10"] / ["centos.stream9", "centos.stream10"]

utilities/os_utils.py
```diff
@@ -213,30 +214,37 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
...
     """
-    Generate a list of dictionaries representing the instance type matrix for RHEL OS.
+    Generate a list of dictionaries representing the instance type matrix for a given OS type.
 
     Each dictionary represents a specific instance type and its configuration.
 
     Args:
-        preferences (list[str]): A list of preferences for the instance types. Preference format is "rhel-<version>".
+        os_name (str): The name of the OS.
+        preferences (list[str]): A list of preferences for the instance types. Preference format is "<os>-<version>".
```
rnetser, line 225:
```suggestion
        preferences (list[str]): A list of preferences for the instance types. Preference format is "<os>.<version>".
```

rnetser (changes requested):

openshift-virtualization-qe-bot-3:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-1411 published

utilities/os_utils.py
geetikakay: so this is for non windows os ?

utilities/os_utils.py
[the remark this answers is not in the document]
geetikakay: this is a generic function it should handle all use cases, but it doesn't correctly handle valid preferences like rhel.8.dpdk and windows . The current logic turns that into rhel8dpdk, which doesn’t match any real datasource. So either we update the function to properly support variants and Windows mappings, or we make it clear that only base preferences like rhel.8 are supported(non windows).  we should log a warning when an unexpected or unsupported preference is passed in, so we don’t silently proceed and later figure out during tests

utilities/os_utils.py
RoniKishner: Yes, for now it is only for centos and rhel, as we create the windows test we need to see if we can use it or not

tests/global_config_x86_64.py
```diff
@@ -28,7 +31,12 @@
 
 fedora_os_matrix = generate_os_matrix_dict(os_name="fedora", supported_operating_systems=["fedora-42"])
 centos_os_matrix = generate_os_matrix_dict(os_name="centos", supported_operating_systems=["centos-stream-9"])
-instance_type_rhel_os_matrix = generate_instance_type_rhel_os_matrix(preferences=["rhel-8", "rhel-9", "rhel-10"])
+instance_type_rhel_os_matrix = generate_instance_type_os_matrix(
+    os_name="rhel", preferences=["rhel.8", "rhel.9", "rhel.10"]
```
geetikakay, line 35: do we want to save and reuse. called at multiple places.

tests/global_config_x86_64.py
```diff
@@ -28,7 +31,12 @@
...
+)
+instance_type_centos_os_matrix = generate_instance_type_os_matrix(
+    os_name="centos.stream", preferences=["centos.stream9", "centos.stream10"]
```
geetikakay, line 38: do we want to save and reuse. called at multiple places.

utilities/os_utils.py
RoniKishner: Yes, but there is no windows/fedora tests yet, so in this current state it works for all of the operating systems tests.
I don't think we should reflect on something we haven't implemented yet.

geetikakay: but this is part of utilities which means it is generic and it supports limited os. How will someone know that it's just for rhel and centos. somewhere we should reflect it either in function name or docstrings

geetikakay: if it is not generic, please move it locally in infrastructure

utilities/os_utils.py
```diff
@@ -218,30 +219,46 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
...
+        return -1
+
+    def _format_data_source_name(preference: str, os_name: str) -> str:
+        if os_name == "rhel":
+            if match := re.search(r"rhel\.(\d+)", preference):
+                return f"rhel{match.group(1)}"
+        elif os_name == "centos":
+            if match := re.search(r"centos\.stream(\d+)", preference):
+                return f"centos-stream{match.group(1)}"
+        return preference.replace(".", "-")
```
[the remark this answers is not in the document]
geetikakay: -- maybe like this..just an example. merge def _extract_version with it to avoid regex duplication 
def _format_data_source_name(preference: str):
    match os_name:
        case OS_FLAVOR_RHEL:
            if match := re.search(rf"{OS_FLAVOR_RHEL}\.(\d+)", preference):
                version = match.group(1)
                return f"{OS_FLAVOR_RHEL}{version}", int(version)
        case OS_FLAVOR_CENTOS:
.............
        case "ubuntu" | OS_FLAVOR_FEDORA | "sles":
            return os_name, None
        case _:
            raise ValueError(f"Unsupported OS name: {os_name}")

utilities/os_utils.py
```diff
@@ -218,30 +219,46 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
...
+
     """
-    latest_rhel = f"rhel-{max([preference.split('-')[1] for preference in preferences], key=int)}"
 
+    def _extract_version(pref: str) -> int:
+        if match := re.search(r"\d+", pref):
+            return int(match.group())
+        return -1
+
+    def _format_data_source_name(preference: str, os_name: str) -> str:
```
geetikakay, line 242: we can skip os_name here.it's already available.

tests/global_config_x86_64.py
RoniKishner: yes, but in a later PR - https://issues.redhat.com/browse/CNV-65113

utilities/os_utils.py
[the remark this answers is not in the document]
RoniKishner: Removed the default

tests/global_config_x86_64.py
geetikakay: we are using ["rhel.8", "rhel.9", "rhel.10"] twice in this PR so fixing here could be a better choice .
Maybe a have second commit for such cosmetic fixes in the same PR

utilities/os_utils.py
```diff
@@ -218,30 +219,46 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
...
     """
-    Generate a list of dictionaries representing the instance type matrix for RHEL OS.
+    Generate a list of dictionaries representing the instance type matrix for a given OS type.
 
     Each dictionary represents a specific instance type and its configuration.
 
     Args:
-        preferences (list[str]): A list of preferences for the instance types. Preference format is "rhel-<version>".
+        os_name (str): The name of the OS.
+        preferences (list[str]): A list of preferences for the instance types. Preference format is "<os>.<version>".
```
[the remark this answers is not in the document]
geetikakay: i see it is resolved ..what is the change?

utilities/os_utils.py
```diff
@@ -218,30 +219,46 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
...
     Returns:
         list[dict[str, dict[str, Any]]]: A list of dictionaries representing the instance type matrix.
+
     """
-    latest_rhel = f"rhel-{max([preference.split('-')[1] for preference in preferences], key=int)}"
 
+    def _extract_version(pref: str) -> int:
+        if match := re.search(r"\d+", pref):
+            return int(match.group())
+        return -1
```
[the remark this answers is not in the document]
geetikakay: I personally think None is better but depends what everyone else thinks.

tests/global_config_x86_64.py
```diff
@@ -28,7 +31,12 @@
 
 fedora_os_matrix = generate_os_matrix_dict(os_name="fedora", supported_operating_systems=["fedora-42"])
 centos_os_matrix = generate_os_matrix_dict(os_name="centos", supported_operating_systems=["centos-stream-9"])
-instance_type_rhel_os_matrix = generate_instance_type_rhel_os_matrix(preferences=["rhel-8", "rhel-9", "rhel-10"])
+instance_type_rhel_os_matrix = generate_linux_instance_type_os_matrix(
+    os_name="rhel", preferences=["rhel.8", "rhel.9", "rhel.10"]
+)
+instance_type_centos_os_matrix = generate_linux_instance_type_os_matrix(
+    os_name="centos.stream", preferences=["centos.stream9", "centos.stream10"]
```
geetikakay, line 38: they already exist in utilities/constants.py.

utilities/os_utils.py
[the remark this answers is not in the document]
RoniKishner: I changed the function naming
What do you suggest we change the doc string?

utilities/os_utils.py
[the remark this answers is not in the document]
RoniKishner: Removed `return -1` it so it won't return, which is the same as `return None`

utilities/os_utils.py
```diff
@@ -218,30 +223,43 @@ def generate_os_matrix_dict(os_name: str, supported_operating_systems: list[str]
...
+            return int(match.group())
+        return None
+
+    def _format_data_source_name(preference_name: str) -> str:
+        match = re.search(rf"{os_name}.*?(\d+)", preference_name)
+        if not match:
+            return preference_name
+        data_source_name = f"{os_name.replace('.', '-')}{match.group(1)}"
+        return data_source_name
+
```
geetikakay, line 251: Maybe :

def _extract_version(pref: str) -> int | None:
    match = re.search(r"\d+", pref)
    return int(match.group()) if match else None

def _format_data_source_name(preference_name: str) -> str:
    version = _extract_version(preference_name)
    if version is None:
        return preference_name
    return f"{os_name.replace('.', '-')}{version}"   -> that way we save regex duplication?

openshift-virtualization-qe-bot-2:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-1411.

PR #1447, opened 2025-07-13
Stabilize SR-IOV post-rebot test

yossisegev:
Occasionally some iterrations of this test fail, because of mismtach between pre- and post-reboot interfaces in irrelevant fields, e.g. `infoSource`. This fix limits the comparison only to the critical fields.

### DIFF
--- a/tests/network/sriov/test_sriov.py
+++ b/tests/network/sriov/test_sriov.py
@@ -16,6 +16,15 @@ LOGGER = logging.getLogger(__name__)
 pytestmark = [pytest.mark.special_infra, pytest.mark.sriov]
 
 
+def assert_similar_interfaces(interface1, interface2):
+    assert (
+        interface1.interfaceName == interface2.interfaceName
+        and interface1.ipAddress == interface2.ipAddress
+        and interface1.mac == interface2.mac
+        and interface1.name == interface2.name
+    ), f"Interfaces mismatch: {interface1} vs. {interface2}"
+
+
 class TestPingConnectivity:
     @pytest.mark.post_upgrade
     @pytest.mark.ipv4
@@ -81,7 +90,10 @@ class TestPingConnectivity:
         restarted_sriov_vm4,
     ):
         # Check only the second interface (SR-IOV interface).
-        assert restarted_sriov_vm4.vmi.interfaces[1] == vm4_interfaces[1]
+        assert_similar_interfaces(
+            interface1=restarted_sriov_vm4.vmi.interfaces[1],
+            interface2=vm4_interfaces[1],
+        )
 
 
 @pytest.mark.special_infra


tests/network/sriov/test_sriov.py
```diff
@@ -16,6 +16,15 @@
 pytestmark = [pytest.mark.special_infra, pytest.mark.sriov]
 
 
+def assert_similar_interfaces(interface1, interface2):
```
servolkov, line 19: I propose to create a place where assertions like this will be stored. Some kind of lib of assertions. Now there are plenty of places where we store lib/utils functions without strict logic how to choose a place for the new one (at least, I don't know this logic and can't calculate it looking at the dir structure). Same with assertions. I suggest, for example `libs/net/assertions/` as a common place for files with assertion functions. For sure, at some point we need to move all our available assertions there. I hope other teams will follow such flow as well, and we will have a transparent common place to store any assertion we need. Maybe this PR is a good one to create this place and move this func there?

yossisegev: Until now we usually treated the assertion functions as utility functions, and stored them in a `utils.py` module (you can see these modules on several packages along the project), so it might make sense to have a dedicated `assertions` package, but it might also make sense to go with storing these in utils.

Anatw: Thank you Sergei for bringing up this discussion!
The way we currently work is placing the functionality where it is needed. Currently this interface comparison is only needed only here. If other modules will need it, it should move to a higher location.
Having said that, I do think that when such functionalities move to a higher position, like utils for example, these higher level modules can benefit from some re-organization so that all the different functionalities are stored by sections, and such assert functionalities are easier to locate.

yossisegev:
> > Occasionally some iterrations of this test fail, because of mismtach between pre- and post-reboot interfaces
> > in irrelevant fields, e.g. `infoSource`. This fix limits the comparison only to the critical fields.
> 
> I'm unclear if this is expected or not. Could you please add here the exact information of the diff between the two (which caused the failure)?

Sure.
Here is an example:
```
00:15:58    - {'infoSource': 'domain, guest-agent, multus-status',
00:15:58    ?                        -------------
00:15:58    + {'infoSource': 'domain, multus-status',
00:15:58    -  'interfaceName': 'sriov1',
00:15:58    -  'ipAddress': '10.200.3.2',
00:15:58    -  'ipAddresses': ['10.200.3.2', 'fe80::b5ff:feb5:b511'],
00:15:58       'mac': '02:00:b5:b5:b5:11',
00:15:58       'name': 'sriov-test-network-vlan',
00:15:58       'podInterfaceName': 'pod5bbc590b508'}
```

EdDev:
> > > Occasionally some iterrations of this test fail, because of mismtach between pre- and post-reboot interfaces
> > > in irrelevant fields, e.g. `infoSource`. This fix limits the comparison only to the critical fields.
> > 
> > 
> > I'm unclear if this is expected or not. Could you please add here the exact information of the diff between the two (which caused the failure)?
> 
> Sure. Here is an example:
> 
> ```
> 00:15:58    - {'infoSource': 'domain, guest-agent, multus-status',
> 00:15:58    ?                        -------------
> 00:15:58    + {'infoSource': 'domain, multus-status',
> 00:15:58    -  'interfaceName': 'sriov1',
> 00:15:58    -  'ipAddress': '10.200.3.2',
> 00:15:58    -  'ipAddresses': ['10.200.3.2', 'fe80::b5ff:feb5:b511'],
> 00:15:58       'mac': '02:00:b5:b5:b5:11',
> 00:15:58       'name': 'sriov-test-network-vlan',
> 00:15:58       'podInterfaceName': 'pod5bbc590b508'}
> ```

Thank you.
Based on this, the test should what for the correct `infoSource` value before asserting.
(i.e. that the guest-agent is back alive and reports proper information)

yossisegev:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/1462 is more suitable to handle this flakiness.

PR #1453, opened 2025-07-14
remove un-used code in generate_instance_type_rhel_os_matrix

RoniKishner:
##### Short description:
Remove unused code in `generate_instance_type_rhel_os_matrix`  

##### More details:
 - Change test_rhel_os.py validation to use constant `U1_MEDIUM_STR` since it is the only value for preferences
 - Remove `DV_SIZE_STR` and `INSTANCE_TYPE_STR` since they are no longer used
 - Remove `if` condition so that we will always take the highest version available in the cluster

Reduce code complexity


openshift-virtualization-qe-bot-2:

PR #1464, opened 2025-07-15
[CNV-64119]Fix test_subscription_channels.test_cnv_subscription_channel

rlobillo:
##### Short description:
While testing stable releases, we might have both stable & candidate channels on the same packagemanifest version:

```
$ oc get packagemanifest kubevirt-hyperconverged \
  -n openshift-marketplace -o json | jq -r ".status.channels[] |\
  select(.currentCSVDesc.version==\"4.19.1\") | .name"
candidate
stable
```

The behaviour before this change was to take the first one, which was mismatching with the channel set on the subscription ("stable"). The test was failing in that case.

This change is checking that the subscription channel is present in the list of available channels in the packagemanifest, which will provide more stability to the test.

##### More details:
Locally tested on stable and candidate subscription channels.

Automation bug that is addressed thanks to this patch.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-64119


tests/install_upgrade_operators/csv/test_subscription_channels.py
```diff
@@ -11,5 +11,28 @@ def test_channels_in_manifest(kubevirt_package_manifest_channels):
...
+    cnv_subscription_scope_session, kubevirt_package_manifest_channels, cnv_current_version
+):
+    subscription_channel = cnv_subscription_scope_session.instance.spec.channel
+
+    assert subscription_channel, (
+        f"CNV subscription '{cnv_subscription_scope_session.name}' does not have a channel specified. "
+        f"Subscription spec.channel is: {subscription_channel!r}"
+    )
+
+    available_channels_for_version = [
```
rnetser, line 24: each verification should be a different test:
1. a test verifies one thing
2. when something fails, it is clear what functionality failed
3. when a test fails, the other verifications are still tested

rnetser (changes requested):

tests/install_upgrade_operators/csv/test_subscription_channels.py
rlobillo: I see. I will simplify the asserts then, no need to add new tests IMO.

rlobillo:
/verified
openshift-virtualization-tests-runner/2719

OhadRevah (approved):

openshift-virtualization-qe-bot-3:
Cherry-picked PR [CNV-64119]Fix test_subscription_channels.test_cnv_subscription_channel into cnv-4.19

PR #1510, opened 2025-07-18
[CNV-66035]increase must-gather timeout values to prevent collection failures

rlobillo:
##### Short description:
The all-images must-gather operations were timing out before completion, causing test failures and CI job to get stuck. This PR increases the timeouts.

The new timeout values are based on manual generation of the must-gather, which took around 35 minutes:

```
      Started:      Fri, 18 Jul 2025 10:29:05 -0400
      Finished:     Fri, 18 Jul 2025 11:04:30 -0400
```

##### More details:
Changes:
- Add a new parameter in utilities/must_gather.py to be able to set the
  command timeout from the tests. Default value remains set to 20
minutes.
- Set the new parameter in the test to increase the command timeout
  value to 40 minutes while generating  the "all-images" must-gather.

This fixes intermittent test failures related to must-gather timeouts that was in some cases provoking the overall test job to get stuck.

N/A
N/A
##### jira-ticket:
https://issues.redhat.com/browse/CNV-66035

## Summary by CodeRabbit

* **Chores**
  * Increased the timeout duration for certain operations to improve reliability during long-running tasks.
  * Added configurable timeout settings to enhance control over must-gather command execution.

### DIFF
--- a/tests/install_upgrade_operators/must_gather/conftest.py
+++ b/tests/install_upgrade_operators/must_gather/conftest.py
@@ -21,7 +21,7 @@ from tests.install_upgrade_operators.must_gather.utils import (
     get_must_gather_dir,
 )
 from tests.utils import create_vms
-from utilities.constants import LINUX_BRIDGE
+from utilities.constants import LINUX_BRIDGE, TIMEOUT_40MIN
 from utilities.exceptions import MissingResourceException
 from utilities.hco import ResourceEditorValidateHCOReconcile
 from utilities.infra import (
@@ -667,6 +667,7 @@ def collected_must_gather_all_images(
     output = run_must_gather(
         target_base_dir=must_gather_tmpdir_all_images,
         flag_names="all-images",
+        command_timeout=TIMEOUT_40MIN,
     )
     with open(os.path.join(must_gather_tmpdir_all_images, "output.log"), "w") as _file:
         _file.write(output)
--- a/utilities/must_gather.py
+++ b/utilities/must_gather.py
@@ -16,6 +16,7 @@ def run_must_gather(
     node_name: str = "",
     flag_names: str = "",
     timeout: str = f"{TIMEOUT_15MIN}s",
+    command_timeout: int = TIMEOUT_20MIN,
     since: str | None = None,
 ) -> str:
     """
@@ -47,7 +48,7 @@ def run_must_gather(
         base_command += f" --node-name={node_name}"
     if since:
         base_command += f" --since={since}"
-    if timeout:
+    if timeout:  # Only applies to gathering and not copying stage - https://issues.redhat.com/browse/OCPBUGS-59774
         base_command += f" --timeout={timeout}"
     if script_name:
         base_command += f" -- {script_name}"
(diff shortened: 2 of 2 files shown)


tests/install_upgrade_operators/must_gather/conftest.py
```diff
@@ -667,6 +667,7 @@ def collected_must_gather_all_images(
     output = run_must_gather(
         target_base_dir=must_gather_tmpdir_all_images,
         flag_names="all-images",
+        timeout=f"{TIMEOUT_50MIN}s",
```
rnetser, line 670: i think you should discuss in the team how to address this collection; waiting up to 50 minutes each time for must-gather sounds too much

utilities/must_gather.py
```diff
@@ -58,7 +58,7 @@ def run_must_gather(
     did_succeed, output, error = run_command(
         command=shlex.split(base_command),
         check=False,
-        timeout=TIMEOUT_20MIN,
+        timeout=TIMEOUT_60MIN,
```
rnetser, line 61: this affects all calls to `run_must_gather`
1. there's a bug in the existing code which does not use the passed arg `timeout`
2. please fix relevant calls to `timeout` and pass the required timeout.

rnetser (changes requested):

utilities/must_gather.py
rlobillo: @rnetser thanks for your comment! As per my testing, the bug seems to be on the oc binary. The `--timeout` seems to be ignored.  The logs show that the param is correctly passed: 

```
subprocess.TimeoutExpired: Command '['oc', 'adm', 'must-gather', '--dest-dir=/data/tests-collected-info/must_gather_all_images', '--timeout=900s', '--all-images']' timed out after 1200 seconds
```
Are you aware of any bug in the oc binary? I took a look and didn't find any.

tests/install_upgrade_operators/must_gather/conftest.py
rlobillo: That timeout would be the maximum: The idea is to leave room to the must-gather to be generated. As soon as it is finished, it will return and that timeout will be not applied.

I run several must-gathers and it's taking between 18-25 minutes. I am going to reduce it to 30minutes.

utilities/must_gather.py
rlobillo: Actually this value is not the one defined on `--timeout`, but the time that python will wait for the command to finish. This is the timeout that is taking place right now.

I've done some must-gather generations in a row and I observed times between 18 and 25 times. Therefore, I am updating this value from 20 minutes to 30 minutes, so the `oc adm must-gather` can finish before python terminates the command abruptly.

utilities/must_gather.py
```diff
@@ -16,6 +16,7 @@ def run_must_gather(
     node_name: str = "",
     flag_names: str = "",
     timeout: str = f"{TIMEOUT_15MIN}s",
+    command_timeout: str = f"{TIMEOUT_20MIN}",
```
rnetser, line 19: run_command timeout is an int, so this should also be an int

rnetser (changes requested):

utilities/must_gather.py
rlobillo: my bad. Thanks for noticing!

dshchedr (approved):

OhadRevah (approved):

rlobillo:
@rnetser Please help to get this one merge too! Thank you again

openshift-virtualization-qe-bot-3:

PR #1519, opened 2025-07-21
T2 node test adopted for s390x arch

chandramerla:
##### Short description:
Updated Node Tier2 tests after adopting to work for s390x

##### More details:
Added inline comments as necessary in the code.

To run node tier2 tests in s390x in the pipeline

Added pytest.mark.x86_64 markers for all the tests which I triaged to differentiate between:
- Newly added test in future 
vs 
- test which is already triaged for adopting for s390x arch, but as not applicable pytest.mark.s390x marker is not added.

Otherwise there is ambiguity if a test is checked for adopting to s390x or not. If there is one of the arch specific marker, then I'm assuming that that test is checked for adopting to s390x.

##### jira-ticket:

NONE

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -287,7 +287,7 @@ def update_cluster_cpu_model(admin_client, hco_namespace, hco_resource, cpu_mode
             hco_namespace=hco_namespace,
             path=["cpuModel"],
             value=cpu_model,
-            timeout=60,
+            timeout=120,
         )
         yield
 
--- a/utilities/storage.py
+++ b/utilities/storage.py
@@ -33,7 +33,6 @@ from utilities.constants import (
     HOTPLUG_DISK_SERIAL,
     HPP_POOL,
     OS_FLAVOR_CIRROS,
-    OS_FLAVOR_WINDOWS,
     POD_CONTAINER_SPEC,
     TIMEOUT_1SEC,
     TIMEOUT_2MIN,
@@ -44,7 +43,6 @@ from utilities.constants import (
     TIMEOUT_10MIN,
     TIMEOUT_10SEC,
     TIMEOUT_20SEC,
-    TIMEOUT_30MIN,
     TIMEOUT_30SEC,
     TIMEOUT_60MIN,
     Images,
@@ -287,7 +285,7 @@ def data_volume(
                     # It will be in a status 'PendingPopulation' (for csi storage)
                     dv.wait_for_status(status="PendingPopulation", timeout=TIMEOUT_10SEC)
                 else:
-                    dv.wait_for_dv_success(timeout=TIMEOUT_60MIN if OS_FLAVOR_WINDOWS in image else TIMEOUT_30MIN)
+                    dv.wait_for_dv_success(timeout=TIMEOUT_60MIN)
         yield dv
 
 
--- a/utilities/virt.py
+++ b/utilities/virt.py
@@ -77,6 +77,7 @@ from utilities.constants import (
     TIMEOUT_12MIN,
     TIMEOUT_25MIN,
     TIMEOUT_30MIN,
+    TIMEOUT_60MIN,
     VIRT_LAUNCHER,
     VIRTCTL,
     Images,
@@ -1630,7 +1631,7 @@ def assert_vm_not_error_status(vm: VirtualMachineForTests, timeout: int = TIMEOU
 
 def wait_for_running_vm(
     vm: VirtualMachineForTests,
-    wait_until_running_timeout: int = TIMEOUT_4MIN,
+    wait_until_running_timeout: int = TIMEOUT_8MIN,
     wait_for_interfaces: bool = True,
     check_ssh_connectivity: bool = True,
     ssh_timeout: int = TIMEOUT_2MIN,
@@ -1668,7 +1669,7 @@ def running_vm(
     check_ssh_connectivity=True,
     ssh_timeout=TIMEOUT_2MIN,
     wait_for_cloud_init=False,
-    dv_wait_timeout=TIMEOUT_30MIN,
+    dv_wait_timeout=TIMEOUT_60MIN,
 ):
     """
     Wait for the VMI to be in Running state.
@@ -1747,7 +1748,7 @@ def wait_for_cloud_init_complete(vm, timeout=TIMEOUT_4MIN):
 def migrate_vm_and_verify(
     vm: VirtualMachineForTests | BaseVirtualMachine,
     client: DynamicClient | None = None,
-    timeout: int = TIMEOUT_12MIN,
+    timeout: int = TIMEOUT_25MIN,
     wait_for_interfaces: bool = True,
     check_ssh_connectivity: bool = False,
     wait_for_migration_success: bool = True,
--- a/tests/virt/node/node_labeller/cpu_features/test_node_feature_discovery.py
+++ b/tests/virt/node/node_labeller/cpu_features/test_node_feature_discovery.py
@@ -56,7 +56,7 @@ def updated_kubevirt_cpus(
             hco_namespace=hco_namespace,
             path=[OBSOLETE_CPU, cluster_common_node_cpu],
             value=True,
-            timeout=60,
+            timeout=120,
         )
         wait_for_kv_stabilize(admin_client=admin_client, hco_namespace=hco_namespace)
         yield


tests/conftest.py
```diff
@@ -2826,7 +2827,13 @@ def cluster_modern_cpu_model_scope_class(
...
 def machine_type_from_kubevirt_config(kubevirt_config_scope_module, nodes_cpu_architecture):
     """Extract machine type default from kubevirt CR."""
-    return kubevirt_config_scope_module["architectureConfiguration"][nodes_cpu_architecture]["machineType"]
+    # Workaround for s390x (https://github.com/kubevirt/kubevirt/issues/14953), as machine type missing in config and
+    # hardcoded to s390_ccw_virtio in kubevirt code.
+    if nodes_cpu_architecture == S390X:
+        mc_type = "s390-ccw-virtio"
+    else:
+        mc_type = kubevirt_config_scope_module["architectureConfiguration"][nodes_cpu_architecture]["machineType"]
+    return mc_type
```
vsibirsk, line 2836:
```suggestion
    if nodes_cpu_architecture == S390X:
        return "s390-ccw-virtio"
    else:
        return kubevirt_config_scope_module["architectureConfiguration"][nodes_cpu_architecture]["machineType"]
```

tests/utils.py
```diff
@@ -287,7 +287,7 @@ def update_cluster_cpu_model(admin_client, hco_namespace, hco_resource, cpu_mode
             hco_namespace=hco_namespace,
             path=["cpuModel"],
             value=cpu_model,
-            timeout=30,
+            timeout=120,
```
vsibirsk, line 290: why do we need to increase timeout here?

tests/virt/node/conftest.py
```diff
@@ -64,7 +66,8 @@ def hotplugged_vm(
         client=unprivileged_client,
         data_source=golden_image_data_source_scope_class,
         cpu_max_sockets=EIGHT_CPU_SOCKETS,
-        memory_max_guest=TEN_GI_MEMORY,
+        # s390x doesn't support maxGuest as it doesn't support hotplug
+        memory_max_guest=TEN_GI_MEMORY if nodes_cpu_architecture != S390X else None,
```
vsibirsk, line 70: out of curiosity, what happens if the field is still set? does it ruins something in tests?

tests/virt/node/cpu_sockets_threads/test_cpu_support_sockets_threads.py
```diff
@@ -45,22 +45,32 @@ def vm_with_cpu_support(request, namespace, unprivileged_client):
     [
         pytest.param(
             {"sockets": 2, "cores": 2, "threads": 2},
-            marks=(pytest.mark.polarion("CNV-2820"), pytest.mark.gating()),
+            marks=(pytest.mark.polarion("CNV-2820"), pytest.mark.gating(), pytest.mark.x86_64()),
```
vsibirsk, line 48: i don't think you need `()` for marks other then `polarion`

tests/virt/node/general/test_custom_selinux_policy.py
```diff
@@ -12,4 +13,4 @@ def test_customselinuxpolicy(workers_utility_pods, schedulable_nodes):
         out = pod_exec.exec(command="sudo semodule -l")
         if "virt_launcher" in out:
             nodes.append(node.name)
-    assert not nodes, f"node: {nodes} still have virt-launcher policies."
+    assert not nodes, f"node: {nodes} still have virt_launcher policies."
```
vsibirsk, line 16: why this change?

tests/virt/node/migration_and_maintenance/test_post_copy_migration.py
```diff
@@ -142,6 +143,10 @@ def test_hotplug_cpu(self, hotplugged_sockets_memory_guest, hotplugged_vm, vm_ba
     )
     @pytest.mark.dependency(depends=[f"{TESTS_CLASS_NAME}::hotplug_cpu"])
     @pytest.mark.polarion("CNV-11424")
-    def test_hotplug_memory(self, hotplugged_sockets_memory_guest, hotplugged_vm, vm_background_process_id):
+    def test_hotplug_memory(self, request, hotplugged_sockets_memory_guest, hotplugged_vm, vm_background_process_id):
+        # Dynamically skip only this test when class is run with s390x params as memory hotplug is not supported
+        if request.node.get_closest_marker("s390x"):
+            pytest.skip("Skipping test_hotplug_memory for s390x")
```
vsibirsk, line 149: please move this to separate fixture `skip_on_s390x_architecture` and add with `@pytest.mark.usefixtures("skip_on_s390x_architecture")`

tests/virt/node/migration_and_maintenance/test_vm_disk_load_with_migration.py
```diff
@@ -83,7 +83,21 @@ def get_disk_usage(ssh_exec):
...
+                "image": FEDORA_LATEST.get("image_path"),
+                "storage_class": py_config["default_storage_class"],
+                "dv_size": FEDORA_LATEST.get("dv_size"),
+            },
+            {
+                "vm_name": "fedora-load-vm",
+                "template_labels": FEDORA_LATEST_LABELS,
+                "cpu_threads": 1,
+            },
+            marks=[pytest.mark.s390x()],
```
vsibirsk, line 100: why do you need separate param here?

utilities/storage.py
```diff
@@ -287,7 +285,7 @@ def data_volume(
                     # It will be in a status 'PendingPopulation' (for csi storage)
                     dv.wait_for_status(status="PendingPopulation", timeout=TIMEOUT_10SEC)
                 else:
-                    dv.wait_for_dv_success(timeout=TIMEOUT_60MIN if OS_FLAVOR_WINDOWS in image else TIMEOUT_30MIN)
+                    dv.wait_for_dv_success(timeout=TIMEOUT_60MIN)
```
vsibirsk, line 288: why remove 30min for non-windows DVs?

utilities/virt.py
```diff
@@ -1630,7 +1631,7 @@ def assert_vm_not_error_status(vm: VirtualMachineForTests, timeout: int = TIMEOU
 
 def wait_for_running_vm(
     vm: VirtualMachineForTests,
-    wait_until_running_timeout: int = TIMEOUT_4MIN,
+    wait_until_running_timeout: int = TIMEOUT_8MIN,
```
vsibirsk, line 1634: why do we need to x2 all default timeout values here?

vsibirsk (changes requested):

chandramerla (commented):
@vsibirsk Thanks a lot for your review.

I've double timeouts wherever I was seeing TimeOutExpired error. After timeouts have increased, atleast in a baremetal s390x cluster I'm using all s390x marker tests are passing. I will cross check on the timeout once on kvm-vm-as-node cluster as well and update accordingly.

utilities/storage.py
vsibirsk: i guess you're hitting an artifactory slow image download issue (we're currently facing it on all our envs)

you can use this for testing purposes as a w/a but it shouldn't end up in the PR

tests/virt/node/general/test_machinetype.py
```diff
@@ -82,22 +84,39 @@ def migrated_vm(vm, machine_type_from_kubevirt_config):
...
-            marks=pytest.mark.polarion("CNV-3311"),
-        )
+            MachineTypesNames.pc_q35_rhel7_6,
+            marks=[pytest.mark.polarion("CNV-3311"), pytest.mark.x86_64()],
+        ),
+        pytest.param(
+            {"vm_name": "s390-ccw-virtio", "machine_type": MachineTypesNames.s390_ccw_virtio},
+            MachineTypesNames.s390_ccw_virtio,
+            MachineTypesNames.s390_ccw_virtio_rhel9_6,
+            marks=[pytest.mark.s390x()],
```
rnetser, line 113: architecture should not be used for test parameterization, if some config is different, please set it where needed.

tests/virt/node/general/test_machinetype.py
```diff
@@ -113,10 +132,21 @@ def test_pc_q35_vm_machine_type(vm, expected):
...
+    # s390x: machine type missing in config (GH#14953). Same as above.
+    if machine_type_from_kubevirt_config == MachineTypesNames.s390_ccw_virtio:
+        expected_libvirt_machine_type = MachineTypesNames.s390_ccw_virtio_rhel9_6
+    else:
+        expected_libvirt_machine_type = machine_type_from_kubevirt_config
+    validate_machine_type(
+        vm=vm,
+        expected_machine_type=machine_type_from_kubevirt_config,
+        expected_libvirt_machine_type=expected_libvirt_machine_type,
+    )
```
rnetser, line 149: duplicate code, please extract to a function and re-use

tests/virt/node/migration_and_maintenance/test_post_copy_migration.py
rnetser: please mark all `s390x` applicable tests with a marker.
we should not have such a skip in the code

tests/virt/node/migration_and_maintenance/test_vm_disk_load_with_migration.py
rnetser: we should not parametrize on architecture. please set required config where needed

tests/virt/utils.py
```diff
@@ -362,17 +362,21 @@ def wait_for_virt_launcher_pod(vmi):
         raise
 
 
-def validate_machine_type(vm, expected_machine_type):
+def validate_machine_type(
+    vm, expected_machine_type, expected_libvirt_machine_type=None
+):  # Incase of s390x machine_type in VM/VMI are different than one in libvirt xml
```
rnetser, line 367: please use google docstring formatting

utilities/virt.py
rnetser: we need to understand why you hit a timeout; such a change is not acceptable as it adds potentially  4 minutes for each test that calls this function.

rnetser (changes requested):

utilities/storage.py
chandramerla: @vsibirsk Where can I watch on the "artifactory slow image download issue" to find when it is fixed? For me I'm seeing all these TimeoutExpired error only since last week.

tests/virt/node/cpu_sockets_threads/test_cpu_support_sockets_threads.py
```diff
@@ -45,22 +45,32 @@ def vm_with_cpu_support(request, namespace, unprivileged_client):
...
             id="case1: 2 cores, 2 threads, 2 sockets",
         ),
         pytest.param(
             {"sockets": None, "cores": 1, "threads": 2},
-            marks=(pytest.mark.polarion("CNV-2823")),
+            marks=(pytest.mark.polarion("CNV-2823"), pytest.mark.x86_64()),
```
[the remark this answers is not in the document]
chandramerla: @rnetser By adding atleast one architecture marker for all the tests which are triaged for s390x compatibility, I'm trying to avoid the ambiguity between tests which are not triaged yet (incl. a new test which isn't triaged for s390x compatibility) with that of triaged but not compatible with s390x.

I've one question here - In future, will the **person who adds any new test also ensures whether it is s390x compatible or not** and adds the s390x marker if applicable? If yes, then I think we don't need to find a way to identify triage status of a test, as not having s390x label for a new test means it is triaged for s390x compatibility and found not compatible.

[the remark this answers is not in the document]
chandramerla: ok. I will remove the markers

tests/virt/node/general/test_machinetype.py
chandramerla: Moved it into validate_machine_type() itself

tests/virt/node/migration_and_maintenance/test_vm_disk_load_with_migration.py
chandramerla: I've updated vm_instance_from_template() to ignore the value of cpu_threads in case of s390x and set it as 1 otherwise to use the passed value.

tests/virt/utils.py
chandramerla: As I've removed arg expected_libvirt_machine_type in the code, the comment is no more relavent.

PR #1533, opened 2025-07-22
Skip tests for storage class volume mode as block

stesrn:
Non-relevant storage tests are skipped when storage class with volume mode block is available.

##### jira-ticket:
"NONE"


openshift-virtualization-qe-bot-2:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-1533 published

tests/storage/cdi_import/test_import_http.py
```diff
@@ -360,6 +360,7 @@ def test_successful_import_basic_auth(
     indirect=True,
 )
 def test_wrong_content_type(
+    skip_test_if_no_filesystem_sc,
```
jpeimer, line 363: Every kubernetes storage class supports Filesystem-RWO. If some tests need specifically FS, we need to create a DV/PVC by explicitly passing `volume_mode` FS

jpeimer (changes requested):

tests/storage/cdi_import/test_import_http.py
rnetser: no programatic skips should be added

PR #1537, opened 2025-07-22
 [service mesh] fix: correct service mesh traffic blocking test

Anatw:
The test was previously failing due to DNS resolution failure (curl: (6) Could not resolve host) rather than service mesh (SM) policy blocking. This meant the test passed but wasn't actually validating SM functionality.

Changes:
- Use full service FQDN for cross-namespace access to ensure DNS resolution succeeds.
- Check for specific service mesh blocking error (curl: (56) Recv failure) instead of generic assertion.
- Add expected_output parameter to assert_authentication_request for flexible error checking.

Now the test properly validates that service mesh peer authentication blocks outside mesh traffic rather than passing due to unrelated DNS failures.

##### More details:
Fixes this automation bug: https://issues.redhat.com/browse/CNV-65999

## Summary by CodeRabbit

* **Tests**
  * Refined service-mesh tests for traffic originating outside the mesh to use the full service DNS name and validate an expected network failure response.
  * Removed an internal test helper and updated assertions to reduce flakiness and improve accuracy of out-of-mesh authentication/connection checks, increasing test reliability without affecting runtime behavior.

### DIFF
--- a/tests/network/service_mesh/test_service_mesh.py
+++ b/tests/network/service_mesh/test_service_mesh.py
@@ -95,7 +95,6 @@ class TestSMPeerAuthentication:
         httpbin_service_service_mesh,
         outside_mesh_console_ready_vm,
     ):
-        expected_failure_output = "curl: (56) Recv failure"
         # We must specify the full service DNS name since the VM is outside the mesh in a different namespace
         # Format: http://<service_name>.<service_namespace>.svc.cluster.local
         assert_authentication_request(
@@ -103,7 +102,7 @@ class TestSMPeerAuthentication:
             service_app_name=(
                 f"{httpbin_service_service_mesh.app_name}.{httpbin_service_service_mesh.namespace}.svc.cluster.local"
             ),
-            expected_output=expected_failure_output,
+            expected_output="curl: (56) Recv failure",
         )
 
     @pytest.mark.polarion("CNV-7128")
--- a/tests/network/service_mesh/test_service_mesh.py
+++ b/tests/network/service_mesh/test_service_mesh.py
@@ -2,7 +2,6 @@ import pytest
 
 from tests.network.service_mesh.constants import AUTH_COMMAND, EXPECTED_MESH_SUCCESS_OUTPUT
 from tests.network.service_mesh.utils import (
-    assert_authentication_request,
     assert_traffic_management_request,
     inbound_request,
     run_console_command,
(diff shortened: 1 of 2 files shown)


tests/network/utils.py
```diff
@@ -301,9 +301,8 @@ def assert_service_mesh_request(expected_output, request_response):
...
 
 
-def assert_authentication_request(vm, service_app_name):
-    # Envoy proxy IP
-    expected_output = "127.0.0.6"
+def assert_authentication_request(vm, service_app_name, expected_output="127.0.0.6"):
+    # 127.0.0.6 is the default IP for the istio envoy proxy.
     request_response = authentication_request(
         vm=vm,
         service=service_app_name,
```
rnetser, line 308: cannot add a comment on line https://github.com/RedHatQE/openshift-virtualization-tests/pull/1537/files#diff-6daf2198793af420d302c77791bd16f97231df3b9178bdc41b7dd5ae8c48c8a2R311
but in another PR it will be good to drop `assert_service_mesh_request` and create a proper exception

rnetser (changes requested):

tests/network/service_mesh/test_service_mesh.py
```diff
@@ -89,11 +89,16 @@ def test_outside_mesh_traffic_blocked(
         httpbin_service_service_mesh,
         outside_mesh_console_ready_vm,
     ):
-        with pytest.raises(AssertionError):
-            assert_authentication_request(
-                vm=outside_mesh_vm_fedora_with_service_mesh_annotation,
-                service_app_name=httpbin_service_service_mesh.app_name,
-            )
+        expected_failure_output = "curl: (56) Recv failure"
```
EdDev, line 92: Why do you need a variable?

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/pull/1537/commits/71d817ccedf697c1d242b53ebcd4a6ca950ad714):
    Fix according to CR
    
    - Remove variable expected_failure_output and use directly.
    - Add docstring to assert_authentication_request.

tests/network/utils.py
servolkov: I'd like to add my 5 cents. The function is called `assert_authentication_request`, I would expect direct `assert` inside. If we decide that custom exceptions are preferred for clarity (which can make sense), we should align the naming - e.g., `validate_`. Otherwise, it creates inconsistency and can be misleading.

Would be great to define naming conventions (as well as location) for functions like `assert_`, `check_`, `validate_`, `verify_` , now we have plenty of them, but I don't figure out clearly the difference.

openshift-virtualization-qe-bot-2:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-1537 published

yossisegev (approved):
Thank you

Anatw:
/verified

net-awax-420-sm
Cluster version is 4.20.0-ec.4
kubevirt-hyperconverged-operator.v4.20.0
iib: 1016173
openshift-virtualization-tests-runner/2878

PR #1600, opened 2025-07-28
Give the console password prompt grace time to appear

yossisegev:
## Summary by CodeRabbit

* **Chores**
  * Updated the console connection process to use a defined timeout when waiting for the password prompt, improving reliability during login.

### DIFF
--- a/utilities/console.py
+++ b/utilities/console.py
@@ -5,6 +5,7 @@ import pexpect
 from timeout_sampler import TimeoutSampler
 
 from utilities.constants import (
+    TIMEOUT_1MIN,
     TIMEOUT_5MIN,
     VIRTCTL,
 )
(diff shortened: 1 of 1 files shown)


utilities/console.py
```diff
@@ -55,7 +56,7 @@ def _connect(self):
             LOGGER.info(f"{self.vm.name}: Using username {self.username}")
             self.child.sendline(self.username)
             if self.password:
-                self.child.expect("Password:")
+                self.child.expect("Password:", timeout=TIMEOUT_1MIN)
```
EdDev, line 59: I do not understand this timeout.
At line 55 above, the login prompt is waited for 5min (a bit wild and not reasonable by itself).
Then, for the password we need another minute? I.e. between posting the username it may take 1min to get the password prompt? 

IMO the current 30 sec is enough and cannot explain any flake.

Based on the documentation [1]:

> When the keyword argument timeout is specified as a number, (default: 30), then [TIMEOUT](https://pexpect.readthedocs.io/en/stable/api/pexpect.html?highlight=timeout#pexpect.TIMEOUT) will be raised after the value specified has elapsed, in seconds, for any of the [expect()](https://pexpect.readthedocs.io/en/stable/api/pexpect.html?highlight=timeout#pexpect.spawn.expect) family of method calls.

[1] https://pexpect.readthedocs.io/en/stable/api/pexpect.html?highlight=timeout#spawn-class

PR #1669, opened 2025-08-04
  Virt: Add test cases to confirm that guest agent online when reboot the vm

SiboWang1997:
##### Short description:
The Closed Loop Automation, we need to confirm that guest agent online when reboot the vm
##### jira-ticket:

https://issues.redhat.com/browse/CNV-65111

## Summary by CodeRabbit

* **Tests**
  * Added cross-OS tests verifying guest-agent info persists after an in-guest reboot for CentOS, Fedora, RHEL, and Windows templates; tests reuse shared validation helpers and updated test dependency naming for clarity.

* **Refactor**
  * Moved in-guest reboot, remote command execution, and post-reboot guest-agent validation into shared utilities.
  * Standardized OS constants and OS-aware timeouts for more stable lifecycle tests.

### DIFF
--- a/tests/virt/cluster/common_templates/centos/test_centos_os_support.py
+++ b/tests/virt/cluster/common_templates/centos/test_centos_os_support.py
@@ -14,6 +14,7 @@ from tests.virt.cluster.common_templates.utils import (
     vm_os_version,
 )
 from utilities import console
+from utilities.constants import LINUX_STR
 from utilities.infra import validate_os_info_vmi_vs_linux_os
 from utilities.virt import (
     check_vm_xml_smbios,
@@ -21,6 +22,7 @@ from utilities.virt import (
     running_vm,
     validate_libvirt_persistent_domain,
     validate_pause_optional_migrate_unpause_linux_vm,
+    validate_virtctl_guest_agent_after_guest_reboot,
     validate_virtctl_guest_agent_data_over_time,
     wait_for_console,
 )
@@ -78,7 +80,9 @@ class TestCommonTemplatesCentos:
             tcp_timeout=120
         ), "Failed to login via SSH"
 
-    @pytest.mark.dependency(depends=[f"{TESTS_CLASS_NAME}::vm_expose_ssh"])
+    @pytest.mark.dependency(
+        name=f"{TESTS_CLASS_NAME}::vmi_guest_agent_info", depends=[f"{TESTS_CLASS_NAME}::vm_expose_ssh"]
+    )
     @pytest.mark.polarion("CNV-5346")
     def test_vmi_guest_agent_info(self, matrix_centos_os_vm_from_template):
         """Test Guest OS agent info."""
@@ -138,6 +142,11 @@ class TestCommonTemplatesCentos:
         assert validate_virtctl_guest_agent_data_over_time(vm=matrix_centos_os_vm_from_template), (
             "Guest agent stopped responding"
         )
+
+    @pytest.mark.polarion("CNV-12222")
+    @pytest.mark.dependency(depends=[f"{TESTS_CLASS_NAME}::vmi_guest_agent_info"])
+    def test_vmi_guest_agent_info_after_guest_reboot(self, matrix_centos_os_vm_from_template):
+        validate_virtctl_guest_agent_after_guest_reboot(vm=matrix_centos_os_vm_from_template, os_type=LINUX_STR)
 
     @pytest.mark.dependency(depends=[f"{TESTS_CLASS_NAME}::create_vm"])
     @pytest.mark.polarion("CNV-5351")
--- a/tests/virt/cluster/common_templates/windows/test_windows_os_support.py
+++ b/tests/virt/cluster/common_templates/windows/test_windows_os_support.py
@@ -16,6 +16,7 @@ from tests.virt.cluster.common_templates.utils import (
     validate_user_info_virtctl_vs_windows_os,
 )
 from tests.virt.utils import validate_pause_optional_migrate_unpause_windows_vm
+from utilities.constants import OS_FLAVOR_WINDOWS
 from utilities.ssp import validate_os_info_vmi_vs_windows_os
 from utilities.virt import (
     assert_vm_xml_efi,
@@ -23,6 +24,7 @@ from utilities.virt import (
     migrate_vm_and_verify,
     running_vm,
     validate_libvirt_persistent_domain,
+    validate_virtctl_guest_agent_after_guest_reboot,
     validate_virtctl_guest_agent_data_over_time,
 )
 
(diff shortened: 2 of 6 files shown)


tests/virt/cluster/common_templates/windows/test_windows_os_support.py
```diff
@@ -193,6 +196,17 @@ def test_verify_virtctl_guest_agent_data_after_migrate(
             vm=golden_image_vm_object_from_template_multi_windows_os_multi_storage_scope_class
         ), "Guest agent stopped responding"
 
+    @pytest.mark.polarion("CNV-12220")
+    @pytest.mark.dependency(depends=[f"{TESTS_CLASS_NAME}::migrate_vm_and_verify"])
+    def test_vmi_guest_agent_info_after_guest_reboot(
+        self,
+        golden_image_vm_object_from_template_multi_windows_os_multi_storage_scope_class,
+    ):
+        assert validate_virtctl_guest_agent_after_guest_reboot(
```
akri3i, line 205: I don't think we need assert here, 
we can directly call the function  validate_virtctl_guest_agent_after_restart(...)
if guest agent is not up wait_for_vm_interfaces or wait_for_ssh_connectivity should return Timeout error and fail the test

utilities/virt.py
```diff
@@ -2626,3 +2630,64 @@ def username_password_from_cloud_init(vm_volumes: list[dict[str, Any]]) -> tuple
...
+
+    # wait for the VM to come back up
+    wait_for_vm_interfaces(vmi=vm.vmi)
+    wait_for_ssh_connectivity(
+        vm=vm,
+        timeout=TIMEOUT_30MIN if os_type == WIN else TIMEOUT_5MIN,
+        tcp_timeout=TIMEOUT_2MIN,
+    )
+
+    return validate_virtctl_guest_agent_data_over_time(vm=vm)
```
akri3i, line 2646: I don't think we need to return anything here

utilities/virt.py
```diff
@@ -2632,3 +2636,64 @@ def username_password_from_cloud_init(vm_volumes: list[dict[str, Any]]) -> tuple
...
+def validate_virtctl_guest_agent_after_guest_reboot(vm: VirtualMachineForTests, os_type: str) -> bool:
+    guest_reboot(vm=vm, os_type=os_type)
+
+    # wait for the VM to come back up
+    wait_for_vm_interfaces(vmi=vm.vmi)
+    wait_for_ssh_connectivity(
+        vm=vm,
+        timeout=TIMEOUT_30MIN if os_type == WIN else TIMEOUT_5MIN,
+        tcp_timeout=TIMEOUT_2MIN,
+    )
```
dshchedr, line 2650: you can use `wait_for_running_vm` which includes both verifications

tests/virt/cluster/common_templates/centos/test_centos_os_support.py
```diff
@@ -183,6 +185,16 @@ def test_verify_virtctl_guest_agent_data_after_migrate(
             vm=golden_image_vm_object_from_template_multi_centos_multi_storage_scope_class
         ), "Guest agent stopped responding"
 
+    @pytest.mark.polarion("CNV-12222")
+    @pytest.mark.dependency(depends=[f"{TESTS_CLASS_NAME}::migrate_vm_and_verify"])
```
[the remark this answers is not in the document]
dshchedr: +1, don't think we need to depend on migration test here

utilities/virt.py
```diff
@@ -2632,3 +2636,64 @@ def username_password_from_cloud_init(vm_volumes: list[dict[str, Any]]) -> tuple
             return _match["user"], _match["password"]
 
     return "", ""
+
+
+def validate_virtctl_guest_agent_after_guest_reboot(vm: VirtualMachineForTests, os_type: str) -> bool:
```
dshchedr, line 2641: if it is used only inside 'tests/virt/cluster' - you can keep it there, no need to move to common place

also, this function contains too many actions (reboot, wait when running, validate ga), please split it and rename to something like "reboot_guest_and_wait_for_running", it should include only steps `guest_reboot` and `wait_for_running_vm`

and it should be used as a fixture in all tests

in the test body you should call old function: `validate_virtctl_guest_agent_data_over_time`

dshchedr (changes requested):

utilities/virt.py
SiboWang1997: done

dshchedr (changes requested):

tests/virt/cluster/common_templates/centos/test_centos_os_support.py
```diff
@@ -139,6 +140,17 @@ def test_verify_virtctl_guest_agent_data_after_migrate(self, matrix_centos_os_vm
...
         )
 
+    @pytest.mark.polarion("CNV-12222")
+    @pytest.mark.dependency(depends=[f"{TESTS_CLASS_NAME}::start_vm"])
+    def test_vmi_guest_agent_info_after_guest_reboot(
+        self, matrix_centos_os_vm_from_template, reboot_guest_and_wait_for_running
+    ):
+        vm = matrix_centos_os_vm_from_template
+        reboot_guest_and_wait_for_running(vm=vm, os_type=LINUX_STR)
+        assert validate_virtctl_guest_agent_data_over_time(vm=matrix_centos_os_vm_from_template), (
```
akri3i, line 150: why not pass vm=vm if you plan on saving it?,
 validate_virtctl_guest_agent_data_over_time(vm=vm)

akri3i (changes requested):

tests/virt/cluster/common_templates/centos/test_centos_os_support.py
SiboWang1997: Thanks for pointing out, yes your are right, no need to save vm, just use matrix_centos_os_vm_from_template directly, I’ll push the revised patch soon

dshchedr (dismissed):

vsibirsk (approved):

utilities/virt.py
```diff
@@ -2638,3 +2640,58 @@
...
+    for sample in TimeoutSampler(
+        wait_timeout=timeout,
+        sleep=2,
+        func=lambda: [
+            condition for condition in vm.vmi.instance.status.conditions if condition["type"] == "AgentConnected"
+        ],
+    ):
+        # Consider agent "down" when condition is absent OR explicitly not True
+        if not sample or all(condition.get("status") != "True" for condition in sample):
+            break
```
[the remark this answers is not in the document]
SiboWang1997: `if not sample or all(condition.get("status") != "True" for condition in sample)` will handle  status.conditions is None/False

utilities/virt.py
```diff
@@ -2638,3 +2640,58 @@ def username_password_from_cloud_init(vm_volumes: list[dict[str, Any]]) -> tuple
             return _match["user"], _match["password"]
 
     return "", ""
+
+
+def validate_virtctl_guest_agent_after_guest_reboot(vm: VirtualMachineForTests, os_type: str) -> None:
+    guest_reboot(vm=vm, os_type=os_type)
+    wait_for_running_vm(vm=vm, ssh_timeout=TIMEOUT_30MIN if os_type == OS_FLAVOR_WINDOWS else TIMEOUT_5MIN)
+    assert validate_virtctl_guest_agent_data_over_time(vm=vm), "Guest agent stopped responding after guest reboot"
+
```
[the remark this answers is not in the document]
SiboWang1997: Windows os matrix reboots reach to `Running` time will not exceed 10 sec , so the default `wait_until_running_timeout` of 4 minutes should already be sufficient. Unless we found that reboots to reach `Running` taking significantly longer, I’d suggest keeping the default.

PR #1757, opened 2025-08-12
Apply dependency between positive and negative VLAN tests

yossisegev:
test_negative_vlan_linux_bridge might pass as a false-negative. It eventually verifies no-connectivity between different VLANs, so it mgith pass even if, for example, VLAN is not supported on the cluster.
That's why this test is significant only if the positive scenario passes, hence they should be depended.

## Summary by CodeRabbit

* **Tests**
  * Improved stability of VLAN connectivity test suite by enforcing execution order and dependencies. The negative scenario now depends on the positive scenario and is skipped if prerequisites fail, reducing false negatives and CI noise. This speeds up feedback and clarifies pass/fail signals. No changes to runtime behavior or network functionality.

### DIFF
--- a/tests/network/connectivity/test_ovs_linux_bridge.py
+++ b/tests/network/connectivity/test_ovs_linux_bridge.py
@@ -75,6 +75,9 @@ class TestConnectivityLinuxBridge:
     @pytest.mark.polarion("CNV-11123")
     @pytest.mark.ipv4
     @pytest.mark.s390x
+    @pytest.mark.dependency(
+        name="test_positive_vlan_linux_bridge",
+    )
     def test_positive_vlan_linux_bridge(
         self,
         nad_linux_bridge_vlan_1,
@@ -92,6 +95,7 @@ class TestConnectivityLinuxBridge:
     @pytest.mark.polarion("CNV-11131")
     @pytest.mark.ipv4
     @pytest.mark.s390x
+    @pytest.mark.dependency(depends=["test_positive_vlan_linux_bridge"])
     def test_negative_vlan_linux_bridge(
         self,
         nad_linux_bridge_vlan_3,


yossisegev:
/verified

Verified the following way:
1. Trying to run only the negative test (the depending on) - the test was skipped due to the dependency:
```
tests/network/connectivity/test_ovs_linux_bridge.py::TestConnectivityLinuxBridge::test_negative_vlan_linux_bridge SKIPPED (test_negative_vlan_linux_bridge depends on test_positiv...)
```

2. Trying to run both tests with a valid VLAN tag (i.e. supported in the cluster) - both tests passed.

3. Trying to run both tests with an unsupported VLAN - the positive test failed, and then the negative one was skipped.

vsibirsk (approved):

EdDev:
> I am pretty sure it is not worth the effort. Sounds like a complex spaghetti solution, while pytest provides us the most suitable way to handle it.

The dependency plugin is not coming from pytest, it comes from its community. In fact, pytest core is has not embraced it as a core feature, which hints as it is not a good practice.
I think the usage of the dependency plugin is not worth it, it adds more harm than good.

However, I will not block it if you find someone else who is willing to support it (i.e. add an approve).

My preferred solution: Add a network sanity check to allow running all the test suite only if VLANs are supported on the infra. If that will be in place, then this test will never run.

rnetser:
> > > The dependency plugin is not coming from pytest, it comes from its community. In fact, pytest core has not embraced it as a core feature, which hints it is not a good practice. I think the usage of the dependency plugin is not worth it, it adds more harm than good.
> > 
> > 
> > As much as I appreciate and respect your professional opinion (and I really do), saying that something "...hints it is not a good practice" is a guess, not a concrete fact. Same for "it adds more harm than good": I understand that this is your opinion and you really believe in it, but I would appreciate if you explained how it adds more harm than good. Otherwise it is just someone's opinion from my POV.
> 
> I believe I expressed why I think it is not good before that, but lets try again.
> 
> 1. Skipping a test because the depending test was not executed will many times result in a passing execution. Skips are not failures. Therefore, it can easily be missed that the intended test has passed "empty".
> 2. Having a calculated guess based on experience can be considered opinionated. So yes, I am saying that test dependency is discouraged by not making it available in pytest core. I could add that test isolation is a core property which enables cleaner, stabler and maintainable tests codebase. But I do not think we should elaborate in this in this PR context, it is a topic that fits other media.
> 3. The scenario in which this fix is supporting is rare (running it without the others), so our investment in it with the cost of that dependency is IMO (opinionated) not worth it.
> 4. Alternatives have been provided.
> 
> > > However, I will not block it if you find someone else who is willing to support it (i.e. add an approve).
> > 
> > 
> > Then you are in-fact blocking it.
> 
> As long as I have not voted against it, I am not blocking. It is not important enough for me to approve something I feel uncomfortable with, but at the same time, it is also not important enough to vote explicitly against it. Staying neutral gives you the opportunity to convince someone else, (e.g. @rnetser or some other root approver) to take it.
> 
> > > My preferred solution: Add a network sanity check to allow running all the test suite only if VLANs are supported on the infra. If that will be in place, then this test will never run.
> > 
> > 
> > Which means another marker, and another network testing pipeline to handle. Do we really want it? Will DevOps be willing to handle it?
> 
> I do not know if it requires a new marker. I think you can depend on an existing marker. You need to assure that our tests run on infra that supports VLANs, which means that if a test collected the VLAN fixture, it better assure the infra supports it. How to assure it supports it? One options is to define a VLAN based connectivity with VMs or Pods to check it. I think there are several options/ways to handle it.

I agree with @EdDev , the way to solve this is not by adding a dependancy but by adding a marker to define what the tests require (vlan) , running on an env that has vlans and failing in sanity if there's not vlan support

rnetser:
> > > > > The dependency plugin is not coming from pytest, it comes from its community. In fact, pytest core has not embraced it as a core feature, which hints it is not a good practice. I think the usage of the dependency plugin is not worth it, it adds more harm than good.
> > > > 
> > > > 
> > > > As much as I appreciate and respect your professional opinion (and I really do), saying that something "...hints it is not a good practice" is a guess, not a concrete fact. Same for "it adds more harm than good": I understand that this is your opinion and you really believe in it, but I would appreciate if you explained how it adds more harm than good. Otherwise it is just someone's opinion from my POV.
> > > 
> > > 
> > > I believe I expressed why I think it is not good before that, but lets try again.
> > > 
> > > 1. Skipping a test because the depending test was not executed will many times result in a passing execution. Skips are not failures. Therefore, it can easily be missed that the intended test has passed "empty".
> > > 2. Having a calculated guess based on experience can be considered opinionated. So yes, I am saying that test dependency is discouraged by not making it available in pytest core. I could add that test isolation is a core property which enables cleaner, stabler and maintainable tests codebase. But I do not think we should elaborate in this in this PR context, it is a topic that fits other media.
> > > 3. The scenario in which this fix is supporting is rare (running it without the others), so our investment in it with the cost of that dependency is IMO (opinionated) not worth it.
> > > 4. Alternatives have been provided.
> > > 
> > > > > However, I will not block it if you find someone else who is willing to support it (i.e. add an approve).
> > > > 
> > > > 
> > > > Then you are in-fact blocking it.
> > > 
> > > 
> > > As long as I have not voted against it, I am not blocking. It is not important enough for me to approve something I feel uncomfortable with, but at the same time, it is also not important enough to vote explicitly against it. Staying neutral gives you the opportunity to convince someone else, (e.g. @rnetser or some other root approver) to take it.
> > > > > My preferred solution: Add a network sanity check to allow running all the test suite only if VLANs are supported on the infra. If that will be in place, then this test will never run.
> > > > 
> > > > 
> > > > Which means another marker, and another network testing pipeline to handle. Do we really want it? Will DevOps be willing to handle it?
> > > 
> > > 
> > > I do not know if it requires a new marker. I think you can depend on an existing marker. You need to assure that our tests run on infra that supports VLANs, which means that if a test collected the VLAN fixture, it better assure the infra supports it. How to assure it supports it? One options is to define a VLAN based connectivity with VMs or Pods to check it. I think there are several options/ways to handle it.
> > 
> > 
> > I agree with @EdDev , the way to solve this is not by adding a dependancy but by adding a marker to define what the tests require (vlan) , running on an env that has vlans and failing in sanity if there's not vlan support
> 
> I see the points of both of you. My problem is with adding another marker. This means another thing for DevOps to handle, and my concern is that this is becoming to cumbersome to handle.

But this is the direction we are going, this is why eventually the `special_infra` marker will be dropped and tests will be selected based on the required env prerequisites

EdDev:
> > I see the points of both of you. My problem is with adding another marker. This means another thing for DevOps to handle, and my concern is that this is becoming to cumbersome to handle.
> 
> But this is the direction we are going, this is why eventually the `special_infra` marker will be dropped and tests will be selected based on the required env prerequisites

We do not need to add a new marker, at least not one which require awareness from the test runner (e.g. CI).
We just need to make sure that the jobs which run sig-network tests, support VLANs.
When the VLAN fixture is ref in the collection of the tests, we should make sure a sanity check is performed that it is really supported on the running cluster. I think we can handle the details offline to sort the technicalities.
BTW, we know that the servers we run on are not always uniform, this is something we need to wither request to uniform or to accept as input the differences (e.g. the VLAN ID ranges).

PR #1762, opened 2025-08-12
Exceptional skip of IPv6 scenario on non-IPv4-single-stack cluster

yossisegev:
Although we avoid programmatic skips, this one is an exception where we must skip. The reason is that this test, which is a regression test of an old bug, is required to run only on a single-stack IPv4 cluster. We currenly don't have a way of marking a cluster as single-stack, and in addition - it wouldn't make sense to maintain a dedicated test lane just for this one test.
This is necessarily for stabilizing network test lanes, and avoid this test from erroneously fail.

## Summary by CodeRabbit

* **Tests**
  * Updated a network test to skip on non-IPv4 single-stack environments instead of failing, reducing false negatives and CI noise.
  * Clarifies test outcomes across mixed-stack setups and eases triage.
  * Improves test-suite stability without changing runtime behavior or user-facing functionality.

### DIFF
--- a/tests/network/general/test_dhcp6_configurations_on_ipv4_single_stack_cluster.py
+++ b/tests/network/general/test_dhcp6_configurations_on_ipv4_single_stack_cluster.py
@@ -31,9 +31,9 @@ def vm_cirros(
 
 
 @pytest.fixture(scope="module")
-def fail_if_not_ipv4_single_stack_cluster(ipv4_supported_cluster, ipv6_supported_cluster):
+def skip_if_not_ipv4_single_stack_cluster(ipv4_supported_cluster, ipv6_supported_cluster):
     if not ipv4_supported_cluster or ipv6_supported_cluster:
-        pytest.fail(reason="Test should only run on an IPv4 single-stack cluster")
+        pytest.skip("Test should only run on an IPv4 single-stack cluster")
 
 
 @pytest.fixture()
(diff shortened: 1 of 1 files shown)


EdDev (commented):
I'm not convinced, sorry.

The test is checking under-the-hood low lever details which is plain wrong, especially for tier-2 tests.
If you think it is important enough, make sure it is added to tier-1.

Perhaps we can try and handle it from the origin, i.e. to re-examine the bug and think how we can check it in a more generic upper level way.
@yossisegev , can you please share the bug link?

yossisegev:
> > > I'm not convinced, sorry.
> > > The test is checking under-the-hood low lever details which is plain wrong, especially for tier-2 tests. If you think it is important enough, make sure it is added to tier-1.t
> > 
> > 
> > @EdDev Can you please explain why it is plain wrong? And I don't think this check is considered an "under-the-hood" check. It is viewing the contents of a valid k8s `Network` resource.
> 
> The test which this thing serves is checking if the virt-launcher DHCPv6 is active or not. It does so by entering the pod and looking for list ports, etc. This is 1) too low level and not a user/product fronted API and 2) it is irrelevant in terms of interesting scenario for testing. AFAIK, we do not have an agenda to test single-stack IPv4. I think 1 is enough to drop this whole thing from the tests.
> 
> > > Perhaps we can try and handle it from the origin, i.e. to re-examine the bug and think how we can check it in a more generic upper level way. @yossisegev , can you please share the bug link?
> > 
> > 
> > I would love sharing it, but it is not mentioned anywhere in the in the matching polarion case, so I cannot locate it.
> 
> Great, yet another reason to delete it now. If such a bug will surface, we can re-examine everything at that time.

I asked Gemini to process everything you wrote here, and its response is:
```
I'm convinced!
```

PR #1769, opened 2025-08-13
Allow flaky test_connectivity_is_preserved_after_live_migration to run

yossisegev:
test_connectivity_is_preserved_after_live_migration usually pass, and when it fails - it's
not destructive, meaning that it doesn't leave the cluster broken.
Therefore, to allow better coverage until the issue is completely fixed, it should run.

### DIFF
--- a/tests/network/user_defined_network/test_user_defined_network.py
+++ b/tests/network/user_defined_network/test_user_defined_network.py
@@ -147,7 +147,6 @@ class TestPrimaryUdn:
     @pytest.mark.gating
     @pytest.mark.xfail(
         reason=f"{QUARANTINED}: Flaky test, fails on connecting to VM console; tracked in CNV-66779",
-        run=False,
     )
     def test_connectivity_is_preserved_after_live_migration(self, vma_udn, server, client):
         migrate_vm_and_verify(vm=vma_udn)


tests/network/user_defined_network/test_user_defined_network.py
```diff
@@ -147,7 +147,7 @@ def test_basic_connectivity_between_udn_vms(self, vma_udn, vmb_udn_non_migratabl
     @pytest.mark.gating
     @pytest.mark.xfail(
         reason=f"{QUARANTINED}: Flaky test, fails on connecting to VM console; tracked in CNV-66779",
-        run=False,
+        run=True,
```
RoniKishner, line 150: I suggest removing this line for consistency.
```suggestion
```
In the repository, there are other `xfail` cases without the `run` parameter, and none use `run=True`.

yossisegev: I guess what you mean to say, in much simpler (and fewer) words, is that `run=True` is the default. The fact that no other call to xfail in the repo sends it is not relevant.
But is it indeed the default? @rnetser I removed the explicit sending of `run=True` according to @RoniKishner comment, and verified that the behavior is as I expect (i.e. that the test is executed). However, I am looking at the [xfail implementation](https://github.com/pytest-dev/pytest/blob/main/src/_pytest/mark/structures.py#L503), and I see that `run` has no default value. Am I missing something? Because to me it seems that I do need to explicitly send `run=True`. WDYT?

yossisegev: /build-and-push-container

EdDev (approved):
/approve

yossisegev:
> > @yossisegev @EdDev quarantined process is there to not run =flaky tests. either this test is reported as automation blocker and marked as jira bug or it is quarantined and not running.
> 
> Then something is wrong. A flaky test says that 1) it passes part of the time (hopefully most of the time) and 2) fails randomly for unclear reason. That means that xfail is a fit as it will always pass, once with xpass and once with xfail, allowing to evaluate the flake.
> 
> I see no value for it to never run, it is basically equivalent to removing it from the code. What is the logic behind it?

@rnetser In case you missed @EdDev 's response.

PR #1780, opened 2025-08-14
add s390x marker for infrastructure tests

ashokpariya0:
##### Short description:

Add Tier2 infrastructure marker changes for s390x.

This PR adds the s390x markers to the existing Tier2 infrastructure tests. This helps enable the Tier2 infrastructure pipeline for s390x.

## Summary by CodeRabbit

* **Tests**
  * Added s390x architecture markers across multiple test suites (golden images, boot sources, data sources, instance types, vhostmd, VM console proxy) to refine test selection on s390x platforms.
  * Renamed one public test for clarity; no changes to test logic, parameters, or assertions.
  * No behavioral changes to the product; updates affect only test metadata and execution scope.

### DIFF
--- a/tests/infrastructure/golden_images/update_boot_source/test_latest_rhel_image.py
+++ b/tests/infrastructure/golden_images/update_boot_source/test_latest_rhel_image.py
@@ -85,6 +85,7 @@ def rhel_vm_minor_ver_num(rhel_vm):
     ],
     indirect=True,
 )
+@pytest.mark.s390x
 def test_latest_minor_ver_rhel(libosinfo_rhel_minor_ver_num, rhel_vm_minor_ver_num):
     assert libosinfo_rhel_minor_ver_num == rhel_vm_minor_ver_num, (
         f"os versions mismatch, VM minor version: {rhel_vm_minor_ver_num}, "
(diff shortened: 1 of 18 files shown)


tests/infrastructure/golden_images/update_boot_source/test_latest_rhel_image.py
```diff
@@ -85,6 +85,7 @@ def rhel_vm_minor_ver_num(rhel_vm):
     ],
     indirect=True,
 )
+@pytest.mark.s390x
```
RoniKishner: I think you can remove this, it has no effect on the cluster type so running it on any cluster is ok, it tests that the image we use from the data import cron is updated to the latest version

tests/infrastructure/instance_types/test_update_vm_with_instancetype_preference.py
```diff
@@ -88,6 +88,7 @@ def updated_vm_with_instancetype_and_preference(
 
 @pytest.mark.gating
 @pytest.mark.polarion("CNV-9680")
+@pytest.mark.s390x
```
RoniKishner: I think you can remove it as it tests the validator and arch type has no effect on it.

PR #1794, opened 2025-08-17
[4.19]Changing test implementation for metric kubevirt_nodes_with_kvm (#1731)

OhadRevah:
##### Short description:
The test now is more simple then before, instead of changing image in daemonSet, I simply get the number of schedulable nodes and compare it with the prometheus result.
Also removed the alert that depends on this metric test since we are not testing alert in T2 anymore. first if a pod exists before deletion and in this way I make sure it won't fail in this part of the setup.
##### jira-ticket:
https://issues.redhat.com/browse/CNV-67428


openshift-virtualization-qe-bot-3:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:cnv-4.19 published

PR #1799, opened 2025-08-17
net, dhcp: isolate DHCP - VLAN-scope NAD, intended client/iface

Anatw:
- DHCP NAD: add VLAN tag and include it in name.
In server:
- Bind dhcpd to eth3 via DHCPDARGS and restart (instead of start) service.
- dhcpd.conf: restrict to known-clients; add host by client iface MAC; set router/DNS.
In test:
- Reorder fixtures for correct bring-up
- Add minor typing.
- Use DHCP_INTERFACE_NAME const instead of hardcoded "eth3" (cloud-init/nmcli).

Prevents node NIC leases that destabilized the cluster.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-20812


tests/network/l2_bridge/conftest.py
```diff
@@ -161,7 +167,7 @@ def _cloud_init_data(
     mpls_route_next_hop,
     cloud_init_extra_user_data,
 ):
-    network_data_data = {
+    network_data_data: dict[str, Any] = {
```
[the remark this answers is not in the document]
EdDev: The value of typing comes when we can validate it. I do not think we can validate anything without having everything covered.
I do not think we targeted to start adding it to existing test or conftest files, unless you dedicate a PR for it.
We do want and should make sure to add it of any new module or package that is added, especially libs/helpers.

So yes, I think it should be removed.

tests/network/l2_bridge/conftest.py
```diff
@@ -288,11 +294,15 @@ def bridge_attached_vm(
 
 
 @pytest.fixture(scope="class")
-def l2_bridge_running_vm_a(namespace, worker_node1, l2_bridge_all_nads, unprivileged_client):
+def l2_bridge_running_vm_a(namespace, worker_node1, l2_bridge_all_nads, unprivileged_client, l2_bridge_running_vm_b):
+    dhcp_iface_name = l2_bridge_all_nads[2]
```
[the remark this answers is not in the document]
EdDev: > Regarding your question, my assumption is that the meaning in iface_name is to the spec.template.spec.domain.devices.interfaces. In the VM it should reference the NAD name.

We have a spec interface and network logical name, then we have a reference to a NAD name and we even have a pod and guest interface (nic) name.
All are colliding and it is hard to understand what each refers to.
But here, it looks to me like plain wrong. The name on the interface/network in the spec has nothing to do with the NAD name. The fact one defined both with the same name does not mean we should see `iface_name = nad1.name`, because that hints that the two are suppose to have the same name.
This must be fixed in a follow up, better sooner than later.

tests/network/l2_bridge/conftest.py
```diff
@@ -365,7 +375,13 @@ def eth3_nmcli_connection_uuid(l2_bridge_running_vm_b):
 def configured_l2_bridge_vm_a(l2_bridge_running_vm_a):
     run_ssh_commands(
         host=l2_bridge_running_vm_a.ssh_exec,
-        commands=[shlex.split(DHCP_SERVICE_RESTART)],
+        commands=[
+            shlex.split(
+                "sudo bash -c "
+                + shlex.quote(f"cat > /etc/sysconfig/dhcpd <<'EOF'\nDHCPDARGS=\"{DHCP_INTERFACE_NAME}\"\nEOF")
```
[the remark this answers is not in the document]
EdDev: Ohh, this is so odd.
Looks to me like the starting point here is over-complicated. We do not really need to use and hack services, it is an overkill.
It could have been simplified by running the server from the command line: `sudo dhcpd -4 eth0 -cf /etc/dhcp/dhcpd.conf`.

And the fact it is done on a VM is another waste. A simple pod should have been enough, without all the terminal steps.

As the current change hardens and isolates better the DHCP server usage, I can live with this getting in as is.
However, in parallel, from a code maintenance perspective, it feels we continue the original mess and adding to it.

tests/network/l2_bridge/conftest.py
```diff
@@ -288,11 +294,12 @@ def bridge_attached_vm(
 
 
 @pytest.fixture(scope="class")
-def l2_bridge_running_vm_a(namespace, worker_node1, l2_bridge_all_nads, unprivileged_client):
+def l2_bridge_running_vm_a(namespace, worker_node1, dhcp_nad, unprivileged_client, l2_bridge_running_vm_b):
```
EdDev, line 297: Removing `l2_bridge_all_nads` is not possible, it is used below (line 320). You can only add.

EdDev (changes requested):
Thank you for updating the PR/commit text, it is describing well the problem and the solution to it.

For the PR description, please keep the format of the subjects in the template. I usually use one or two of them. There is also no need to quote the text. You can look at PRs from me which have content there to see what I mean.

There is still one issue in the change now, see my inline comments.

Anatw:
Change: remove the small typing addition.

openshift-virtualization-qe-bot-2:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-1799 published

Anatw:
Change: Updated commit msg with link to setting up the DHCP server.

tests/network/l2_bridge/test_l2_ovs_linux_bridge.py
```diff
@@ -69,9 +69,9 @@ def test_connectivity_l2_bridge(
     @pytest.mark.polarion("CNV-2282")
     def test_dhcp_broadcast(
         self,
-        configured_l2_bridge_vm_a,
-        l2_bridge_running_vm_b,
         dhcp_nad,
+        l2_bridge_running_vm_b,
+        configured_l2_bridge_vm_a,
```
EdDev, line 74: What are these changes for? If it was because you wanted the NAD to be created first, it is wrong to depend on the order of fixture listing and in this case not needed (because each fixture already depends on the NAD).

EdDev: It is incorrect and bad practice to depend on the definition order of fixtures in the tests.
If it was not a must, this change should not have been done at all.

There may be exceptional cases when there is no other choice but to depend on the order of definition, but these should be rare enough and avoidable.

Anatw: Agreed, depending on the order is bad practice.
The order here is not practical, it this is done here for readability purposes only.

EdDev: I "readability" you refer here implies order, and we want to say there is no order. Or more accurately, we want to enforce order through the hierarchy of fixtures and not through the order they are referenced.
However, I am not going to argue about this further at this stage. Please try to avoid it next time.

EdDev (approved):
Thank you.

/approve

vsibirsk (approved):

tests/network/l2_bridge/conftest.py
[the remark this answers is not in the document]
Anatw: Ack.

openshift-virtualization-qe-bot-3:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-1799.

PR #1818, opened 2025-08-19
Add workflow to push image to stage

ghost:
Added a workflow_run github action to push the fedora container image to stage repo in quay.io

Note: This workflow_run will be automatically called from component-builder.yml up on its successful completion of this pull_request event. This workflow_run is required as the action triggered from pull_request doesn't have access to repository secrets for security reasons.

##### jira-ticket:
"NONE"

## Summary by CodeRabbit

* **Chores**
  * Added an automated CI workflow that, after checks succeed, publishes multi-architecture container images (x86_64, aarch64, s390x) to the staging registry.
  * Combines per-architecture images into a single manifest for seamless multi-arch pulls.
  * Publishes per-PR tagged images to enable targeted testing in staging and speed up delivery of test images.

### DIFF
--- a/.github/workflows/component-builder-publish-stage.yml
+++ b/.github/workflows/component-builder-publish-stage.yml
@@ -1,0 +1,53 @@
+name: "Push Container Images to Stage"
+on:
+  workflow_run:
+    workflows: [ "Component Builder Check" ]
+    types:
+      - completed
+
+jobs:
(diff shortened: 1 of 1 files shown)


stesrn:
This PR is no longer required as the workflow_run is not required now. GH action is called with the event pull_request_target which helped to achieve the requirement

PR #1842, opened 2025-08-21
Add architecture-specific preference handling for s390x and arm64

ashokpariya0:
Modified global_config.py to define RHEL9_PREFERENCE and RHEL10_PREFERENCE based on architecture (s390x, arm64, or default). Updated test_vm_with_instance_and_pref.py to adjust CLOCK_TIMER for s390x by disabling hyperv and removing kvm timer. Refactored test_downwardmetrics_virtio.py to use get_preference_name() for dynamic RHEL preference naming, supporting s390x and arm64. Updated vm_console_proxy/conftest.py to use dynamic PREFERENCE for rhel.10 based on architecture.

##### Short description:
Add architecture-specific preference and timer handling for s390x and arm64 in test infrastructure.

This PR introduces architecture-aware configurations to support s390x and arm64 architectures in the test suite for virtual machine preferences and timers. The changes ensure that RHEL version preferences and CLOCK_TIMER settings are dynamically adjusted based on the cluster architecture

rhel.8 cluster-preference is same for all arch,
but for rhel.9 and rhel.10 we have append with arch value,

```
ap $ oc get virtualmachineclusterpreferences | grep rhel.8
rhel.8                     14d
rhel.8.desktop             14d
rhel.8.dpdk                14d
```
```
ap $ oc get virtualmachineclusterpreferences | grep rhel.9
rhel.9                     14d
rhel.9.arm64               14d
rhel.9.desktop             14d
rhel.9.dpdk                14d
rhel.9.realtime            14d
rhel.9.s390x               14d
```
```
ap $ oc get virtualmachineclusterpreferences | grep rhel.10
rhel.10                    14d
rhel.10.arm64              14d
rhel.10.s390x              14d
```


tests/global_config.py
```diff
@@ -195,6 +193,10 @@ def _get_default_storage_class(sc_list):
     {"rhel9": {"template_os": "rhel9.0"}},
 ]
 
+
+RHEL9_PREFERENCE = f"rhel.9.{arch}" if arch in ["arm64", "s390x"] else "rhel.9"
+RHEL10_PREFERENCE = f"rhel.10.{arch}" if arch in ["arm64", "s390x"] else "rhel.10"
```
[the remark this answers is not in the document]
geetikakay: yes but since used same code twice please consider reusing as version will continue to increase

tests/infrastructure/instance_types/test_vm_with_instance_and_pref.py
```diff
@@ -22,6 +25,12 @@
     "rtc": {"present": True, "tickPolicy": "catchup", "track": "guest"},
 }
 
+# Modify CLOCK_TIMER for s390x architecture
+if arch == "s390x":
```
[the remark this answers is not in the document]
geetikakay: use S390X

tests/infrastructure/vhostmd/test_downwardmetrics_virtio.py
```diff
@@ -91,6 +92,26 @@ def parsed_metrics_command_data(vm):
...
+    Returns:
+        str: The formatted preference name (e.g., 'rhel.8', 'rhel.10.s390x', 'rhel.10.arm64').
+             For s390x or arm64 architectures, appends the architecture suffix (e.g., '.s390x')
+             unless the version is 'rhel8'. For other architectures or 'rhel8', returns the version
+             with a dot separator (e.g., 'rhel.8', 'rhel.10').
+    """
+    preference_name = re.sub(r"(\D+)(\d+)", r"\1.\2", rhel_version)
+    # Append architecture suffix for s390x or arm64, except for rhel8
+    arch = get_cluster_architecture()
+    if rhel_version != "rhel8" and arch in ("s390x", "arm64"):
```
geetikakay, line 110: save and reuse in utilities constants because we need it in tests/global_config.py

tests/infrastructure/vm_console_proxy/conftest.py
```diff
@@ -38,6 +39,10 @@
 LOGGER = logging.getLogger(__name__)
 
 
+ARCH = get_cluster_architecture()
+PREFERENCE = f"rhel.10.{ARCH}" if ARCH in ["arm64", "s390x"] else "rhel.10"
```
geetikakay, line 43: this can be fetched from tests/global_config.py

tests/infrastructure/vm_console_proxy/conftest.py
```diff
@@ -95,7 +100,7 @@ def vm_for_console_proxy(namespace, unprivileged_client):
         namespace=namespace.name,
         client=unprivileged_client,
         vm_instance_type=VirtualMachineClusterInstancetype(name="u1.small"),
-        vm_preference=VirtualMachineClusterPreference(name="rhel.10"),
+        vm_preference=VirtualMachineClusterPreference(name=PREFERENCE),
```
geetikakay, line 103: for arm will u1.small work?

tests/infrastructure/instance_types/test_vm_with_instance_and_pref.py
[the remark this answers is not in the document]
ashokpariya0: Done.

tests/infrastructure/vm_console_proxy/conftest.py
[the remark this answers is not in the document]
ashokpariya0: Done, I've moved these changes to utilities/constants.py

ashokpariya0: Done.

tests/infrastructure/instance_types/test_vm_with_instance_and_pref.py
```diff
@@ -6,20 +6,38 @@
...
+
+    if get_cluster_architecture() == S390X:
+        clock_timer["hyperv"]["present"] = False
+        if "kvm" in clock_timer:
+            del clock_timer["kvm"]
+
+    return clock_timer
+
+
+CLOCK_TIMER = get_clock_timer_config()
```
dshchedr, line 40: it should be a fixture, because with current approach it will be executed all the time when pytest initiate the scope of the test runs (even if you don't run your tests)

tests/infrastructure/instance_types/test_vm_with_instance_and_pref.py
```diff
@@ -6,20 +6,38 @@
...
+    """
+    clock_timer = {
+        "hpet": {"present": False},
+        "hyperv": {"present": True},
+        "kvm": {"present": True},
+        "pit": {"present": True, "tickPolicy": "delay"},
+        "rtc": {"present": True, "tickPolicy": "catchup", "track": "guest"},
+    }
+
+    if get_cluster_architecture() == S390X:
```
dshchedr, line 32: don't use function directly, we have a  fixture for that

tests/infrastructure/vhostmd/test_downwardmetrics_virtio.py
```diff
@@ -91,6 +92,26 @@ def parsed_metrics_command_data(vm):
...
+
+    Returns:
+        str: The formatted preference name (e.g., 'rhel.8', 'rhel.10.s390x', 'rhel.10.arm64').
+             For s390x or arm64 architectures, appends the architecture suffix (e.g., '.s390x')
+             unless the version is 'rhel8'. For other architectures or 'rhel8', returns the version
+             with a dot separator (e.g., 'rhel.8', 'rhel.10').
+    """
+    preference_name = re.sub(r"(\D+)(\d+)", r"\1.\2", rhel_version)
+    # Append architecture suffix for s390x or arm64, except for rhel8
+    arch = get_cluster_architecture()
```
dshchedr, line 109: same here - use fixture

utilities/constants.py
```diff
@@ -43,6 +43,11 @@
 X86_64 = "x86_64"
 
 
+def _get_rhel_preference(version):
+    arch = get_cluster_architecture()
+    return f"rhel.{version}.{arch}" if arch in ["arm64", "s390x"] else f"rhel.{version}"
```
dshchedr, line 48: function inside the constants.py file?

utilities/constants.py
```diff
@@ -736,8 +741,8 @@ class NamespacesNames:
 CENTOS_STREAM9_PREFERENCE = "centos.stream9"
 CENTOS_STREAM10_PREFERENCE = "centos.stream10"
 RHEL8_PREFERENCE = "rhel.8"
-RHEL9_PREFERENCE = "rhel.9"
-RHEL10_PREFERENCE = "rhel.10"
+RHEL9_PREFERENCE = _get_rhel_preference(version=9)
+RHEL10_PREFERENCE = _get_rhel_preference(version=10)
```
dshchedr, line 745: you should not use it like that because of the same reason as above

dshchedr (changes requested):

tests/infrastructure/vhostmd/test_downwardmetrics_virtio.py
ashokpariya0: Done.

utilities/constants.py
ashokpariya0: These preference values are reused in many places, so don't want recompute it across the codebase,
also the architecture doesn't change at runtime, so idea here is storing them as constants simplifies usage across the test suite.

tests/infrastructure/instance_types/test_vm_with_instance_and_pref.py
ashokpariya0: clock timer is global variable and used at multiple places in this module, also i think 
Injecting it as a fixture doesn't add much value here, as there's no need for test-level customization. I did try converting it to a fixture, but it became cumbersome(it did not work) — especially in places like below

https://github.com/RedHatQE/openshift-virtualization-tests/blob/be2fb31627392810e20f1e1ecabaea79dd2ecc61/tests/infrastructure/instance_types/test_vm_with_instance_and_pref.py#L126

utilities/constants.py
rnetser: it is much cleaner to overwrite the values in the relvant global_config file and here add a comment above line 744 mentioning that the values are changed there

tests/infrastructure/instance_types/test_vm_with_instance_and_pref.py
rnetser: +1 for a fixture; no need to run the code here, if not needed (e.g during collectonly)

rnetser (changes requested):

PR #1843, opened 2025-08-21
Add connectivity check to test_connectivity_after_migration

Anatw:
Add verify_connectivity_before_migration fixture to assert baseline connectivity before migration. This way, if the test fails we know the migration caused it and not an existing network problem.


Anatw:
Fix test's docstring according to comment.

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -20,6 +21,20 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def ping_via_console(src_vm, dst_vm):
```
servolkov, line 24: 1) why not to move it to libs/net?
2) why not to use typing?

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -56,6 +71,26 @@ def running_vm_for_migration(
...
 
+@pytest.fixture(scope="module")
+def vm_console_connection_ready(running_vm_for_migration):
+    wait_for_console(
+        vm=running_vm_for_migration,
+    )
+
+
+@pytest.fixture()
+def verify_connectivity_before_migration(
```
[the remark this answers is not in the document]
servolkov: why is this a fixture at all? Sometimes I think we have too many fixtures.. Why not just to add this check to `migrated_vmi`, for example?

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -90,24 +118,15 @@ def test_connectivity_after_migration(
...
-    static_vm_ip = running_vm_static.vmi.interfaces[0]["ipAddress"]
+    Validate connectivity after migrating a VM that uses masquerade.
 
-    vm_console_run_commands(
-        vm=running_vm_for_migration,
-        commands=[f"ping {static_vm_ip} -c 10 -w 10"],
-        timeout=10,
-    )
+    - Uses the VM console (not SSH) to ping from the migrated VM to a static VM to avoid masking issues.
+    - It is importand to run the ping immediately after the VM is migrated to the destination node.
```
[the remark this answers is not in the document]
Anatw: It should be there now.

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -20,6 +20,21 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def ping_via_console(src_vm, dst_vm):
+    """
+    Ping from src_vm to dst_vm via VM console to avoid SSH masking issues.
```
rnetser, line 25: please use proper google dcosting formatting

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -20,6 +20,21 @@
...
+    Uses '-c 10 -w 10' so that any packet loss causes a non‑zero exit status.
+    """
+    dst_ip = dst_vm.vmi.interfaces[0]["ipAddress"]
+    console_command_timeout_seconds = 10
```
rnetser, line 29: no reason to save and then use

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -56,15 +71,35 @@ def running_vm_for_migration(
...
+@pytest.fixture()
+def migrated_vmi(
+    running_vm_for_migration,
+    inter_vm_connectivity_exists,
+):
     LOGGER.info(f"Migrating {running_vm_for_migration.name}. Current node: {running_vm_for_migration.vmi.node.name}")
 
     ip_before = running_vm_for_migration.vmi.interfaces[0]["ipAddress"]
     migrated_vmi = migrate_vm_and_verify(vm=running_vm_for_migration, wait_for_migration_success=False)
+    ip_change_timeout_seconds = 60
```
rnetser, line 99: ditto

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -94,20 +122,11 @@ def test_connectivity_after_migration(
...
+    Validate connectivity after migrating a VM that uses masquerade.
 
-    vm_console_run_commands(
-        vm=running_vm_for_migration,
-        commands=[f"ping {static_vm_ip} -c 10 -w 10"],
-        timeout=10,
-    )
+    - Uses the VM console (not SSH) to ping from the migrated VM to a static VM to avoid masking issues.
+    - Run the ping right after the VM migration (once the VM IP changes) to detect a bug that appears
+      only immediately after migration; waiting may hide it.
```
rnetser, line 129: nit: `ping right after the VM migration (once the VM IP changes)` - the test waits for console connecttivity so this is not 100% accurate.

tests/network/migration/test_masquerade_connectivity_after_migration.py
rnetser: why move it? it is only used in this test.

rnetser (changes requested):

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -64,7 +99,7 @@ def migrated_vmi(running_vm_for_migration):
     migrated_vmi = migrate_vm_and_verify(vm=running_vm_for_migration, wait_for_migration_success=False)
 
     for sample in TimeoutSampler(
-        wait_timeout=60,
+        wait_timeout=TIMEOUT_1MIN,
```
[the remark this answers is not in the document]
EdDev: > I saw other PRs in our team moving back to using these constants

Where did you see it?
I expect sig-network to never add new usages of it and to remove usages as part of changes of other things (it is not worth dedicating work on this specific thing, but piggy back on other changes makes sense).

tests/network/migration/test_masquerade_connectivity_after_migration.py
EdDev: > why move it? it is only used in this test.

Because `ping` through a console is a common operation and we do not want someone else to invent it again.
Said that, it can be done with a convincing change to use it in another module/package. But that is nitpicking IMO.

EdDev (changes requested):
The PR does too many things while claiming it does only one.

In practice, it refactored the existing and then added a pre-requirement in a form of a fixture.
It then triggered a whole discussion on the review about things that could be avoided.

You can choose how to handle it:
- Implement only the added pre-requirement by adding one fixture and seeing nothing more in the diff.
- Implement a refactoring in a different PR and then get back to the functionality that this PR asked to introduce.

tests/network/migration/test_masquerade_connectivity_after_migration.py
Anatw: Thanks for the feedback. I agree consistency helps. I updated this function to a proper google-style docstring.  +1 on adding a linter rule to help enforcing consistency across the repo.

tests/network/migration/test_masquerade_connectivity_after_migration.py
Anatw: Ack, I'll update the docstring and remove the fixture.

Anatw:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/1843#pullrequestreview-3164511113
> The PR does too many things while claiming it does only one.
> 
> In practice, it refactored the existing and then added a pre-requirement in a form of a fixture.
> It then triggered a whole discussion on the review about things that could be avoided.
> 
> You can choose how to handle it:
> 
> Implement only the added pre-requirement by adding one fixture and seeing nothing more in the diff.
> Implement a refactoring in a different PR and then get back to the functionality that this PR asked to introduce.

Thanks for the feedback. I’ll scope this PR to the single functional change it intended: add a pre‑migration console-based connectivity check. I’ll move the refactors (docstrings, timeout inlining, fixture cleanup) to a separate PR to keep this one minimal. I think both PRs can be up at the same time, the second to get merged will need conflict management.

Anatw:
Since this PR is only introducing the new connectivity check before migration, I created a new PR for refactoring the existing scenario: https://github.com/RedHatQE/openshift-virtualization-tests/pull/1908

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -57,7 +57,24 @@ def running_vm_for_migration(
 
 
 @pytest.fixture()
-def migrated_vmi(running_vm_for_migration):
+def inter_vm_connectivity_exists(
+    running_vm_static,
```
rnetser, line 61: (not in this PR but `running_vm_static` should be renamed to something meaningful)

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -57,7 +57,24 @@ def running_vm_for_migration(
...
+    running_vm_for_migration,
+):
+    static_vm_ip = running_vm_static.vmi.interfaces[0]["ipAddress"]
+
+    vm_console_run_commands(
```
rnetser, line 66: i'd expect this to be as part of `running_vm_for_migration` - the fixture should yield a properly configured VM and connectivity is one such thing.

tests/network/migration/test_masquerade_connectivity_after_migration.py
Anatw: Ack, I'll change it in the refactoring PR (https://github.com/RedHatQE/openshift-virtualization-tests/pull/1908), and also change the other VM accordingly (running_vm_for_migration).

Anatw: https://github.com/RedHatQE/openshift-virtualization-tests/pull/1908/commits/8c099a536bc4e108ed508431ae9d246c6f4b1641

tests/network/migration/test_masquerade_connectivity_after_migration.py
Anatw: I'm changing it, but I'd like to further discuses it.
To me it seems like connection two separate actions into a single function instead of doing a single action in each fixture. This also introduce dependency in the VM creation - the destination VM must be created before the target VM. Can you please further explain why these two actions should be handled in a single fixture in this case?

openshift-virtualization-qe-bot-3:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-1843 published

tests/network/migration/test_masquerade_connectivity_after_migration.py
yossisegev: > The only missing info is the "sec", however, by default most (if not all) timeouts are set with that scale.

@EdDev You've been deep in python/Go and away from C for quite too long...

tests/network/migration/test_masquerade_connectivity_after_migration.py
```diff
@@ -53,11 +54,22 @@ def running_vm_for_migration(
     ) as vm:
         vm.start(wait=True)
         vm.wait_for_agent_connected()
+
+        static_vm_ip = running_vm_static.vmi.interfaces[0]["ipAddress"]
+
+        vm_console_run_commands(
```
yossisegev, line 60: The rational for this part is clear from the context of this PR, but once merged - it won't be clear what does a connectivity action have to do in a fixture which is supposed to spin-up a VM and make sure it's up (and the fixture name also implies that).
So please add a short comment explaining why this ping is done.
Thank you

yossisegev (changes requested):

PR #2014, opened 2025-09-08
Modify test_disk_image_after_clone to use DataSource

josemacassan:
##### Short description:
Modify the test test_disk_image_after_clone to use a DataSource instead of a DataVolume for getting the vm image.
##### More details:
In order to get the image from DataSource instead from artifactor the test has been modify to create the DV from DS.
##### jira-ticket:
https://issues.redhat.com/browse/CNV-68035


tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +192,54 @@ def test_successful_vm_from_cloned_dv_windows(
...
-                "dv_name": "dv-source",
-                "image": FEDORA_LATEST.get("image_path"),
-                "dv_size": Images.Fedora.DEFAULT_DV_SIZE,
-            },
-            marks=(pytest.mark.polarion("CNV-4035")),
-        )
-    ],
-    indirect=True,
-)
 @pytest.mark.s390x
```
jpeimer, line 196: Please move `sno` and `s390x` marks to the test

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +192,54 @@ def test_successful_vm_from_cloned_dv_windows(
...
+@pytest.fixture(scope="class")
```
jpeimer, line 197: Can be scope `module`

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +192,54 @@ def test_successful_vm_from_cloned_dv_windows(
...
+def fedora_data_source_scope_module(golden_images_namespace):
+    return DataSource(
+        namespace=golden_images_namespace.name,
+        name=OS_FLAVOR_FEDORA,
+        client=golden_images_namespace.client,
+        ensure_exists=True,
+    )
```
jpeimer, line 204: Please move the fixutre to the `tests/storage/cdi_clone/conftest.py`

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +192,54 @@ def test_successful_vm_from_cloned_dv_windows(
...
+        or source_dict.get("status", {}).get("restoreSize"),
+        api_name="storage",
+        storage_class=storage_class_name_scope_module,
+        source_ref={
+            "kind": data_source.kind,
+            "name": data_source.name,
+            "namespace": data_source.namespace,
+        },
     ) as cdv:
         cdv.wait_for_dv_success()
```
jpeimer, line 233: Please make sure it works for WaitForFirstConsumer storages (like `hostpath-csi-pvc-block`).
I think we can remove `cdv.wait_for_dv_success()` because `running_vm()` will check that the DV succeeded.

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +192,54 @@ def test_successful_vm_from_cloned_dv_windows(
...
-        create_vm_and_verify_image_permission(dv=cdv)
+
+        with create_vm_from_dv(
+            dv=cdv, vm_name="fedora-vm", os_flavor=OS_FLAVOR_FEDORA, memory_guest=Images.Fedora.DEFAULT_MEMORY_SIZE
```
jpeimer, line 236: We need to add `wait_for_interfaces` arg to `create_vm_from_dv`, and pass True for the non-cirros VMs

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +192,54 @@ def test_successful_vm_from_cloned_dv_windows(
...
+        ) as vm_dv:
+            check_disk_count_in_vm(vm=vm_dv)
```
jpeimer, line 238: I don't think we need to have `check_disk_count_in_vm` in this test.
`create_vm_and_verify_image_permission` was calling `verify_vm_disk_image_permission`, please ask the scrum team if we still want to have this check

jpeimer (changes requested):

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +192,54 @@ def test_successful_vm_from_cloned_dv_windows(
...
-    data_volume_multi_storage_scope_function,
+    storage_class_name_scope_module,
+    fedora_data_source_scope_module,
     cluster_csi_drivers_names,
 ):
-    storage_class = data_volume_multi_storage_scope_function.storage_class
-    with create_dv(
-        source="pvc",
-        dv_name="dv-cnv-4035",
+    data_source = fedora_data_source_scope_module
```
jpeimer, line 215: No need to save

tests/storage/cdi_clone/test_clone.py
josemacassan: In this case, using other storageClasses with `VolumeBindingMode: WaitForFirstConsumer` and removing `cdv.wait_for_dv_success()` doen't work out. 
 The creation of the VM is sequential right after DV creation. When the test tries to create the VM from the DV is not available yet, so I think is needed any kind of waiting, `cdv.wait_for_dv_success()` is not the best choice but for sure needed some wait.

tests/storage/cdi_clone/test_clone.py
josemacassan: Workaround for this is to add `start=False` to `create_vm_from_dv` as `wait_for_interfaces` is defined into `running_vm` function. The logic of the code makes possible to not start the VM and then call `running_vm(wait_for_interfaces=True)`.

tests/storage/cdi_clone/test_clone.py
josemacassan: Done!

tests/storage/cdi_clone/test_clone.py
josemacassan: Done!

tests/storage/cdi_clone/test_clone.py
jpeimer: The idea of the `WaitForFirstConsumer` storage is that we can create a DV/PVC, but not bind the PVC until there's a first consumer for it (for example, to avoid PVC and VM scheduled on different nodes). So VM can be its first consumer, and they will be scheduled on the same node.

Where exactly does it fail if we remove `wait_for_dv_success`?

jpeimer (changes requested):

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +191,49 @@ def test_successful_vm_from_cloned_dv_windows(
...
-        create_vm_and_verify_image_permission(dv=cdv)
+
+        with create_vm_from_dv(
+            dv=cdv,
+            vm_name="fedora-vm",
+            os_flavor=OS_FLAVOR_FEDORA,
+            memory_guest=Images.Fedora.DEFAULT_MEMORY_SIZE,
+            start=False,
+        ) as vm_dv:
+            running_vm(vm=vm_dv, wait_for_interfaces=True)
```
jpeimer, line 231: Please add `wait_for_interfaces` argument to the function `def create_vm_from_dv`. We will use more and more non-cirros images, and we want to reuse this function.

jpeimer (changes requested):

tests/storage/cdi_clone/test_clone.py
josemacassan: I get a 404 not found error when trying to access the DV.

tests/storage/cdi_clone/test_clone.py
josemacassan: Done!

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +191,50 @@ def test_successful_vm_from_cloned_dv_windows(
...
+        storage_class=storage_class_name_scope_module,
+        source_ref={
+            "kind": fedora_data_source_scope_module.kind,
+            "name": fedora_data_source_scope_module.name,
+            "namespace": fedora_data_source_scope_module.namespace,
+        },
     ) as cdv:
-        cdv.wait_for_dv_success()
-        create_vm_and_verify_image_permission(dv=cdv)
+        cdv.wait(timeout=TIMEOUT_10MIN, wait_for_exists_only=True)
```
jpeimer, line 223: No need for such a long timeout 

```suggestion
        cdv.wait(wait_for_exists_only=True)
```

jpeimer (changes requested):

tests/storage/cdi_clone/test_clone.py
```diff
@@ -193,48 +192,54 @@ def test_successful_vm_from_cloned_dv_windows(
...
-        storage_class=storage_class,
+        size=source_spec_dict.get("resources", {}).get("requests", {}).get("storage")
+        or source_dict.get("status", {}).get("restoreSize"),
+        api_name="storage",
+        storage_class=storage_class_name_scope_module,
+        source_ref={
+            "kind": data_source.kind,
+            "name": data_source.name,
+            "namespace": data_source.namespace,
+        },
```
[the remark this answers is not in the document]
jpeimer: we should use storage api

jpeimer (approved):

vsibirsk (approved):

PR #2065, opened 2025-09-14
Storage: retry certificate retrieval

dalia-frank:
##### Short description:

When trying to retrieve the certificate, we might fail occasionally due to
connection errors. This change adds retry logic so the fixture will attempt
multiple times before giving up. If retries still fail, the fixture raises
an error so the issue can be investigated.


tests/storage/conftest.py
```diff
@@ -115,18 +117,37 @@ def hpp_resources(request, admin_client):
...
+    def _fetch_cert():
+        try:
+            return ExecCommandOnPod(utility_pods=workers_utility_pods, node=worker_node1).exec(
+                command=(
+                    f"openssl s_client -showcerts -connect {svc_ip}:443 </dev/null 2>/dev/null | "
+                    "sed -n '/-----BEGIN/,/-----END/p'"
+                )
+            )
+        except workers_utility_pods.ExecOnPodError as e:
+            LOGGER.warning(f"Transient error fetching cert from {svc_ip}: {e}")
```
vsibirsk, line 130: won't error log from sampler be enough? (this will spam 12 logs if timeout reached)

dalia-frank: you are right

vsibirsk (dismissed):

tests/storage/conftest.py
```diff
@@ -115,18 +117,36 @@ def hpp_resources(request, admin_client):
...
+
+    def _fetch_cert():
+        try:
+            return ExecCommandOnPod(utility_pods=workers_utility_pods, node=worker_node1).exec(
+                command=(
+                    f"openssl s_client -showcerts -connect {svc_ip}:443 </dev/null 2>/dev/null | "
+                    "sed -n '/-----BEGIN/,/-----END/p'"
+                )
+            )
+        except workers_utility_pods.ExecOnPodError:
```
jpeimer, line 129: If I understand it correctly,  `ExecOnPodError` is not an attribute of `pod` instances. Should it be

```
from ocp_resources.exceptions import ExecOnPodError
  ...
  except ExecOnPodError:
```

jpeimer (changes requested):

jpeimer (approved):

PR #2091, opened 2025-09-17
Storage: fix concurrent uploads test for s390x.

vamsikrishna-siddu:
##### Short description:
The test_successful_concurrent_uploads is using the fedora qcow2 for s390x. The size of fedora is ~900mb and the test is timing out because it has to download the image from artifactory and create datavolumes concurrently.  The test is working fine in X86 because it is  using cirros which is ~12mb. I have replaced this test with alpine which will work on both x86 and s390x.  The alpine size ~ 80 mb. I have tested it on s390x and it is working fine.

## Summary by CodeRabbit

- **New Features**
  - Added support for Alpine VM images across architectures (new Alpine image variants available).

- **Tests**
  - Concurrent upload test now runs with Alpine image inputs to improve cross-platform coverage.

- **Chores**
  - Test data wiring updated to supply Alpine variants via fixtures; no public APIs or signatures changed.

### DIFF
--- a/libs/infra/images.py
+++ b/libs/infra/images.py
@@ -20,6 +20,14 @@ class Cirros:
 
 
 @dataclass
+class Alpine:
+    QCOW2_IMG: str | None = None
+    DIR: str = f"{BASE_IMAGES_DIR}/alpine-images"
+    DEFAULT_DV_SIZE: str = "1Gi"
+    DEFAULT_MEMORY_SIZE: str = "128M"
+
+
+@dataclass
 class Rhel:
     RHEL7_9_IMG: str | None = None
     RHEL8_0_IMG: str | None = None
--- a/tests/storage/cdi_upload/test_upload.py
+++ b/tests/storage/cdi_upload/test_upload.py
@@ -298,6 +298,18 @@ def _upload_image(dv_name, namespace, storage_class, local_name, size=None):
 @pytest.mark.sno
 @pytest.mark.s390x
 @pytest.mark.polarion("CNV-2015")
+@pytest.mark.parametrize(
+    "upload_file_path",
+    [
+        pytest.param(
+            {
+                "remote_image_dir": Images.Alpine.DIR,
+                "remote_image_name": Images.Alpine.QCOW2_IMG,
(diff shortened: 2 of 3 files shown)


tests/storage/cdi_upload/test_upload.py
```diff
@@ -296,8 +296,27 @@ def _upload_image(dv_name, namespace, storage_class, local_name, size=None):
...
+        pytest.param(
+            {
+                "remote_image_dir": Images.Cirros.DIR,
+                "remote_image_name": Images.Cirros.QCOW2_IMG,
+            },
+            marks=(pytest.mark.polarion("CNV-2015")),
+        ),
+    ],
+    indirect=True,
+)
```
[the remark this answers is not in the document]
vamsikrishna-siddu: I tried to use fedora . Eventhough i increased the timeout to 10mins the test is still failing because of big size. As discussed I am going with alpine as it is also less size. i have uploaded the alpine qcow2 files to artifactory for x86,arm and s390x. I have ran this for s390x and it is working with existing timeout of 1min.

libs/infra/images.py
```diff
@@ -19,6 +19,20 @@ class Cirros:
...
 
+@dataclass
+class Alpine:
+    RAW_IMG: str | None = None
+    RAW_IMG_GZ: str | None = None
+    RAW_IMG_XZ: str | None = None
+    QCOW2_IMG: str | None = None
+    QCOW2_IMG_GZ: str | None = None
+    QCOW2_IMG_XZ: str | None = None
+    DISK_DEMO: str | None = None
```
jpeimer, line 30: With this PR, you only add the `QCOW2_IMG`, right? So we can remove the rest for now?

```suggestion
    QCOW2_IMG: str | None = None
```

jpeimer (changes requested):
Thanks for the PR! Just a small suggestion

libs/infra/images.py
vamsikrishna-siddu: sure

jpeimer (approved):

PR #2113, opened 2025-09-18
WIP:  Update data_volume_template_with_source_ref_dict

RoniKishner:
##### Short description:
Update data_volume_template_with_source_ref_dict so you can set an empty storage class

##### More details:
 - Previously data_volume_template_with_source_ref_dict would always set the field `storageClassName`, this prevent from us from running test cases where there is storage class defined.
This PR update the function so it is possible to define `None` in the storage class
 - removes the QUARANTINED state of test_vm_from_data_source_missing_default_storage_class
 - Change order of fixtures, this way the access_mode field for the VM object can be set before the default storage class is removed, thus preventing an error in `get_storage_configuration`

Fix quarantined test test_vm_from_data_source_missing_default_storage_class

##### jira-ticket:

https://issues.redhat.com/browse/CNV-68779

BLOCKED BY: https://issues.redhat.com/browse/CNV-74930

## Summary by CodeRabbit

* **Tests**
  * Standardized timeout behavior across tests to a consistent 1-minute timeout and improved timeout assertions and error detection.
  * Removed a previously expected test failure, replacing it with explicit timeout-based validation.

* **Chores**
  * Data volume templates and test helpers now accept and propagate an explicit storage class parameter.
  * Template logic no longer silently falls back to the source's storage class; provided value is used.

### DIFF
--- a/tests/infrastructure/golden_images/test_common_templates_data_volumes.py
+++ b/tests/infrastructure/golden_images/test_common_templates_data_volumes.py
@@ -2,10 +2,11 @@ import pytest
 from ocp_resources.data_source import DataSource
 from ocp_resources.virtual_machine_cluster_instancetype import VirtualMachineClusterInstancetype
 from pytest_testconfig import config as py_config
+from timeout_sampler import TimeoutExpiredError
 
 from tests.infrastructure.golden_images.constants import PVC_NOT_FOUND_ERROR
 from tests.os_params import FEDORA_LATEST, FEDORA_LATEST_LABELS, FEDORA_LATEST_OS
-from utilities.constants import OS_FLAVOR_FEDORA, QUARANTINED, U1_SMALL, Images
+from utilities.constants import OS_FLAVOR_FEDORA, TIMEOUT_1MIN, U1_SMALL, Images
 from utilities.storage import data_volume_template_with_source_ref_dict
 from utilities.virt import VirtualMachineForTests, running_vm
 
@@ -22,6 +23,7 @@ def vm_from_golden_image_multi_storage(
     namespace,
     golden_image_data_source_multi_storage_scope_function,
 ):
+    source_spec = golden_image_data_source_multi_storage_scope_function.source.instance.to_dict()["spec"]
     with VirtualMachineForTests(
         name="vm-from-golden-image",
         namespace=namespace.name,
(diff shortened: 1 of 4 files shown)


utilities/storage.py
```diff
@@ -569,15 +569,15 @@ def data_volume_template_dict(
...
+def data_volume_template_with_source_ref_dict(data_source, storage_class=...):
     source_dict = data_source.source.instance.to_dict()
     source_spec_dict = source_dict["spec"]
     dv = DataVolume(
         name=data_source.name,
         namespace=data_source.namespace,
         size=source_spec_dict.get("resources", {}).get("requests", {}).get("storage")
         or source_dict.get("status", {}).get("restoreSize"),
-        storage_class=storage_class or source_spec_dict.get("storageClassName"),
+        storage_class=storage_class if storage_class is not ... else source_spec_dict.get("storageClassName"),
```
jpeimer, line 580: Why don't we keep `storage_class=None` by default, and do here `storage_class=storage_class`?

tests/infrastructure/golden_images/test_common_templates_data_volumes.py
```diff
@@ -29,6 +30,9 @@ def vm_from_golden_image_multi_storage(
         vm_instance_type=VirtualMachineClusterInstancetype(name=U1_SMALL),
         data_volume_template=data_volume_template_with_source_ref_dict(
             data_source=golden_image_data_source_multi_storage_scope_function,
+            storage_class=golden_image_data_source_multi_storage_scope_function.source.instance.to_dict()["spec"].get(
+                "storageClassName"
```
rnetser: why `get` is needed? is there a case where `storageClassName` is missing?

tests/infrastructure/golden_images/test_common_templates_data_volumes.py
```diff
@@ -163,10 +163,11 @@ def test_vm_dv_with_different_sc(
     indirect=True,
 )
 def test_vm_from_data_source_missing_default_storage_class(
-    removed_default_storage_classes,
     fedora_vm_from_data_source,
+    removed_default_storage_classes,
 ):
-    fedora_vm_from_data_source.start()
+    with pytest.raises(TimeoutExpiredError):
+        fedora_vm_from_data_source.start(timeout=TIMEOUT_1MIN, wait=True)
```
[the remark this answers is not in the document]
RoniKishner: You are right, we need to create a new DV first, so there will be an image to point to.

jpeimer (changes requested):

PR #2121, opened 2025-09-19
tests, net, bgp: add connectivity test suite

servolkov:
##### Short description:
This PR adds initial tests for BGP connectivity: initial BGP testing includes checking the tcp connection between the CUDN VM and the external network, as well as this connection preservation during VM migration. In this PR:
1) `udn_vm` function has been moved to `udn.py` lib file;
2) added traffic generator client functionality for pod;
3) added new fixtures and connectivity BGP tests.

##### jira-ticket: https://issues.redhat.com/browse/CNV-66592


libs/net/udn.py
```diff
@@ -24,3 +27,18 @@ def create_udn_namespace(
...
+    spec.template.spec.domain.devices.interfaces = [iface]  # type: ignore
+    spec.template.spec.networks = [network]
+    if template_labels:
+        spec.template.metadata.labels = spec.template.metadata.labels or {}  # type: ignore
+        spec.template.metadata.labels.update(template_labels)  # type: ignore
+        # Use the first label key and first value as the anti-affinity label to use:
+        label, *_ = template_labels.items()
+        spec.template.spec.affinity = new_pod_anti_affinity(label=label)
+
+    return fedora_vm(namespace=namespace_name, name=name, spec=spec)
```
rnetser, line 44: in a different PR - fedora_vm needs to get client

tests/network/bgp/test_bgp_connectivity.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+from libs.net.traffic_generator import is_tcp_connection
+from utilities.virt import migrate_vm_and_verify
+
+pytestmark = [pytest.mark.bgp, pytest.mark.usefixtures("bgp_setup_ready")]
```
rnetser, line 6: `pytest.mark.bgp` should be commented out and tests should be quarantined, right?

libs/net/traffic_generator.py
```diff
@@ -115,5 +116,51 @@ def _is_process_running(  # type: ignore[return]
...
+        self._ensure_is_running()
+
+        return self
+
+    def __exit__(self, exc_type: BaseException, exc_value: BaseException, traceback: object) -> None:
+        self._pod.execute(
+            command=["pkill", "-f", self._cmd],
+        )
+
+    def is_running(self) -> bool:
```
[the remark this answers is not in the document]
rnetser: `server.is_running()` -  what is running - the pod? a service? could be anything

tests/network/bgp/test_bgp_connectivity.py
rnetser: let's leave this comment unresolved then till this change is added

libs/net/udn.py
```diff
@@ -24,3 +27,18 @@ def create_udn_namespace(
         labels={"k8s.ovn.org/primary-user-defined-network": "", **(labels or {})},
         admin_client=client,
     )
+
+
+def udn_vm(namespace_name: str, name: str, template_labels: dict | None = None) -> BaseVirtualMachine:
+    spec = base_vmspec()
+    iface, network = udn_primary_network(name="udn-primary")
+    spec.template.spec.domain.devices.interfaces = [iface]  # type: ignore
```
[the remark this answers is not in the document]
servolkov: If remove ignore typing then `mypy` error:
```
libs\net\udn.py:35: error: Item "None" of "Devices | None" has no attribute "interfaces"  [union-attr]
Found 1 error in 1 file (checked 364 source files)

```

tests/network/bgp/test_bgp_connectivity.py
```diff
@@ -0,0 +1,26 @@
...
+    pytest.mark.usefixtures("bgp_setup_ready")
+]
+
+
+@pytest.mark.polarion("CNV-12276")
+def test_connectivity_cudn_vm_and_external_network(
+    tcp_server_cudn_vm,
+    tcp_client_external_network
+):
+    assert is_tcp_connection(server=tcp_server_cudn_vm, client=tcp_client_external_network)
```
EdDev: I do not like tests with just assertions... but it does look more informative this way.
I think that if there is a failure in the connection, it will not reach the body and fail in the fixture setup at the client establishment.

tests/network/libs/bgp.py
```diff
@@ -217,6 +222,7 @@ def deploy_external_frr_pod(
         containers=containers,
         volumes=volumes,
         client=client,
+        share_process_namespace=True,
```
EdDev, line 225: Why is this required? It is an extreme and should be rarely used.

libs/net/traffic_generator.py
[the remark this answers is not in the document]
yossisegev: I agree with @servolkov here. What we care about here is assuring the TCP entity (client or server) is running, regardless of which k8s resource is used to run it.

tests/network/bgp/test_bgp_connectivity.py
servolkov: >I do not like tests with just assertions...

me either. But it's really hard to come up with test steps here and not duplicate code then. 

>I think that if there is a failure in the connection, it will not reach the body and fail in the fixture setup at the client establishment.

I thing you are right. In test design we have 2 tests, the first one simply test tcp connectivity. Based on the implementation, this check is already inside the second test with migration. So, we can:

1. not to use `iperf3` client\server to check simple tcp connectivity and implement something simple for such purpose
2. remove this test at all, but we will not have a beautiful report (one test instead of two)
3. something else?

btw, why this comment is outdated?

servolkov:
rebase from main

libs/net/udn.py
```diff
@@ -24,3 +27,18 @@ def create_udn_namespace(
         labels={"k8s.ovn.org/primary-user-defined-network": "", **(labels or {})},
         admin_client=client,
     )
+
+
+def udn_vm(namespace_name: str, name: str, template_labels: dict | None = None) -> BaseVirtualMachine:
```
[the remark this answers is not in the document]
EdDev: The promoted function moved from a specific test file/package to a more general library.
Therefore, there is a need to generalize it so it can be re-used and make sense to a wider potential use scenarios.

libs/net/traffic_generator.py
```diff
@@ -115,5 +116,51 @@ def _is_process_running(  # type: ignore[return]
         return False
 
 
-def is_tcp_connection(server: Server, client: Client) -> bool:
+class PodClient:
```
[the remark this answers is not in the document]
EdDev: The naming you can play with in a temporary manner... e.g. leabe the current with the same name and use the abstract with a name like `BaseClient`. In a follow up PR you can change the names.
You can also have the correct name and just alias it, e.g. `Client(VMClient)` with no content (`pass`).

libs/net/traffic_generator.py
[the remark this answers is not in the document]
EdDev: > What is running?

The client or server, i.e. the object represented by the class.

tests/network/bgp/conftest.py
```diff
@@ -12,7 +13,11 @@
 from ocp_resources.pod import Pod
 
 from libs.net import netattachdef as libnad
-from libs.net.udn import create_udn_namespace
+from libs.net.traffic_generator import PodClient as TcpClient
+from libs.net.traffic_generator import Server as TcpServer
```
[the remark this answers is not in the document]
EdDev: Right, this is why you better leave the module in order to have context.
`treaffic_generator.Client` is clearer.

tests/network/bgp/conftest.py
```diff
@@ -179,6 +186,8 @@ def frr_external_pod(
         default_route=br_ex_gateway_v4,
         client=admin_client,
     ) as pod:
+        # Assign a secondary IP to emulate the external subnet
+        pod.execute(command=shlex.split(f"ip addr add {EXTERNAL_IP_V4} dev net1"), container="frr")
```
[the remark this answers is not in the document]
EdDev: In a regular pod, we do not define IP addresses by accessing its terminal.
We define it by IPAM in the network annotation (multus CNI pod specification).

You can pass a specific IP is you want through an optional IPs argument (I suggest a list).

tests/network/bgp/conftest.py
```diff
@@ -191,3 +200,35 @@ def bgp_setup_ready(
 ) -> None:
     node_names = [worker.name for worker in workers]
     wait_for_bgp_connection_established(node_names=node_names)
+
+
+@pytest.fixture(scope="module")
+def vm_cudn(
```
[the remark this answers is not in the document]
EdDev: CUDN us an implementation detail which is not important. It can be CUDN, UDN and even the NAD directly. What is important is that the VM is connected to a UDN type network, not how.

tests/network/bgp/test_bgp_connectivity.py
EdDev: I think that if you want to keep a dedicated test for the tcp connectivity, you should have something like this: https://github.com/RedHatQE/openshift-virtualization-tests/blob/d19438deb6f6e8b4eea7fbbf197327455aa43927/tests/network/localnet/test_default_bridge.py#L29
I.e. the connection is established in the body.

The other test can assume a connection is there already and then migrate (in that case, having the fixture makes sense).

WDYT?

tests/network/libs/bgp.py
EdDev: IMO you should be able to exec into the correct container. If the current helpers cannot support it, we should add support for it.
Because it is a test, I guess the security concern is not high, but it is a but ugly.

We can take it in a follow up if you see it is not possible to exec into the right container.

tests/network/bgp/conftest.py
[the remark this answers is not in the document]
servolkov: we discussed it in person, but the implementation implies using cudn. I just see the clarity and alignment here (prerequisite for the tests: vm is connected to cudn -> `vm_cudn`). I still think it's quite straitforward. But if you think my mind is broken - I can rename it (I guess we need to rename namespace fixture as well, and some others).

tests/network/libs/vm.py
EdDev, line 1: How about we call it `vm_factory`?

libs/net/traffic_generator.py
```diff
@@ -115,5 +136,49 @@ def _is_process_running(  # type: ignore[return]
...
+
+    def is_running(self) -> bool:
+        out = self._pod.execute(command=["pgrep", "-f", self._cmd], ignore_rc=True)
+        return bool(out.strip())
+
+    @retry(wait_timeout=30, sleep=2, exceptions_dict={ProcessLookupError: []})
+    def _ensure_is_running(self) -> bool:
+        if self.is_running():
+            return True
+        raise ProcessLookupError(f"{_IPERF_BIN} client process did not start in the pod {self._pod.name}")
```
EdDev, line 180: Can't this be a simple `return self.is_running()` ?
The exception dict should then be an empty dict.

BTW, can you on a follow up PR adjust the VM version to use `retry` as well?
I see there are more things to fix there (e.g. return type warning workarounds).

tests/network/bgp/conftest.py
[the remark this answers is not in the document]
EdDev: Ok, lets do it in a follow up.

tests/network/bgp/conftest.py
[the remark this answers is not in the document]
EdDev: Fine, I won't insist on this further.

tests/network/libs/bgp.py
```diff
@@ -205,7 +210,12 @@ def deploy_external_frr_pod(
             "image": _EXTERNAL_FRR_IMAGE,
             "securityContext": {"privileged": True, "capabilities": {"add": ["NET_ADMIN"]}},
             "volumeMounts": [{"name": frr_configmap_name, "mountPath": "/etc/frr"}],
-        }
+        },
+        {
+            "name": "iperf3",
+            "image": "networkstatic/iperf3",
```
EdDev, line 216: Are you sure this is working on CI? This is not a D/S image.
Please try to run this with Jenkins and see if it works.
I would also ask DevOps if this is ok. I am unsure if we are good with running U/S images on D/S infra.

EdDev (commented):
Thanks.
There are still some small open points, but we seem to be very close to finalize this one.

tests/network/libs/bgp.py
EdDev: I'm a bit worried, the image on both docker and quay are 1 year old with vulnerabilities showing on quay scan.
It is also not fixed on a specific image (sha), which may cause our tests to be sensitive to it changing.

Moreover, it seems to be based on Debian, which is a bit odd.

I think we need to find an official source or perhaps create one ourself.
Basing it on scratch to just deliver the binary file or on UBI or Fedora would be nicer.
Creating one ourself is a bit costly, but we can automate it easily.

Lets see if @rnetser and @myakove have any feedback on this.

EdDev: BTW, to build it seems straight forward: https://github.com/esnet/iperf/blob/master/.github/workflows/build.yml

servolkov: btw, we have many other iperf3 images in `quay.io`, the fresh one I like - https://quay.io/repository/skaplons/iperf3. As I see it uses `centos-stream-10` and just installs `iperf3` from the repo. It looks good, imho.

libs/net/traffic_generator.py
servolkov: > Because it does not make sense to return a boolean which only returns true, it is not symetric and smells something is off.

I do not completely agree. We need to return smth positive to stop retrying. `True` is a good relevant value (btw, we already did it in some other places in our codebase).  Hence, we cant returns "nothing". Then we should decide if we want just to return `False` in negative result or raise an exception. For the sake of explicitness, I prefer a clear exception. We cant log just a final message if eventually fail with retry (by retry design). If I want to see a reason of retry fail - I have to raise this exception.

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  You will create a change request comment on the 1st line of the 1st file in the pr with the test execution plan.

  **Analysis Requirements:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

  - Do not add a follow-up comment in the PR, only the change request one

</details>

libs/net/traffic_generator.py
```diff
@@ -1,8 +1,10 @@
 import logging
```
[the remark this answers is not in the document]
servolkov: verified by running `-m bgp`, `tests/network/user_defined_network/test_user_defined_network.py`, and  `tests/network/localnet` tests. All passed.

libs/net/traffic_generator.py
rnetser: i think `_wait_until_running` makes sense because this is what the function does

tests/network/bgp/test_bgp_connectivity.py
```diff
@@ -0,0 +1,29 @@
...
+@pytest.mark.xfail(
+    reason=f"{QUARANTINED}: BGP test suite infra dependencies are not met, tracked in CNV-69734",
+    run=False,
+)
+def test_connectivity_is_preserved_during_cudn_vm_migration(
+    tcp_server_cudn_vm,
+    tcp_client_external_network,
+):
+    migrate_vm_and_verify(vm=tcp_server_cudn_vm.vm)
+    assert is_tcp_connection(server=tcp_server_cudn_vm, client=tcp_client_external_network)
```
rnetser: from qe perspective - this is wrong
either the connectivity is tested during migration (in which `wait_for_migration_success=False should be passed to `migrate_vm_and_verify` )
or it is post migration (and migration happens in test setup).
the way it is written now, the test is testing 2 things - migration and connectivity.

will not block on this as this pattern appears network tests but this is something that should be discussed

tests/network/libs/vm_factory.py
```diff
@@ -0,0 +1,21 @@
...
+from libs.net.udn import udn_primary_network
+from libs.vm.affinity import new_pod_anti_affinity
+from libs.vm.factory import base_vmspec, fedora_vm
+from libs.vm.vm import BaseVirtualMachine
+
+
+def udn_vm(namespace_name: str, name: str, template_labels: dict | None = None) -> BaseVirtualMachine:
+    spec = base_vmspec()
+    iface, network = udn_primary_network(name="udn-primary")
+    spec.template.spec.domain.devices.interfaces = [iface]  # type: ignore
```
rnetser: why are the ignored typing needed in this file?

libs/net/traffic_generator.py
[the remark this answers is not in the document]
EdDev: `PodClient` represents a client that lives in a Pod.
I find it very clear.

BTW the class was changed from the original thread, although that does not change the reasoning.

servolkov:
fixed comments from @EdDev

libs/net/traffic_generator.py
```diff
@@ -115,5 +136,47 @@ def _is_process_running(  # type: ignore[return]
...
+    def __exit__(self, exc_type: BaseException, exc_value: BaseException, traceback: object) -> None:
+        self._pod.execute(
+            command=["pkill", "-f", self._cmd],
+        )
+
+    def is_running(self) -> bool:
+        out = self._pod.execute(command=["pgrep", "-f", self._cmd], ignore_rc=True)
+        return bool(out.strip())
+
+    @retry(wait_timeout=30, sleep=2, exceptions_dict={})
```
rnetser: 1. if you want all exceptions, nothing should be passed
2. TimeoutSampler is setting it to `{Exception: []}` - did you test `exceptions_dict={}` ? 

```
self.exceptions_dict = exceptions_dict if exceptions_dict is not None else {Exception: []}
```

EdDev: It is correctly set now.
When an empty dict is set, it means that any exception will fail imidiatly and that is the desired thing here.
When nothing is set, it defaults to what you mentioned above and will mean that all exceptions are retried, which is not what we want.

Per the response [here](https://github.com/RedHatQE/openshift-virtualization-tests/pull/2121#discussion_r2445311343), it was tested.

libs/net/traffic_generator.py
```diff
@@ -115,5 +136,47 @@ def _is_process_running(  # type: ignore[return]
         return False
 
 
-def is_tcp_connection(server: Server, client: Client) -> bool:
+class PodClient(BaseClient):
```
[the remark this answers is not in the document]
rnetser: @servolkov please fix the naming in a follow up pr
the fact that this class is used as
```
from libs.net.traffic_generator import PodClient as TcpClient
```
shows that the name is not descriptive enough

tests/network/libs/bgp.py
```diff
@@ -19,8 +19,10 @@
 
 _BGP_ASN: Final[int] = 64512
 _EXTERNAL_FRR_IMAGE: Final[str] = "quay.io/frrouting/frr:9.1.2"
+_IPERF3_IMAGE: Final[str] = "quay.io/networkstatic/iperf3"
```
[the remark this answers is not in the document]
servolkov: will be resolved in a follow-up together with https://issues.redhat.com/browse/CNV-70956.

openshift-virtualization-qe-bot-2:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-2121.

openshift-virtualization-qe-bot-2:

PR #2139, opened 2025-09-23
feat: add OS_FLAVOR class variable to cirros class for better flexibility

vamsikrishna-siddu:
##### Short description:
The OS_FLAVOR_CIRROS variable is responsible for authentication of vm. Since s390x does not support cirros. We are using fedora and we are mapping OS_FLAVOR_CIRROS constant to fedora. But this is not working and the OS_FLAVOR_CIRROS is not getting populated with fedora for s390x in some test files. Due to this some of the tests are trying to login with cirros and they are failing. So I have created a os_flavor  class variable and initializing this variable based on ArchClass for better flexibility.

## Summary by CodeRabbit

* **New Features**
  * Cirros images now expose a concrete OS flavor value.

* **Bug Fixes**
  * Aligned OS flavor per architecture to prevent mismatches (e.g., s390x).

* **Refactor**
  * Consolidated OS flavor sourcing and removed the legacy global constant.

* **Tests**
  * Updated test suite to use the new Cirros OS flavor source and removed the obsolete override.


dshchedr (dismissed):

utilities/constants.py
```diff
@@ -55,6 +61,7 @@ class X86_64:  # noqa: N801
             QCOW2_IMG_GZ=f"{BASE_CIRROS_NAME}.qcow2.gz",
             QCOW2_IMG_XZ=f"{BASE_CIRROS_NAME}.qcow2.xz",
             DISK_DEMO="cirros-registry-disk-demo",
+            OS_FLAVOR=OS_FLAVOR_CIRROS,
```
rnetser, line 64: please add also under `ARM64`

rnetser (dismissed):

utilities/constants.py
vamsikrishna-siddu: done

RoniKishner:
Are the tests passing now? from what I understand in the code there should be no changes:
The default value of `VirtualMachineForTests` is `os_flavor=OS_FLAVOR_FEDORA`, and you set `OS_FLAVOR: str | None = None`, so I unless i'm missing something the tests should still get the same "fedora" value.

vsibirsk (approved):

vamsikrishna-siddu:
Verified by Running the  entire tier2 storage test suite for s390x and all the tier2 storage  tests are passing.

PR #2144, opened 2025-09-23
[CNV-66017]skip log files that were rotated on the OCP nodes

rlobillo:
##### Short description:

As a continuation of the work done in https://github.com/RedHatQE/openshift-virtualization-tests/pull/1506, add graceful handling for audit logs that have been rotated. It will return empty results to allow processing to continue when logs are unavailable due to rotation.

##### More details:
Observed on CI.

N/A
N/A
N/A
##### jira-ticket:
https://issues.redhat.com/browse/CNV-66017

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of audit log processing: when logs show errors alongside “404 page not found” entries (indicative of rotation), the system now skips processing and logs a warning instead of failing.
  * Users should see fewer false alerts and smoother audit runs.
  * No changes to workflows or settings required.

### DIFF
--- a/utilities/infra.py
+++ b/utilities/infra.py
@@ -1072,6 +1072,9 @@ def get_node_audit_log_entries(log, node, log_entry):
     ).splitlines()
     has_errors = any(line.startswith("error:") for line in lines)
     if has_errors:
+        if any(line.startswith("404 page not found") for line in lines):
+            LOGGER.warning(f"Skipping {log} check as it was rotated:\n{lines}")
+            return True, []
         LOGGER.warning(f"oc command failed for node {node}, log {log}:\n{lines}")
         raise RuntimeError
     return True, lines


kshvaika (approved):
/lgtm

utilities/infra.py
```diff
@@ -1072,6 +1072,9 @@ def get_node_audit_log_entries(log, node, log_entry):
     ).splitlines()
     has_errors = any(line.startswith("error:") for line in lines)
     if has_errors:
+        if any(line.startswith("404 page not found") for line in lines):
```
geetikakay, line 1075: hints missing for this function?
can we check the return code instead of checking "404 page not found"? checking return code gives reliable information

geetikakay: There are couple of things here 
1. I kind of avoid getoutput because it never clearly mentions if error is in stderr or stdout. Error should be in stderr only else i consider it as a bug in negative test cases .
2. getoutput never gives return code in some cases we might want to see it and if !=0 then only start checking if's.
3. there is a possibility that logs might not exist. maybe fetch first, check rc and then search in python

geetikakay: @rnetser should we consider replacing grep with python search buildin at later if fetching is successful.

rlobillo: The output message is a string  `404 page not found`, not a http response or something like that. This is what openshift responds to the oc command that tries to access to a file that does not exist.

I agree with @rnetser that it's not worth refactoring the entire test. We simply need to skip checking for a file that doesn't exist (because it was rotated since the file list was fetched) and have a suitable event to identify that situation: the '404 page not found'.

openshift-virtualization-qe-bot-3:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-2144.

openshift-virtualization-qe-bot-3:

PR #2148, opened 2025-09-23
Fix the valid_cdi_certificate fn to fetch utc time instead of localtime.

vamsikrishna-siddu:
##### Short description:
The Current implementation of valid_cdi_certificate assumes the local time on the server where we run the tests is utc. But the timezone can be different in machine we are running the tests. So I have changed the function to use utc timezone as the certificate before and after times are in utc.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected certificate validity checks to use timezone-aware UTC comparisons, avoiding incorrect expired/not-yet-valid results across time zones.

* **Tests**
  * Updated tests to use UTC-aware datetime parsing and comparisons for consistent, reliable validation behavior.

### DIFF
--- a/tests/storage/test_cdi_certificate.py
+++ b/tests/storage/test_cdi_certificate.py
@@ -7,7 +7,6 @@ Automatic refresh of CDI certificates test suite
 import datetime
 import logging
 import subprocess
-import time
 
 import pytest
 from ocp_resources.cdi import CDI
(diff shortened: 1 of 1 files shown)


rnetser (approved):
nit:
see CodeRabbit's suggestion

```
+                start_dt = datetime.datetime.strptime(start, RFC3339_FORMAT).replace(tzinfo=datetime.timezone.utc)
+                end_dt = datetime.datetime.strptime(end, RFC3339_FORMAT).replace(tzinfo=datetime.timezone.utc)
+                now_dt = datetime.datetime.now(datetime.timezone.utc)
+                assert start_dt <= now_dt <= end_dt, f"Certificate of {cdi_secret} not valid at current time"
```

PR #2240, opened 2025-09-30
Fix: Dynamically set CPU threads per architecture

aditi-sharma-1:
This PR introduces dynamic CPU thread configuration based on the cluster architecture.

Previously, VMs were always configured with two CPU threads. The s390x architecture supports only one CPU thread per core.

This change adds a helper function, `get_cpu_threads_arch()`, which sets the CPU threads as follows:

- Returns `ONE_CPU_THREAD` for s390x  
- Defaults to `TWO_CPU_THREADS` for other architectures  

The `vm_with_cpu_spec` fixture has been updated to use this dynamic value, improving cross-architecture compatibility and ensuring optimal CPU thread settings.

##### Short description:
Enable CPU thread configuration to be set dynamically per architecture.

##### More details:
- Introduces a helper function `get_cpu_threads_arch()` to return the correct CPU thread count based on cluster architecture.  
- Updates `vm_with_cpu_spec` to use this value.  
- Ensures that VM definitions remain valid and consistent across supported architectures (e.g., s390x, x86).  

Some architectures, such as s390x, support only one CPU thread per core.  
With this change, CPU threads are set dynamically based on the architecture, improving compatibility and ensuring VMs are configured correctly across different environments.

### DIFF
--- a/tests/observability/metrics/conftest.py
+++ b/tests/observability/metrics/conftest.py
@@ -36,6 +36,7 @@ from tests.observability.metrics.utils import (
 from tests.observability.utils import validate_metrics_value
 from tests.utils import create_vms
 from utilities import console
+from utilities.architecture import get_cluster_architecture
 from utilities.constants import (
     IPV4_STR,
     KUBEVIRT_VMI_MEMORY_PGMAJFAULT_TOTAL,
@@ -55,6 +56,7 @@ from utilities.constants import (
     TIMEOUT_15SEC,
     TWO_CPU_CORES,
     TWO_CPU_SOCKETS,
+    ONE_CPU_THREAD,
     TWO_CPU_THREADS,
     VIRT_TEMPLATE_VALIDATOR,
     Images,
@@ -95,6 +97,9 @@ METRICS_WITH_WINDOWS_VM_BUGS = [
     KUBEVIRT_VMI_MEMORY_PGMINFAULT_TOTAL,
 ]
 
+def get_cpu_threads_arch():
+    arch = get_cluster_architecture()
+    return ONE_CPU_THREAD if arch == "s390x" else TWO_CPU_THREADS
 
 @pytest.fixture(scope="module")
 def unique_namespace(admin_client, unprivileged_client):
@@ -276,12 +281,14 @@ def windows_vm_for_test_interface_name(windows_vm_for_test):
 @pytest.fixture(scope="class")
 def vm_with_cpu_spec(namespace, unprivileged_client):
     name = "vm-resource-test"
+    cpu_threads = get_cpu_threads_arch()  # dynamically set based on arch
+
     with VirtualMachineForTests(
         name=name,
         namespace=namespace.name,
         cpu_cores=TWO_CPU_CORES,
         cpu_sockets=TWO_CPU_SOCKETS,
-        cpu_threads=TWO_CPU_THREADS,
+        cpu_threads=cpu_threads,
         body=fedora_vm_body(name=name),
         client=unprivileged_client,
     ) as vm:


tests/observability/metrics/conftest.py
```diff
@@ -274,14 +276,17 @@ def windows_vm_for_test_interface_name(windows_vm_for_test):
 
 
 @pytest.fixture(scope="class")
-def vm_with_cpu_spec(namespace, unprivileged_client):
+def vm_with_cpu_spec(namespace, unprivileged_client, nodes_cpu_architecture):
     name = "vm-resource-test"
+    arch = nodes_cpu_architecture
```
rnetser: no need to assign a new variable, please use `nodes_cpu_architecture` directly

rnetser (changes requested):

tests/observability/metrics/conftest.py
```diff
@@ -274,14 +276,17 @@ def windows_vm_for_test_interface_name(windows_vm_for_test):
...
+    cpu_threads = ONE_CPU_THREAD if arch == S390X else TWO_CPU_THREADS  # dynamically set based on arch
```
dshchedr: I see the similar check `"if nodes_cpu_architecture == S390X"` in another files. If you are planning to add same in other places in future - pls consider creating a separate session scope fixture, something like `is_s390x_cluster()`

aditi-sharma-1: Hi @dshchedr, 
Thank you for the suggestion. I have added a fixture, is_s390x_cluster(), to handle this check, and the function is now using it. This should also make it easier to reuse in future tests.

dshchedr (approved):

RoniKishner:
> The `modified_vm_cpu_requests` fixture fetches the CPU spec from the created VM object and increments each parameter by 1. This only updates the Python dictionary representing the VM spec for `test_metric_kubevirt_vm_resource_requests` to validate Prometheus metrics; it does not modify the running VM. So while the dictionary shows an increment in CPU threads, the VM itself is still created with threads=1 on s390x, avoiding boot issues.

Your comment is wrong, `modified_vm_cpu_requests` uses `ResourceEditor` which updated the VM resource, not just the python dictionary (just as it's naming indicates) 
I am not against changing the test to use 1 CPU thread if it is needed for the VM to run, but I don't like the idea of using `is_s390x_cluster` to decide on the number of CPU threads, you should consider changing it to have a permanent value of 1.

@vsibirsk was there a reason to use the value of 2, or can we change it to 1 for all arch?

chandramerla:
> > The `modified_vm_cpu_requests` fixture fetches the CPU spec from the created VM object and increments each parameter by 1. This only updates the Python dictionary representing the VM spec for `test_metric_kubevirt_vm_resource_requests` to validate Prometheus metrics; it does not modify the running VM. So while the dictionary shows an increment in CPU threads, the VM itself is still created with threads=1 on s390x, avoiding boot issues.
> 
> Your comment is wrong, `modified_vm_cpu_requests` uses `ResourceEditor` which updated the VM resource, not just the python dictionary (just as it's naming indicates) I am not against changing the test to use 1 CPU thread if it is needed for the VM to run, but I don't like the idea of using `is_s390x_cluster` to decide on the number of CPU threads, you should consider changing it to have a permanent value of 1.
> 
> @vsibirsk was there a reason to use the value of 2, or can we change it to 1 for all arch?

@RoniKishner @vsibirsk Do you think we should change it to 1 CPU thread for all archs? Or merge the PR as it is?
P.S: @aditi-sharma-1 is on long leave. So, if that change is required, one of us can do.

rnetser (commented):
@aditi-sharma-1 where are all the other changes?

rnetser:
/hold

requested in https://github.com/RedHatQE/openshift-virtualization-tests/pull/2240#issuecomment-3642663911

PR #2265, opened 2025-10-05
[IUO] add jira cnv-70252

hmeir:
##### Short description:
Add specific skip for hco-bearer-auth resource, who might be missing from hco-related-objects because of this bug https://issues.redhat.com/browse/CNV-70252

##### jira-ticket:


tests/install_upgrade_operators/conftest.py
```diff
@@ -212,6 +213,8 @@ def related_object_from_hco_status(hco_status_related_objects, cnv_related_objec
     for obj in hco_status_related_objects:
         if obj.name == related_object_name and obj.kind == kind_name:
             return obj
+    if related_object_name == HCO_BEARER_AUTH and is_jira_open(jira_id="CNV-70252"):
+        pytest.xfail(f"Skipping related object {related_object_name}, kind {kind_name} because of CNV-70252")
```
dshchedr, line 217: this if should be before the for loop

no point for going thru the for loop if you know in advance that the object is not present

dshchedr (changes requested):

PR #2370, opened 2025-10-23
localnet: Add new scenario verifying IP in VMI when defining a localnet network with no VLAN

Anatw:
##### Short description:
Add new localnet scenario. This require changes in the existing codebase, splitted into three commits.
commit 1: Make CUDN VLAN optional: Localnet.vlan accepts None; localnet_cudn(vlan_id: int | None).
Commit 2: 
 - Refactor tests: move localnet fixtures to scope="class" and group existing tests under TestLocalnetVlan (no logic change).
 - Add no-VLAN scenario: new fixtures (cudn_localnet_no_vlan, vm_localnet_no_vlan) and TestLocalnetNoVlan asserting VMI reports IPv4 on LOCALNET_BR_EX_NETWORK.

##### More details:
VLAN is part of the Localnet class specifically because localnet networks connect directly to physical network interfaces on the nodes, where VLANs are applicable for network segmentation.
When vlan is specified, the localnet uses the specified VLAN ID.

This additional test scenario is based on a bug - https://issues.redhat.com/browse/CNV-38477

## Summary by CodeRabbit

* **New Features**
  * LocalNet can be provisioned without VLANs (VLAN optional).
  * Multi‑network / multi‑interface support with explicit cloud‑init network data for multi‑NIC VMs.
  * Primary/interface naming introduced for bridge/OVS VM scenarios.

* **Tests**
  * New fixtures and tests validate no‑VLAN secondary interfaces, secondary IP allocation, and IP reporting for dual‑interface VMs; existing bridge tests use new interface names.

* **Chores**
  * Volume handling improved to ensure disks are created when volumes are added; legacy network-wiring helper removed.

### DIFF
--- a/tests/network/localnet/conftest.py
+++ b/tests/network/localnet/conftest.py
@@ -9,12 +9,18 @@ import tests.network.libs.nodenetworkconfigurationpolicy as libnncp
 from libs.net.traffic_generator import TcpServer
 from libs.net.traffic_generator import VMTcpClient as TcpClient
 from libs.net.vmspec import lookup_iface_status
+from libs.vm.spec import Interface, Multus, Network
 from libs.vm.vm import BaseVirtualMachine
+from tests.network.libs import cloudinit
 from tests.network.libs import cluster_user_defined_network as libcudn
 from tests.network.libs.ip import random_ipv4_address
 from tests.network.localnet.liblocalnet import (
     LINK_STATE_DOWN,
+    LOCALNET_BR_EX_INTERFACE,
+    LOCALNET_BR_EX_INTERFACE_NO_VLAN,
     LOCALNET_BR_EX_NETWORK,
+    LOCALNET_BR_EX_NETWORK_NO_VLAN,
+    LOCALNET_OVS_BRIDGE_INTERFACE,
     LOCALNET_OVS_BRIDGE_NETWORK,
     LOCALNET_TEST_LABEL,
     client_server_active_connection,
@@ -31,6 +37,7 @@ from utilities.infra import create_ns
 from utilities.virt import migrate_vm_and_verify
 
 NNCP_INTERFACE_TYPE_OVS_BRIDGE = "ovs-bridge"
+PRIMARY_INTERFACE_NAME = "eth0"
 
 
 @pytest.fixture(scope="module")
@@ -96,24 +103,60 @@ def cudn_localnet(
 
 
 @pytest.fixture(scope="module")
+def cudn_localnet_no_vlan(
+    namespace_localnet_1: Namespace,
+) -> Generator[libcudn.ClusterUserDefinedNetwork]:
+    with localnet_cudn(
+        name=LOCALNET_BR_EX_NETWORK_NO_VLAN,
+        match_labels=LOCALNET_TEST_LABEL,
+        physical_network_name=LOCALNET_BR_EX_NETWORK,
+    ) as cudn:
+        cudn.wait_for_status_success()
+        yield cudn
+
+
+@pytest.fixture(scope="module")
 def ipv4_localnet_address_pool() -> Generator[str]:
     return (f"{random_ipv4_address(net_seed=0, host_address=host_value)}/24" for host_value in range(1, 254))
+
+
+@pytest.fixture(scope="module")
+def vm_localnet_1_secondary_ip(ipv4_localnet_address_pool: Generator[str]) -> str:
+    return next(ipv4_localnet_address_pool)
 
 
 @pytest.fixture(scope="module")
 def vm_localnet_1(
     namespace_localnet_1: Namespace,
     ipv4_localnet_address_pool: Generator[str],
+    vm_localnet_1_secondary_ip: str,
     cudn_localnet: libcudn.ClusterUserDefinedNetwork,
+    cudn_localnet_no_vlan: libcudn.ClusterUserDefinedNetwork,
     unprivileged_client: DynamicClient,
 ) -> Generator[BaseVirtualMachine]:
+    """
+    Creates a VM with two interfaces:
+    - Primary interface (eth0): connected to VLAN-enabled localnet
+    - Secondary interface (eth1): connected to no-VLAN localnet
+    """
     with localnet_vm(
         namespace=namespace_localnet_1.name,
         name="test-vm1",
-        physical_network_name=cudn_localnet.name,
-        spec_logical_network=LOCALNET_BR_EX_NETWORK,
-        cidr=next(ipv4_localnet_address_pool),
         client=unprivileged_client,
+        networks=[
+            Network(name=LOCALNET_BR_EX_INTERFACE, multus=Multus(networkName=cudn_localnet.name)),
+            Network(name=LOCALNET_BR_EX_INTERFACE_NO_VLAN, multus=Multus(networkName=cudn_localnet_no_vlan.name)),
+        ],
+        interfaces=[
+            Interface(name=LOCALNET_BR_EX_INTERFACE, bridge={}),
+            Interface(name=LOCALNET_BR_EX_INTERFACE_NO_VLAN, bridge={}),
+        ],
+        network_data=cloudinit.NetworkData(
+            ethernets={
+                PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)]),
+                "eth1": cloudinit.EthernetDevice(addresses=[vm_localnet_1_secondary_ip]),
+            }
+        ),
     ) as vm:
         yield vm
 
@@ -128,10 +171,12 @@ def vm_localnet_2(
     with localnet_vm(
         namespace=namespace_localnet_2.name,
         name="test-vm2",
-        physical_network_name=cudn_localnet.name,
-        spec_logical_network=LOCALNET_BR_EX_NETWORK,
-        cidr=next(ipv4_localnet_address_pool),
         client=unprivileged_client,
+        networks=[Network(name=LOCALNET_BR_EX_INTERFACE, multus=Multus(networkName=cudn_localnet.name))],
+        interfaces=[Interface(name=LOCALNET_BR_EX_INTERFACE, bridge={})],
+        network_data=cloudinit.NetworkData(
+            ethernets={PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)])}
+        ),
     ) as vm:
         yield vm
 
@@ -156,7 +201,7 @@ def localnet_client(localnet_running_vms: tuple[BaseVirtualMachine, BaseVirtualM
     with create_traffic_client(
         server_vm=localnet_running_vms[0],
         client_vm=localnet_running_vms[1],
-        spec_logical_network=LOCALNET_BR_EX_NETWORK,
+        spec_logical_network=LOCALNET_BR_EX_INTERFACE,
     ) as client:
         assert client.is_running()
         yield client
@@ -227,11 +272,14 @@ def vm_ovs_bridge_localnet_link_down(
     with localnet_vm(
         namespace=namespace_localnet_1.name,
         name="localnet-ovs-link-down-vm",
-        physical_network_name=cudn_localnet_ovs_bridge.name,
-        spec_logical_network=LOCALNET_OVS_BRIDGE_NETWORK,
-        cidr=next(ipv4_localnet_address_pool),
         client=unprivileged_client,
-        interface_state=LINK_STATE_DOWN,
+        networks=[
+            Network(name=LOCALNET_OVS_BRIDGE_INTERFACE, multus=Multus(networkName=cudn_localnet_ovs_bridge.name))
+        ],
+        interfaces=[Interface(name=LOCALNET_OVS_BRIDGE_INTERFACE, bridge={}, state=LINK_STATE_DOWN)],
+        network_data=cloudinit.NetworkData(
+            ethernets={PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)])}
+        ),
     ) as vm:
         yield vm
 
@@ -246,10 +294,14 @@ def vm_ovs_bridge_localnet_1(
     with localnet_vm(
         namespace=namespace_localnet_1.name,
         name="localnet-ovs-vm1",
-        physical_network_name=cudn_localnet_ovs_bridge.name,
-        spec_logical_network=LOCALNET_OVS_BRIDGE_NETWORK,
-        cidr=next(ipv4_localnet_address_pool),
         client=unprivileged_client,
+        networks=[
+            Network(name=LOCALNET_OVS_BRIDGE_INTERFACE, multus=Multus(networkName=cudn_localnet_ovs_bridge.name))
+        ],
+        interfaces=[Interface(name=LOCALNET_OVS_BRIDGE_INTERFACE, bridge={})],
+        network_data=cloudinit.NetworkData(
+            ethernets={PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)])}
+        ),
     ) as vm:
         yield vm
 
@@ -264,10 +316,14 @@ def vm_ovs_bridge_localnet_2(
     with localnet_vm(
         namespace=namespace_localnet_1.name,
         name="localnet-ovs-vm2",
-        physical_network_name=cudn_localnet_ovs_bridge.name,
-        spec_logical_network=LOCALNET_OVS_BRIDGE_NETWORK,
-        cidr=next(ipv4_localnet_address_pool),
         client=unprivileged_client,
+        networks=[
+            Network(name=LOCALNET_OVS_BRIDGE_INTERFACE, multus=Multus(networkName=cudn_localnet_ovs_bridge.name))
+        ],
+        interfaces=[Interface(name=LOCALNET_OVS_BRIDGE_INTERFACE, bridge={})],
+        network_data=cloudinit.NetworkData(
+            ethernets={PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)])}
+        ),
     ) as vm:
         yield vm
 
@@ -279,7 +335,7 @@ def ovs_bridge_localnet_running_vms_one_with_interface_down(
     vm1, vm2 = run_vms(vms=(vm_ovs_bridge_localnet_link_down, vm_ovs_bridge_localnet_1))
     lookup_iface_status(
         vm=vm_ovs_bridge_localnet_link_down,
-        iface_name=LOCALNET_OVS_BRIDGE_NETWORK,
+        iface_name=LOCALNET_OVS_BRIDGE_INTERFACE,
         predicate=lambda interface: "guest-agent" in interface["infoSource"]
         and interface["linkState"] == LINK_STATE_DOWN,
     )
@@ -310,7 +366,7 @@ def localnet_ovs_bridge_client(
     with create_traffic_client(
         server_vm=ovs_bridge_localnet_running_vms[0],
         client_vm=ovs_bridge_localnet_running_vms[1],
-        spec_logical_network=LOCALNET_OVS_BRIDGE_NETWORK,
+        spec_logical_network=LOCALNET_OVS_BRIDGE_INTERFACE,
     ) as client:
         assert client.is_running()
         yield client
@@ -321,7 +377,7 @@ def localnet_vms_have_connectivity(localnet_running_vms: tuple[BaseVirtualMachin
     with client_server_active_connection(
         client_vm=localnet_running_vms[0],
         server_vm=localnet_running_vms[1],
-        spec_logical_network=LOCALNET_BR_EX_NETWORK,
+        spec_logical_network=LOCALNET_BR_EX_INTERFACE,
     ):
         pass
 
--- a/tests/network/localnet/liblocalnet.py
+++ b/tests/network/localnet/liblocalnet.py
@@ -10,7 +10,7 @@ from libs.net.traffic_generator import VMTcpClient as TcpClient
 from libs.net.vmspec import IP_ADDRESS, add_volume_disk, lookup_iface_status
 from libs.vm.affinity import new_pod_anti_affinity
 from libs.vm.factory import base_vmspec, fedora_vm
-from libs.vm.spec import CloudInitNoCloud, Interface, Metadata, Network
+from libs.vm.spec import CloudInitNoCloud, Devices, Interface, Metadata, Network
 from libs.vm.vm import BaseVirtualMachine, cloudinitdisk_storage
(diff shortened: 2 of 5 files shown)


tests/network/localnet/conftest.py
```diff
@@ -83,12 +83,49 @@ def cudn_localnet(
...
+    with localnet_vm(
+        namespace=namespace_localnet_1.name,
+        name="localnet-no-vlan-vm",
+        physical_network_name=cudn_localnet_no_vlan.name,
+        spec_logical_network=LOCALNET_BR_EX_NETWORK,
+        cidr=vm_localnet_no_vlan_ip_address,
+    ) as vm:
+        vm.start()
+        vm.wait_for_ready_status(status=True)
+        vm.wait_for_agent_connected()
```
OhadRevah: why not using `running_vm`?

Anatw: Thanks for the question. Running_vm is performing several checks, not all are needed by the network tests. In the current scenario we need to make sure that the VM is ready (VMI is running) and that AgentConnected has succeeded.
While these checks do happen as part of running_vm, not using this function is reducing flakiness of depending on other un-needed checks inside this functionality.

yossisegev:
Thank you

tests/network/libs/cluster_user_defined_network.py
```diff
@@ -49,8 +49,8 @@ class Role(Enum):
 
     role: str
     physicalNetworkName: str  # noqa: N815
-    vlan: Vlan
     ipam: Ipam
+    vlan: Vlan | None = None
```
azhivovk: Can you please add to the commit description a documentation of why vlan is specifically configured here?

tests/network/localnet/conftest.py
EdDev: sig-network stopped using that (bad) helper some time ago and it is no longer accepted in our codebase.

Please resolve this thread.

rnetser: these checks appear multiple times in sig-network, you should centarlize under one function

tests/network/localnet/test_default_bridge.py
```diff
@@ -1,35 +1,51 @@
...
+        vms.remove(migrated_localnet_vm)
+        (base_localnet_vm,) = vms
+
+        with client_server_active_connection(
+            client_vm=base_localnet_vm,
+            server_vm=migrated_localnet_vm,
+            spec_logical_network=LOCALNET_BR_EX_NETWORK,
+            port=8888,
+        ) as (client, server):
+            assert is_tcp_connection(server=server, client=client)
```
rnetser: `is_tcp_connection` return a boolean if both client and server are running - to assert on it forcing a user to dig in the logs to look for the failure is not good. there should be a clear assertion, visible in test summary and report portal.


@EdDev

tests/network/localnet/test_default_bridge.py
```diff
@@ -1,35 +1,51 @@
...
+    @pytest.mark.ipv4
+    @pytest.mark.single_nic
+    @pytest.mark.s390x
+    @pytest.mark.usefixtures("nncp_localnet")
+    @pytest.mark.polarion("CNV-12363")
+    def test_vmi_reports_ip_on_localnet_without_vlan(self, vm_localnet_no_vlan, vm_localnet_no_vlan_ip_address):
+        assert (
+            get_vmi_ip_v4_by_name(vm=vm_localnet_no_vlan, name=LOCALNET_BR_EX_NETWORK)
+            == ip_interface(vm_localnet_no_vlan_ip_address).ip
+        )
```
rnetser: same here - seeing an assert that x != y means nothing and has no context (btw, neither get_vmi_ip_v4_by_name nor ip_interface print anything)


@EdDev

tests/network/localnet/conftest.py
EdDev: Attempts to do so failed, we agreed at that time to disagree :)
It can be opened again.

EdDev: I am unsure if there is one correct thing to do in this regard.
To run a VM one needs to 1) define it, 2) start it, 3) check the VM actually started at the domain level and finally 4) that the OS is up and running.
There are pros and cons to separate these steps and not combine them.

tests/network/localnet/test_default_bridge.py
EdDev: The assertion would express that an expected IP X is not what has been detected on the interface, providing the other IP that was seen or raising with an `IpNotFound` exception.

The only information which may be useful as an imidiate observation is the whole VMI interfaces status section. We can run simulations and examine this and ask the team what is missing.
@Anatw , can you post an output of different fail scenarios here? Then we will know if something is missing or not.

tests/network/localnet/test_default_bridge.py
```diff
@@ -5,31 +5,31 @@
...
-@pytest.mark.polarion("CNV-11775")
-def test_connectivity_over_migration_between_localnet_vms(localnet_server, localnet_client):
-    migrate_vm_and_verify(vm=localnet_client.vm)
-    assert is_tcp_connection(server=localnet_server, client=localnet_client)
+class TestLocalnetVlan:
+    @pytest.mark.gating
+    @pytest.mark.ipv4
+    @pytest.mark.single_nic
+    @pytest.mark.s390x
+    @pytest.mark.usefixtures("nncp_localnet")
```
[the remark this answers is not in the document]
Anatw: Absolutely

Anatw:
> I have a few issues with this new test suggestion and the other changes:
> 
> * I do not understand why you need to touch the existing fixtures and tests organization. The new fixtures created do not seem to use the old ones.
> * I think the bug treated is negligible and not worth spawning a VM to test it.
> * Based on latest findings about IP collisions, adding a new IP in the non-vlan LAN sounds risky and against the direction we aim at.
> 
> For the 2nd point, you can just add a 2nd secondary interface to one of the VMs and check it in an additional test. For the 3rd point, you should use the new added mechanism to randomize the IP addresses.

1. Since I wanted to use the existing NNCP created as base for all the scenario, I needed to separate the fixtures to module+class, so that the 2 existing VM's will be available for their scenarios, and the new VM - for the new case.
2. Agreed, but the the scenario require defining a UDN without a VLAN. I'll implement your suggestion about using the existing VM for that.
3. Sure, I'll change it.

tests/network/localnet/test_default_bridge.py
```diff
@@ -33,3 +40,20 @@ def test_connectivity_post_migration_between_localnet_vms(migrated_localnet_vm,
...
+def test_vmi_reports_ip_on_secondary_interface_without_vlan(
+    vm_localnet_1,
+    vm_localnet_1_primary_secondary_ips,
+):
+    """
+    Test that vm_localnet_1's secondary interface on a no-VLAN localnet
+    correctly reports the IP address for that interface.
+    """
+    _, secondary_ip = vm_localnet_1_primary_secondary_ips
+    assert get_vmi_ip_v4_by_name(vm=vm_localnet_1, name=LOCALNET_BR_EX_NETWORK_NO_VLAN) == ip_interface(secondary_ip).ip
```
EdDev: Will `lookup_iface_status` fit here?

EdDev (changes requested):

tests/network/localnet/conftest.py
```diff
@@ -93,24 +95,51 @@ def cudn_localnet(
...
+
 @pytest.fixture(scope="module")
 def ipv4_localnet_address_pool() -> Generator[str]:
-    return (f"10.0.0.{host_value}/24" for host_value in range(1, 254))
+    return (f"{random_ipv4_address(net_seed=0, host_address=host_value)}/24" for host_value in range(1, 254))
+
+
+@pytest.fixture(scope="module")
+def vm_localnet_1_primary_secondary_ips(ipv4_localnet_address_pool: Generator[str]) -> tuple[str, str]:
+    return next(ipv4_localnet_address_pool), next(ipv4_localnet_address_pool)
```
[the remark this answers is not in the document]
Anatw: The test needs the configured IP value for validation. Without this fixture, 
there's no source of truth to verify that the VMI reports the correct IP 
address. I've refactored it to only store the secondary IP, which is what 
the test validates.

tests/network/localnet/test_default_bridge.py
EdDev: The `infoSource` is the added value here as it provides the source of the IP address (is it from the guest-agent or from the pod/domain).

We should probably create some assertion helpers as a follow up.
These are probably missing all over the place.

tests/network/localnet/conftest.py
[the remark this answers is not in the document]
EdDev: Thank you for the explanation.
I understand, I do not like that the expected IP is passed through a fixture but I also do not have an immediate solution.
The current state in which you used only a single IP is good enough, thanks. You can resolve this.

tests/network/localnet/test_default_bridge.py
EdDev: `lookup_iface_status` fetches an interface status from a given VM.
I find it trivial that from the information received, an assertion can be done on one of its values.

tests/network/localnet/liblocalnet.py
```diff
@@ -56,55 +57,68 @@ def create_traffic_client(
...
+    interfaces: list[Interface],
+    network_data: cloudinit.NetworkData,
 ) -> BaseVirtualMachine:
     """
-    Create a Fedora-based Virtual Machine connected to a given localnet network with a static IP configuration.
+    Create a Fedora-based Virtual Machine connected to localnet network(s) with static IP configuration.
 
     The VM will:
-    - Attach to a Multus network using a bridge interface.
+    - Attach to one or more Multus networks using bridge interfaces.
```
EdDev: This is now redundant. It will attach the to provided interfaces/networks.

tests/network/localnet/liblocalnet.py
```diff
@@ -56,55 +57,68 @@ def create_traffic_client(
...
     - Apply a specific label for anti-affinity scheduling.
-    - Use cloud-init to configure a static IP address.
+    - Use cloud-init to configure static IP addresses.
```
EdDev: Not really. It will configure whatever is provided in the cloud-init spec. IP addresses is a possibility, not something this function knows anything about now.

* It is fine to leave examples about it in the example section.

tests/network/localnet/liblocalnet.py
```diff
@@ -56,55 +57,68 @@ def create_traffic_client(
...
     userdata = cloudinit.UserData(users=[])
     disk, volume = cloudinitdisk_storage(
         data=CloudInitNoCloud(
-            networkData=cloudinit.asyaml(no_cloud=netdata), userData=cloudinit.format_cloud_config(userdata=userdata)
+            networkData=cloudinit.asyaml(no_cloud=network_data),
+            userData=cloudinit.format_cloud_config(userdata=userdata),
         )
     )
-    vmi_spec = add_volume_disk(vmi_spec=vmi_spec, volume=volume, disk=disk)
+    add_volume_disk(vmi_spec=vmi_spec, volume=volume, disk=disk)
```
EdDev: Why do we have a change here?

tests/network/localnet/test_default_bridge.py
```diff
@@ -33,3 +40,22 @@ def test_connectivity_post_migration_between_localnet_vms(migrated_localnet_vm,
...
+
+
+@pytest.mark.ipv4
+@pytest.mark.single_nic
+@pytest.mark.s390x
+@pytest.mark.usefixtures("nncp_localnet")
+@pytest.mark.polarion("CNV-12363")
+def test_vmi_reports_ip_on_secondary_interface_without_vlan(
+    vm_localnet_1,
+    vm_localnet_1_secondary_ip,
```
EdDev: This info could have been fetched from the spec itself.
Said that, this could be done in a follow up.

tests/network/localnet/conftest.py
```diff
@@ -95,25 +99,64 @@ def cudn_localnet(
...
+    interfaces = [
+        Interface(name=LOCALNET_BR_EX_NETWORK, bridge={}),
+        Interface(name=LOCALNET_BR_EX_NETWORK_NO_VLAN, bridge={}),
+    ]
+    network_data = cloudinit.NetworkData(
+        ethernets={
+            PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)]),
+            "eth1": cloudinit.EthernetDevice(addresses=[vm_localnet_1_secondary_ip]),
+        }
+    )
```
EdDev: All these could be inlined. I see no special usefulness to have it with intermediate variables.

EdDev (changes requested):

tests/network/localnet/conftest.py
Anatw: Ack. It seems more readable that way but now that I see them inline it looks good as well.

tests/network/localnet/liblocalnet.py
Anatw: Ack, and actually this line is redundant as well. I'm removing it. I also fixed the fixture description accordingly.

tests/network/localnet/liblocalnet.py
```diff
@@ -56,55 +57,68 @@ def create_traffic_client(
...
     spec.template.metadata.labels.update(LOCALNET_TEST_LABEL)
     vmi_spec = spec.template.spec
 
-    vmi_spec = add_network_interface(
-        vmi_spec=vmi_spec,
-        network=Network(name=spec_logical_network, multus=Multus(networkName=physical_network_name)),
-        interface=Interface(name=spec_logical_network, bridge={}, state=interface_state),
-    )
+    for network, interface in zip(networks, interfaces):
+        add_network_interface(vmi_spec=vmi_spec, network=network, interface=interface)
```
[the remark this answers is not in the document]
Anatw: Good idea. I've added an explicit length check before the zip() call.
Using strict=True alone isn't sufficient because its error message doesn't include the actual data, just a general mismatch comment. The explicit check raises a ValueError with a clear message showing the actual networks and interfaces provided (e.g., "Networks: ['net1', 'net2'], Interfaces: ['net1', 'net2', 'net3']"), making debugging easier.

tests/network/localnet/liblocalnet.py
Anatw: Ack.

openshift-virtualization-qe-bot-3:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-2370 published

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  You will create a change request comment on the 1st line of the 1st file in the pr with the test execution plan.

  **Analysis Requirements:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

  - Do not add a follow-up comment in the PR, only the change request one

</details>

tests/network/localnet/conftest.py
```diff
@@ -125,13 +165,18 @@ def vm_localnet_2(
     cudn_localnet: libcudn.ClusterUserDefinedNetwork,
     unprivileged_client: DynamicClient,
 ) -> Generator[BaseVirtualMachine]:
+    networks = [Network(name=LOCALNET_BR_EX_NETWORK, multus=Multus(networkName=cudn_localnet.name))]
+    interfaces = [Interface(name=LOCALNET_BR_EX_NETWORK, bridge={})]
+    network_data = cloudinit.NetworkData(
+        ethernets={PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)])}
+    )
```
EdDev: This looks similar to the previous one [1], please follow the same pattern.|

Ref: https://github.com/RedHatQE/openshift-virtualization-tests/pull/2370#discussion_r2507756854

tests/network/localnet/conftest.py
```diff
@@ -243,13 +292,18 @@ def vm_ovs_bridge_localnet_1(
     cudn_localnet_ovs_bridge: libcudn.ClusterUserDefinedNetwork,
     unprivileged_client: DynamicClient,
 ) -> Generator[BaseVirtualMachine]:
+    networks = [Network(name=LOCALNET_OVS_BRIDGE_NETWORK, multus=Multus(networkName=cudn_localnet_ovs_bridge.name))]
+    interfaces = [Interface(name=LOCALNET_OVS_BRIDGE_NETWORK, bridge={})]
+    network_data = cloudinit.NetworkData(
+        ethernets={PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)])}
+    )
```
EdDev: This looks similar to the previous one [1], please follow the same pattern.|

Ref: https://github.com/RedHatQE/openshift-virtualization-tests/pull/2370#discussion_r2507756854

tests/network/localnet/conftest.py
```diff
@@ -261,13 +315,18 @@ def vm_ovs_bridge_localnet_2(
...
```
EdDev: This looks similar to the previous one [1], please follow the same pattern.|

Ref: https://github.com/RedHatQE/openshift-virtualization-tests/pull/2370#discussion_r2507756854

tests/network/localnet/liblocalnet.py
```diff
@@ -56,52 +57,70 @@ def create_traffic_client(
...
         namespace (str): The namespace where the VM should be created.
         name (str): The name of the VM.
-        physical_network_name (str): The name of the Multus network to attach.
-        cidr (str): The CIDR address to assign to the VM's interface.
         client (DynamicClient): The Kubernetes dynamic client for resource creation.
-        spec_logical_network (str): The name of the localnet network to attach.
-        interface_state (str): The state of the interface (optional).
-            Possible values are "up" or "down". When not specified, it behaves as "up".
+        networks (list[Network]): List of Network objects defining the networks to attach.
+            Each Network should have a name and Multus configuration.
```
EdDev: `Multus` is not part of this function action and/or assumptions. The caller defines what the network and interface contains.

tests/network/localnet/liblocalnet.py
```diff
@@ -56,52 +57,70 @@ def create_traffic_client(
...
+        interfaces (list[Interface]): List of Interface objects defining the interface configurations.
+            Each Interface should have a name matching a Network, bridge configuration, and optional state.
```
EdDev: Same, `bridge` should not be assumed.

tests/network/localnet/conftest.py
Anatw: Can I resolve this thread?

tests/network/localnet/test_default_bridge.py
Anatw: Ack. Marking this as resolved.

tests/network/localnet/conftest.py
Anatw: Can I resolve this?

tests/network/localnet/conftest.py
```diff
@@ -224,14 +269,18 @@ def vm_ovs_bridge_localnet_link_down(
     cudn_localnet_ovs_bridge: libcudn.ClusterUserDefinedNetwork,
     unprivileged_client: DynamicClient,
 ) -> Generator[BaseVirtualMachine]:
+    networks = [Network(name=LOCALNET_OVS_BRIDGE_NETWORK, multus=Multus(networkName=cudn_localnet_ovs_bridge.name))]
+    interfaces = [Interface(name=LOCALNET_OVS_BRIDGE_NETWORK, bridge={}, state=LINK_STATE_DOWN)]
+    network_data = cloudinit.NetworkData(
+        ethernets={PRIMARY_INTERFACE_NAME: cloudinit.EthernetDevice(addresses=[next(ipv4_localnet_address_pool)])}
+    )
```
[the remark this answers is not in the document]
Anatw: ack.

tests/network/localnet/conftest.py
Anatw: Ack.

tests/network/localnet/liblocalnet.py
Anatw: Ack.

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/71391bb1bec663b063a894288d8a74f934437f2a..66f7baf99d134c50164011b113174f7ce152c0a2): fix code rabbit concern about VLAN 0.

tests/network/localnet/liblocalnet.py
```diff
@@ -106,6 +106,7 @@ def localnet_vm(
 
     vmi_spec = spec.template.spec
     vmi_spec.networks = networks
+    vmi_spec.domain.devices = vmi_spec.domain.devices or Devices()
```
EdDev: Why do you need this?

tests/network/localnet/conftest.py
```diff
@@ -95,25 +99,61 @@ def cudn_localnet(
...
     with localnet_vm(
         namespace=namespace_localnet_1.name,
         name="test-vm1",
-        physical_network_name=cudn_localnet.name,
-        spec_logical_network=LOCALNET_BR_EX_NETWORK,
-        cidr=next(ipv4_localnet_address_pool),
         client=unprivileged_client,
+        networks=[
+            Network(name=LOCALNET_BR_EX_NETWORK, multus=Multus(networkName=cudn_localnet.name)),
+            Network(name=LOCALNET_BR_EX_NETWORK_NO_VLAN, multus=Multus(networkName=cudn_localnet_no_vlan.name)),
```
EdDev: I'm seeing this a bit late: It is confusing to use the same name for the logical network/interface and the CUDN name. Please use differentiate between them to make it clear these are not related.

tests/network/localnet/test_default_bridge.py
Anatw: Ack. I'll add this in a follow up PR.

tests/network/localnet/test_default_bridge.py
Anatw: May I resolve this thread?

tests/network/localnet/conftest.py
OhadRevah: Yes please

tests/network/localnet/liblocalnet.py
Anatw: To make pre-commit happy - vmi_spec.domain.devices may be None, but .interfaces was accessed without a check.

> tests/network/localnet/liblocalnet.py:109: error: Item "None" of "Devices | None" has no attribute "interfaces"  [union-attr]

tests/network/localnet/conftest.py
Anatw: Ack. I'm changing the OVS interface names at the same time even though it is not the scope of the current PR.

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/66f7baf99d134c50164011b113174f7ce152c0a2..8e30ccddbf45673d38cbc40cc88a3cd4244b52a8): fix according to CR:
 - Change commit message for both commits.
 - Have a network/interface name different from the CUDN object name.

openshift-virtualization-qe-bot-2:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-2370.

openshift-virtualization-qe-bot-2:

PR #2381, opened 2025-10-26
Add basic cx1 Numa test

geetikakay:
- Add basic NUMA tests (QoS, CPU manager selector, NUMA pinning, live migration)
- Move hugepages checks to shared conftest

##### Short description:
covers very basic tests that can be run on any numa capable node irrespective of it's size .This tests should work irrespective of topology policy.
##### jira-ticket:
https://issues.redhat.com/browse/CNV-31736

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive NUMA test suite for CX1 VMs: resource limits, QoS checks, CPU allocation, virt-launcher placement, and live migration.
  * Added utilities and fixtures to provision CX1 VMs and auto-select compatible instance types.
  * Added hugepage-detection fixtures that xfail/skip when 1Gi hugepages are absent and expose a session-wide hugepages cap.
  * Removed an older hugepage gating fixture (may require test adjustments).

### DIFF
--- a/tests/infrastructure/conftest.py
+++ b/tests/infrastructure/conftest.py
@@ -1,0 +1,26 @@
+import bitmath
+import pytest
+
+
+@pytest.fixture(scope="session")
+def hugepages_gib_values(workers):
+    """Return the list of hugepage sizes (in GiB) across all worker nodes."""
+    return [
+        int(bitmath.parse_string_unsafe(value).GiB)
+        for worker in workers
+        if (value := worker.instance.status.allocatable.get("hugepages-1Gi"))
+    ]
+
+
+@pytest.fixture(scope="session")
+def xfail_if_no_huge_pages(hugepages_gib_values):
+    """Mark tests as xfail if the cluster lacks 1Gi hugepages."""
+    if not hugepages_gib_values or max(hugepages_gib_values) < 1:
+        pytest.xfail("Requires at least 1Gi hugepages on some node")
+
+
+@pytest.fixture(scope="session")
+@pytest.mark.usefixtures("xfail_if_no_huge_pages")
+def hugepages_gib_max(hugepages_gib_values):
+    """Return the maximum 1Gi hugepage size, capped at 64Gi."""
+    return min(max(hugepages_gib_values), 64)
--- a/tests/infrastructure/numa/test_basic_numa.py
+++ b/tests/infrastructure/numa/test_basic_numa.py
@@ -1,0 +1,54 @@
+import pytest
+
+from tests.infrastructure.numa.utils import assert_qos_guaranteed
+from tests.utils import (
+    assert_numa_cpu_allocation,
+    assert_virt_launcher_pod_cpu_manager_node_selector,
+    get_numa_node_cpu_dict,
+    get_vm_cpu_list,
+)
+from utilities.virt import validate_libvirt_persistent_domain
+
+
+@pytest.mark.special_infra
+@pytest.mark.hugepages
+class TestBasicCx1Numa:
+    @pytest.mark.polarion("CNV-12364")
+    def test_numa_pod_resource_limits_match_requests(self, created_vm_cx1_instancetype):
+        container = created_vm_cx1_instancetype.vmi.virt_launcher_pod.instance.spec.containers[0]
+        limits = container.resources.limits or {}
+        requests = container.resources.requests or {}
+
+        mismatches = []
+        for key in ("cpu", "memory", "hugepages-1Gi"):  # ignoring devices and ephemeral-storage
+            if limits[key] != requests[key]:
+                mismatches.append(f"{key}: limit={limits[key]}, request={requests[key]}")
+        assert not mismatches, f"Mismatches found: {mismatches}"
+
+    @pytest.mark.polarion("CNV-12365")
+    def test_numa_pod_qos_class_guaranteed(self, created_vm_cx1_instancetype):
+        assert_qos_guaranteed(vm=created_vm_cx1_instancetype)
+
+    @pytest.mark.polarion("CNV-12366")
+    def test_numa_virt_launcher_pod_cpu_manager_node_selector(self, created_vm_cx1_instancetype):
+        assert_virt_launcher_pod_cpu_manager_node_selector(
+            virt_launcher_pod=created_vm_cx1_instancetype.vmi.virt_launcher_pod.instance
+        )
+
+    @pytest.mark.polarion("CNV-12367")
+    def test_numa_cpu_allocation(self, created_vm_cx1_instancetype):
+        assert_numa_cpu_allocation(
+            vm_cpus=get_vm_cpu_list(vm=created_vm_cx1_instancetype),
+            numa_nodes=get_numa_node_cpu_dict(vm=created_vm_cx1_instancetype),
+        )
+
+    @pytest.mark.polarion("CNV-12368")
+    def test_live_migrate_numa_vm(
+        self, skip_access_mode_rwo_scope_class, created_vm_cx1_instancetype, migrated_numa_cx1_vm
+    ):
+        validate_libvirt_persistent_domain(vm=created_vm_cx1_instancetype)
+        assert_qos_guaranteed(vm=created_vm_cx1_instancetype)
+        assert_numa_cpu_allocation(
+            vm_cpus=get_vm_cpu_list(vm=created_vm_cx1_instancetype),
+            numa_nodes=get_numa_node_cpu_dict(vm=created_vm_cx1_instancetype),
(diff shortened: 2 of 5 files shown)


tests/infrastructure/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+    values = get_hugepages_gib_values(workers=workers)
+    if not values or max(values) < 1:
+        pytest.xfail("Requires at least 1Gi hugepages on some node")
+
+
+@pytest.fixture(scope="session")
+def hugepages_gib_max(workers):
+    values = get_hugepages_gib_values(workers=workers)
+    # virtualmachineclusterinstancetype cx support maximum 64Gi
+    return min(max(values), 64)
```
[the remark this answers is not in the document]
geetikakay: fixed

tests/infrastructure/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_hugepages_gib_values(workers):
+    """Return list of hugepages sizes (in Gi) across all workers."""
+    values = []
+    for worker in workers:
+        if value := worker.instance.status.allocatable.get("hugepages-1Gi"):
+            gib = int(bitmath.parse_string(f"{value}B").GiB)
+            values.append(gib)
+    return values
```
vsibirsk: pls move to `utils.py`

tests/infrastructure/numa/test_basic_numa.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@pytest.fixture(scope="module")
+def created_vm_cx1_instancetype_profile(
+    unprivileged_client,
+    namespace,
+    golden_images_namespace,
+    modern_cpu_for_migration,
+    instance_type_rhel_os_matrix__module__,
+    storage_class_matrix__module__,
```
rnetser: for this feature, does it matter which sc is used?

tests/infrastructure/numa/test_basic_numa.py
```diff
@@ -0,0 +1,120 @@
...
+    pod = vm.vmi.virt_launcher_pod.instance
+    container = pod.spec.containers[0]
+    return {
+        "vm": vm,
+        "pod": pod,
+        "limits": container.resources.limits or {},
+        "requests": container.resources.requests or {},
+        "vm_cpus": get_vm_cpu_list(vm=vm),
+        "numa_nodes": get_numa_node_cpu_dict(vm=vm),
+    }
```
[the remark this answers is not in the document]
rnetser: +1
this fixture adds an overhead to understand the code - instead of accessing the needed values in the object, they are stored in a dict

tests/infrastructure/numa/test_basic_numa.py
```diff
@@ -0,0 +1,113 @@
...
+CX1_CLUSTER_INSTANCETYPE_MEMORY_SIZES = [2, 4, 8, 16, 32, 64]  # provided cluster cx1 instancetype sizes (Gi)
+
+
+def assert_qos_guaranteed(vm):
+    qos = vm.vmi.virt_launcher_pod.instance.status.qosClass
+    assert qos == "Guaranteed", f"QoS class is {qos}"
+
+
+@pytest.fixture(scope="module")
+def preferred_cluster_cx1_instance_type(admin_client, hugepages_gib_max):
```
vsibirsk: move to conftest

vsibirsk (changes requested):

tests/infrastructure/numa/test_basic_numa.py
```diff
@@ -0,0 +1,48 @@
...
+    @pytest.mark.polarion("CNV-12367")
+    def test_numa_cpu_allocation(self, created_vm_cx1_instancetype):
+        assert_numa_cpu_allocation(
+            vm_cpus=get_vm_cpu_list(vm=created_vm_cx1_instancetype),
+            numa_nodes=get_numa_node_cpu_dict(vm=created_vm_cx1_instancetype),
+        )
+
+    @pytest.mark.polarion("CNV-12368")
+    def test_live_migrate_numa_vm(self, skip_access_mode_rwo_scope_class, created_vm_cx1_instancetype):
+        migrate_vm_and_verify(vm=created_vm_cx1_instancetype, check_ssh_connectivity=True)
```
dshchedr: migration - should be a fixture

tests/infrastructure/utils.py
```diff
@@ -0,0 +1,12 @@
+from typing import Any, List
+
+import bitmath
+
+
+def get_hugepages_gib_values(workers: List[Any]) -> List[int]:
```
dshchedr: you use it in multiple places, better to make it as a session scope fixture
or at least add caching

geetikakay: get_hugepages_gib_values is called twice and at a session scope level(xfail_if_no_huge_pages and hugepages_gib_max) so it will be calculated twice.which is kind of ok'ish. 
Best solution would be to make it afixture so that it will be calculated once in session. what do you think. cache would be overhead in this case

tests/infrastructure/numa/test_basic_numa.py
dshchedr: In the test body should have only one action. In your test you have separate actions:
1. migration: `migrate_vm_and_verify()`
2. verification after migration: `validate_libvirt_`, assert_qos_`, `assert_numa_`

tests/infrastructure/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+@pytest.fixture(scope="session")
+def xfail_if_no_huge_pages(hugepages_gib_values):
+    """Mark tests as xfail if the cluster lacks 1Gi hugepages."""
+    if not hugepages_gib_values or max(hugepages_gib_values) < 1:
+        pytest.xfail("Requires at least 1Gi hugepages on some node")
+
+
+@pytest.fixture(scope="session")
+@pytest.mark.usefixtures("xfail_if_no_huge_pages")
+def hugepages_gib_max(hugepages_gib_values):
```
dshchedr:
```suggestion
def hugepages_gib_max(hugepages_gib_values, xfail_if_no_huge_pages):
```

tests/infrastructure/numa/test_basic_numa.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TestBasicCx1Numa:
+    @pytest.mark.polarion("CNV-12364")
+    def test_numa_pod_resource_limits_match_requests(self, created_vm_cx1_instancetype):
+        container = created_vm_cx1_instancetype.vmi.virt_launcher_pod.instance.spec.containers[0]
+        limits = container.resources.limits or {}
+        requests = container.resources.requests or {}
+
+        mismatches = []
+        for key in ("cpu", "memory", "hugepages-1Gi"):
```
rnetser: are these the only keys that exist under limits and requests? if so you can `requests == limits`

geetikakay: there are other keys too but their behavior doesn't impact us . Added the comment back which i accidentally removed

tests/infrastructure/conftest.py
dshchedr: `xfail_if_no_huge_pages` depends on `hugepages_gib_values`, so logically it should be called after

Also, I see you also still call `xfail_if_no_huge_pages` in old place:
https://github.com/RedHatQE/openshift-virtualization-tests/pull/2381/files#diff-f4f868da16a297cb9b2d2528aa94fed805889a495a4536e1eff09f028cd6f525R24

no point for having it here and there.. 

and one more - for hugepages we have a pytest mark `hugepages`, so probably you can add the sanity check - fail at the beginning of the test run if test mark "hugepages" selected, but the cluster does not have it.

tests/infrastructure/numa/test_basic_numa.py
dshchedr: migration - fixture
all validations in one test is ok for me

tests/infrastructure/conftest.py
geetikakay: https://issues.redhat.com/browse/CNV-71286 , with this in i can use markers and prevent calling these checks.

dshchedr: you can add marker now - it will help you to filter out all hugepages tests when they not needed

ksimon1:
@RoniKishner, @rnetser, @dshchedr would you please review this PR?

tests/infrastructure/conftest.py
geetikakay: added markers

dshchedr (approved):

RoniKishner (approved):

PR #2403, opened 2025-10-27
Remove irrelevant test from gating

yossisegev:
This is a customer-0 specific test, and doesn't belong as a gating test.


yossisegev:
> Can you please explain why is it no longer part of gating tests?

@azhivovk This test is relevant only for a specific customer, so even if it fails it only affects that customer and shouldn't fail the entire gating lane.

PR #2409, opened 2025-10-28
Storage: retry vm migration

dalia-frank:
##### Short description:
VMI migration retry in case of failures

##### More details:
As part of our efforts to reduce test flakiness and improve visibility of real bugs, we aim to minimize environment-related failures.

During VM live migrations, we sometimes see errors like the following:

failureReason: "Live migration failed error encountered during MigrateToURI3 libvirt api call: virError(Code=1, Domain=10, Message='internal error: process exited while connecting to monitor: qemu-kvm: -blockdev {\"driver\":\"host_device\",\"filename\":\"/var/run/kubevirt/hotplug-disks/blank-dv-ocs\",\"aio\":\"native\",\"node-name\":\"libvirt-1-storage\",\"auto-read-only\":true,\"discard\":\"unmap\",\"cache\":{\"direct\":true,\"no-flush\":false}}: Could not open '/var/run/kubevirt/hotplug-disks/blank-dv-ocs': Operation not permitted')"

These failures are not caused by the environment, but are transient, timing-related issues. Retrying the migration helps avoid these temporary failures and ensures that tests reflect real bugs rather than intermittent migration timing conditions.

## Summary by CodeRabbit

* **Improvements**
  * Added retry semantics and clearer timeout behavior for virtual machine migrations to improve reliability.
  * Performs explicit wait for migration completion and stronger failure reporting when timeouts occur.
  * Updated documentation and parameter descriptions for clearer usage and expected behavior.

### DIFF
--- a/utilities/virt.py
+++ b/utilities/virt.py
@@ -1848,7 +1848,7 @@ def migrate_vm_and_verify(
             if sample:
                 break
     except TimeoutExpiredError:
-        raise RuntimeError(f"Migration for VM {vm.vmi.name} failed after TimeoutSampler wait")
+        raise RuntimeError(f"Migration for VM {vm.vmi.name} failed")
 
     verify_vm_migrated(
         vm=vm,


utilities/virt.py
```diff
@@ -1811,43 +1811,52 @@ def migrate_vm_and_verify(
...
+                client=client,
+                namespace=vm.namespace,
+                vmi_name=vm.vmi.name,
+                teardown=wait_for_migration_success,
+            ) as migration:
+                if wait_for_migration_success:
+                    wait_for_migration_finished(namespace=vm.namespace, migration=migration, timeout=timeout)
+                return migration
+        except TimeoutExpiredError:
+            return False
```
jpeimer: The function needs to return `VirtualMachineInstanceMigration | None`, but now it can also return `False`

utilities/virt.py
```diff
@@ -1811,43 +1811,52 @@ def migrate_vm_and_verify(
...
+
+    if not wait_for_migration_success:
+        return _single_migration()
+
+    try:
+        for sample in TimeoutSampler(wait_timeout=timeout * 2, sleep=TIMEOUT_10SEC, func=_single_migration):
```
jpeimer: `_single_migration` calls `wait_for_migration_finished`, so it makes a `TimeoutSampler` call inside a `TimeoutSampler`. Please let's only have one TimeoutSampler at a time.

jpeimer (changes requested):
> Could not open '/var/run/kubevirt/hotplug-disks/blank-dv-ocs': Operation not permitted')"

Doesn't it look like a bug? Is it getting resolved by itself, and VMI eventually migrates? Do the users see the migration failure reason in the VMIM? How do they know what to do when they see it?

utilities/virt.py
```diff
@@ -1811,43 +1811,52 @@ def migrate_vm_and_verify(
     wait_for_migration_success: bool = True,
 ) -> VirtualMachineInstanceMigration | None:
     """
-    Create a migration instance. You may choose to wait for migration
-    success or not.
+    Migrate a VM with retry logic using TimeoutSampler, then verify migration success.
```
rnetser: this can hide bugs; we should not use retry on migration

rnetser (changes requested):

PR #2630, opened 2025-11-12
net: Update OWNERS file per active members

EdDev:

Update OWNERS file per active members


EdDev:
No changes to the codebase

/verify

rnetser:
this is an owners file, can be merged without the PR being labeled as verified

PR #2644, opened 2025-11-12
[4.18] net: quarantine localnet test_ovs_bridge

servolkov:
The test fails only in CI on a specific cluster. Need to investigate the cluster issue and then return the test back.


servolkov:
/verified

```
tests/network/localnet/test_ovs_bridge.py [NOTRUN] quarantined: fails in CI due to cluster issue; tracked in CNV-71535
TEST: test_connectivity_over_migration_between_ovs_bridge_localnet_vms STATUS: QUARANTINED
```

PR #2499, opened 2025-11-04
 Build fedora container images

ghost:
Updated GitHub actions to build fedora container images for all the architectures: x86_64, aarch64, s390x, and upload the images to quay

##### Short description:
Included the changes that build the fedora container images for all the architectures, push those images to quay

##### jira-ticket:
https://issues.redhat.com/browse/CNV-67468

## Summary by CodeRabbit

* **New Features**
  * Multi-architecture image publishing with unified multi-arch manifests (x86_64, aarch64, s390x)
  * PR-aware per-PR tagging, manifest inspection, and gated staging→production promotion

* **Chores**
  * CI moved to a gated pull-request-target flow with per-arch matrixed builds and podman-based authenticated pushes
  * Build flow standardized around a fedora-vars-driven per-arch configuration and consistent tagging; per-arch artifacts promoted into multi-arch manifests

* **Documentation**
  * README updated to describe the PR-driven multi-arch staging→production workflow

PR #2650, opened 2025-11-12
feat: s390x fedora image enablement

nestoracunablanco:
Commit 6
feat: Component Builder Publish s390x Enablement

This change enables the publication of the Fedora s390x image
alongside the amd64 image in a single manifest. To achieve this,
the following steps must be taken:

1. Install qemu-system-s390x.
2. Execute build.sh twice with different environment variables.

After completing these steps, both the amd64 and s390x images
should be added to the manifest.

Commit 5
Refactor: preparation work for Fedora image multiarch publish

The build.sh script primarily utilizes environment variables. To
generate an image for a new architecture, these variables must be
adjusted. This change relocates the generic environment variables
to specific steps, facilitating the transition to multiarch building.

Commit 4
feat: build s390x images in the component builder workflow

Most of the code is a copy-paste from the amd64 section. The main
difference is that qemu-system-s390x must be installed instead of x86.

After the job runs, an s390x artifact should be published.

Commit 3
fix: architecture suffix for the fedora-container-image artifact

This change introduces a suffix to the built artifact to facilitate
multi-architecture image testing. Without this modification, a race
condition occurs when the jobs for s390x and x86_64 run in parallel.

Commit 2
fix: allow build.sh to run multiple times

This change enables the script to be executed multiple times without
requiring a reboot. Previously, the virtual machine was undefined,
but the pool remained intact. When run a second time, no error was
thrown; however, the virsh-sysprep command experienced a timeout.

Additionally, the BUILD_DIR variable now includes an architecture
suffix. This enhancement eliminates the need to delete any folders
or residual data when building for different architectures.

Commit 1
fix: update boot device for s390x using zipl

This change in the user-data configuration is essential to ensure
that the Fedora image is bootable for the s390x architecture.

### DIFF
--- a/.github/workflows/component-builder-publish.yml
+++ b/.github/workflows/component-builder-publish.yml
@@ -24,17 +24,23 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y \
             qemu-system-x86 \
+            qemu-system-s390x \
             libvirt-daemon-system \
             virtinst cloud-image-utils \
             libguestfs-tools
       - name: Tweak hosted runner to enable 'virt-sysprep'
         # https://bugs.launchpad.net/ubuntu/+source/linux/+bug/759725
         run: sudo chmod 0644 /boot/vmlinuz*
-      - name: Fetch base Fedora image
+      - name: Fetch base Fedora amd64 image
         working-directory: ./containers/fedora
         run: wget -q "https://download.fedoraproject.org/pub/fedora/linux/releases/41/Cloud/x86_64/images/${{ env.FEDORA_IMAGE }}"
         env:
           FEDORA_IMAGE: Fedora-Cloud-Base-Generic-41-1.4.x86_64.qcow2
+      - name: Fetch base Fedora s390x image
+        working-directory: ./containers/fedora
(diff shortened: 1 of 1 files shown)


stesrn:
> Hey @rnetser, I've attached the draft for the comprehensive changes that enable the s390x Fedora build image. The plan is for each commit to be in a separate PR to facilitate the review process.
> 
> This may also be of interest to @dshchedr, @vsibirsk, @RoniKishner, @stesrn, and @yossisegev. @stern, I noticed you have PR #2499 open as well. What are your thoughts on this approach?

@nestoracunablanco hi. You are right, I am trying to automate the process of building fedora container image for all the three architectures ( x86_64, aarch64 and s390x ). When some PR touches the files under the directory cnv-tests/containers/fedora then the container images are built using matrix for all the architectures, images are built and pushed to fedora-staging quay repo, when the PR is merged the images are published in quay production repo. This is the entire logic that I tried automating, images are build successfully for all the 3 archs in my verification of PR. I can include the change that you have used in 'user-data' file. 

I feel this PR is redundant to build fedora container images. I request you to correct me, if I misunderstood the logic. Do you see any changes apart from #2499 included here ?

nestoracunablanco:
Hello @stesrn, I saw your PR and left some comments. If I have more time, I'll take a deeper look. There's a fix that needs to be addressed, but it only applies when you want to run the build.sh script multiple times. I will create a new PR for that.

PR #2721, opened 2025-11-18
WIP: Skip storage tests not suited for block storage class

ghost:
OCI supports block storage class: "oci-bv-uhp" only. So in this case, some storage tests that involves testing with DV content type as archive are not applicable. Also some tests that involves testing filesystem overhead are not applicable. This PR skips those tests for block storage class.

##### Short description:
Intention here is to skip the tests that are not relevant for block sc. 

##### More details:
OCI supports only block storage class. So DV.spec.contentType as archive is not valid also FS overhead tests are not valid. So these tests are skipped for OCI

This PR skips the tests that are not relevant for OCI env.

This ensures that the tests are skipped, rather than marked as ERROR or FAILED. This way QA engineer need not triage and review the failures once again.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-67047

## Summary by CodeRabbit

* **Tests**
  * Updated test infrastructure to include conditional test execution across multiple test modules. No changes to end-user functionality or behavior.

### DIFF
--- a/tests/storage/fs_overhead/test_fs_overhead.py
+++ b/tests/storage/fs_overhead/test_fs_overhead.py
@@ -87,9 +87,7 @@ def uploaded_cirros_dv(
 
 @pytest.mark.polarion("CNV-8635")
 def test_import_vm_with_specify_fs_overhead(
-    skip_block_volumemode_scope_module,
-    updated_fs_overhead_20_with_hco,
-    vm_for_fs_overhead_test
+    skip_block_volumemode_scope_module, updated_fs_overhead_20_with_hco, vm_for_fs_overhead_test
 ):
     vm_metadata = vm_for_fs_overhead_test.data_volume_template["metadata"]
     assert_fs_overhead_added(


tests/storage/fs_overhead/test_fs_overhead.py
```diff
@@ -86,7 +86,9 @@ def uploaded_cirros_dv(
 
 
 @pytest.mark.polarion("CNV-8635")
-def test_import_vm_with_specify_fs_overhead(updated_fs_overhead_20_with_hco, vm_for_fs_overhead_test):
+def test_import_vm_with_specify_fs_overhead(
+    skip_block_volumemode_scope_module, updated_fs_overhead_20_with_hco, vm_for_fs_overhead_test
+):
```
jpeimer: 1. We should not add new programmatic skips.
2. Every storage class supports Filesystem. Please pass `volume_mode=Filesystem` for the DVs in the tests that require Filesystem.

PR #2848, opened 2025-11-25
sanity-check: Fail if cluster is already tainted

rlobillo:
##### Short description:

Crypto policy and JSON patch tests fail systematically in teardown when the cluster has pre-existing jsonpatch annotations on HyperConverged CR that were not created by the tests themselves.

Furthermore, testing a cluster with unsupported configuration should not be valid for CI.

This PR is adding a sanity-check for this.

##### More details:

Tests assume clean cluster state but do not validate it before execution. When cluster has pre-existing jsonpatch annotations (from previous tests, manual changes, or incomplete cleanup), the cluster is already in TaintedConfiguration status.

includes a session-scoped fixture that will ensure that the HCO is not tainted before running the tests and, if that is not the case, it will fail fast.

Observed on CI, tests failing with below reason:
```
        AssertionError: assert not [TaintedConfiguration condition]
        lastTransitionTime: 2025-11-14T06:13:05Z
        message: Unsupported feature was activated via an HCO annotation
        reason: UnsupportedFeatureAnnotation
        status: True
        type: TaintedConfiguration
```

##### jira-ticket:
https://issues.redhat.com/browse/CNV-72688

## Tests
* Introduced cluster taint state validation that runs before test execution
* Applied validation to multiple test suites to verify cluster health and prevent tainted states from interfering with test operations

### DIFF
--- a/tests/install_upgrade_operators/conftest.py
+++ b/tests/install_upgrade_operators/conftest.py
@@ -22,7 +22,7 @@ from tests.install_upgrade_operators.utils import (
     get_resource_from_module_name,
 )
 from utilities.constants import HCO_BEARER_AUTH, HPP_POOL
-from utilities.hco import ResourceEditorValidateHCOReconcile, get_hco_version
+from utilities.hco import ResourceEditorValidateHCOReconcile, get_hco_version, is_hco_tainted
 from utilities.infra import (
     get_daemonset_by_name,
     get_deployment_by_name,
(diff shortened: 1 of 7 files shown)


rlobillo:
> @hmeir @rlobillo what is expected when not running in CI; i.e self-validation, is it possible that clusters running the tool will have hco tainted?

Agreed within the team to not block the testsuite run because of this. I have moved the check to the problematic tests so they will fail fast instead of failing on the teardown.

tests/install_upgrade_operators/conftest.py
```diff
@@ -39,6 +39,18 @@
 LOGGER = logging.getLogger(__name__)
 
 
+@pytest.fixture(scope="session")
+def validate_cluster_not_tainted(admin_client, hco_namespace):
```
rnetser: fixture names should be nouns and not verbs; i.e what is the outcome of the fixture or the required setup for the test 
in this case could be `is_cluster_tainted` or `untainted_cluster`

rnetser (changes requested):

PR #2890, opened 2025-11-27
Move `cache_admin_client` to be public

rnetser:
##### Short description:
Admin Client is used to CRUD cluster resources.
`admin_client` provides this client, and all functions calling ocp_resources resources must get a client as an attributes.
There are, however, places in the code (i.e in pytest-native fixtures or during pre-execution hooks) that the `admin_client` cannot be accessible.
While there are pytest hacks to use the fixture, we should take a simple, clear approach.

The usage must be very limited, the existing `_cache_admin_client` is now public to allow re-using it in places described above.

## Summary by CodeRabbit

* **Tests**
  * Expanded architecture detection tests with comprehensive coverage for multiple CPU architectures and node configurations.
  * Updated test infrastructure to support new client caching mechanism.

* **Chores**
  * Centralized client management through a new cached utility function for improved resource efficiency and consistency across codebase.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -21,11 +21,11 @@ from _pytest.nodes import Collector, Node
 from _pytest.reports import CollectReport, TestReport
 from _pytest.runner import CallInfo
 from kubernetes.dynamic.exceptions import ConflictError
-from ocp_resources.resource import get_client
 from pyhelper_utils.shell import run_command
 from pytest import Item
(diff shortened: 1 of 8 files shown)


rnetser:
/verified

build  3754 - smoke with all SC matrix options

vsibirsk (approved):

PR #2942, opened 2025-12-01
 infra/tests, net, nad: mandatory client for NAD

Anatw:
##### Short description:
This change makes the client mandatory for all NAD-related resources. Updated fixtures so that all calls to openshift-python-wrapper NAD resources explicitly pass a client.

While NADs are namespaced CRDs, OpenShift's default "admin" role only grants permissions for standard Kubernetes resources (pods, services, configmaps, etc.) and does not include permissions for any custom resources, even within a user's own namespace. Because network configuration can affect cluster-wide networking, NAD creation requires privileged clients by default.

As of its next releas, openshift-python-wrapper will enforce passing `client` when working with cluster resources.
openshift-virtualization-tests must align with this change.
All calls in the code to openshift-python-wrapper resources  should be updated to pass `client` arg.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-72392

## Summary by CodeRabbit

* **Refactor**
  * Network attachment client is now required, enforcing explicit admin authentication when creating network attachments.

* **Tests**
  * Many test fixtures and utilities updated to accept and forward an admin client so test network attachments are created with authenticated admin context.
  * Some fixtures add explicit cleanup steps after use.

### DIFF
--- a/libs/net/netattachdef.py
+++ b/libs/net/netattachdef.py
@@ -120,8 +120,8 @@ class NetworkAttachmentDefinition(NamespacedResource):
         name: str,
         namespace: str,
         config: NetConfig,
+        client: DynamicClient,
         resource_name: str | None = None,
-        client: DynamicClient | None = None,
     ):
         """
         Create and manage NetworkAttachmentDefinition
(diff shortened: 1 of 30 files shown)


EdDev (approved):
Thank you.

/approve

openshift-virtualization-qe-bot-3:

PR #3149, opened 2025-12-16
Alter chaos multiprocessing start method

sarahbx:
The previous fix of using get_context(method) is no longer working. To remedy this, switching to a fixture that sets the multiprocessing start method instead.
https://github.com/python/cpython/issues/132898
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process

## Summary by CodeRabbit

* **Tests**
  * Added a new fixture to chaos test suites to enforce a fork-based multiprocessing start method during test runs.
  * Enabled this fixture across migration, snapshot, and standard test modules to standardize behavior.
  * Updated test utilities to use the standard multiprocessing process construction for more consistent, more reliable test execution.

### DIFF
--- a/tests/chaos/conftest.py
+++ b/tests/chaos/conftest.py
@@ -406,3 +406,14 @@ def deleted_pod_by_name_prefix(admin_client, cnv_pod_deletion_test_matrix__class
         namespace=pod_deletion_config["namespace_name"],
         pod_prefix=pod_deletion_config["pod_prefix"],
     )
+
+
+@pytest.fixture(scope="module")
+def multiprocessing_start_method_fork():
+    # Use fork context to avoid pickling issues with nested functions
+    # https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process
+    # https://github.com/python/cpython/issues/132898
+    original_start_method = multiprocessing.get_start_method()
+    multiprocessing.set_start_method("fork", force=True)
+    yield
+    multiprocessing.set_start_method(original_start_method, force=True)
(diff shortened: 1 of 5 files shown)


tests/chaos/conftest.py
```diff
@@ -406,3 +406,14 @@ def deleted_pod_by_name_prefix(admin_client, cnv_pod_deletion_test_matrix__class
...
+
+@pytest.fixture(scope="module")
+def multiprocessing_start_method_fork():
+    # Use fork context to avoid pickling issues with nested functions
+    # https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process
+    # https://github.com/python/cpython/issues/132898
+    original_start_method = multiprocessing.get_start_method()
+    multiprocessing.set_start_method("fork", force=True)
+    yield
+    multiprocessing.set_start_method(original_start_method, force=True)
```
dshchedr: keep in mind, Multiprocessing was not working correctly with pytest and in some cases could freeze the testrun indefinitely. We have observed it multiple times (mostly in upgrade tests).

sarahbx: Understood. This is a workaround for the existing get_context() use-case. Ultimately all code in the repo that relies on the fork start method should consider a refactor.

> Changed in version 3.14: This is no longer the default start method on any platform. Code that requires fork must explicitly specify that via [get_context()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context) or [set_start_method()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method).

PR #3255, opened 2025-12-25
CherryPicked: [cnv-4.18][IUO] skip collection for simple tests to reduce time

hmeir:
##### Short description:
If the test is simple, and the fail message is clear enough, we shouldn't collect must gather on failure point. This significantly reduces execution time during
multiple test failures.
Cherry pick from https://github.com/RedHatQE/openshift-virtualization-tests/pull/3219
##### jira-ticket:


hmeir:
/verified

This is a cherry-pick performing marker addition only - if tox passing we are good to go.

PR #3328, opened 2026-01-06
Virt: replace get_client() usage in utilities/virt.py (part 1)

SiboWang1997:
##### Short description:
Replace the get_client() used in VirtualMachineForTestsFromTemplate object and ServiceForVirtualMachineForTests object with self.client

##### More details:
As drop get_client() in utilities/virt.py will lead to many relevant places change, this PR only updates the objects where get_client() can be directly replaced with self.client 
##### jira-ticket:

https://issues.redhat.com/browse/CNV-73714

## Summary by CodeRabbit

* **Refactor**
  * Switched VM-related operations to use instance-specific API clients rather than a global client, streamlining client context for virtual machine creation, template processing, and node/pod lookups. This reduces ambiguity in client usage and improves reliability of VM workflows across the platform.

### DIFF
--- a/utilities/virt.py
+++ b/utilities/virt.py
@@ -1096,7 +1096,7 @@ class VirtualMachineForTests(VirtualMachine):
 
     @property
     def privileged_vmi(self):
-        return VirtualMachineInstance(client=get_client(), name=self.name, namespace=self.namespace)
+        return VirtualMachineInstance(client=self.client, name=self.name, namespace=self.namespace)
 
     def wait_for_agent_connected(self, timeout: int = TIMEOUT_5MIN):
         self.vmi.wait_for_condition(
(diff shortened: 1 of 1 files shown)


utilities/virt.py
```diff
@@ -48,6 +48,7 @@
 import utilities.cpu
 import utilities.data_utils
 import utilities.infra
+from utilities.cluster import cache_admin_client
```
rnetser: do not use this function. see the function's docsting `This usage of this function is limited ONLY in places where `client` cannot be passed as an argument.`
add client arg where needed

SamAlber: @SiboWang1997 Just to double check, the `get_client` returns an admin client? From the wrapper it seems that it returns the client depending on the authentication credentials provided. 
Is this a performance improvement by using the cached variant?

PR #3416, opened 2026-01-13
Fix _sc_name_for_storage_api

RoniKishner:
##### Short description:
Fix _sc_name_for_storage_api

##### More details:
There are cases where the field `vm_preference` exists, but the preference does not have `preferredStorageClassName`. in such a case the storage class should be fetched from the cluster default storage class.
This fixes such an issue where the function return `None`, instead of moving to the next default option.

Fix error when preference exists but no storage class defined in it

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal fallback logic for storage selection to improve maintainability.
  * Added informational logging when the default storage option is used.
  * No changes to public behavior or exported interfaces; existing functionality remains unchanged.

### DIFF
--- a/utilities/virt.py
+++ b/utilities/virt.py
@@ -1016,8 +1016,9 @@ class VirtualMachineForTests(VirtualMachine):
                 sc_name = self.vm_preference.instance.spec.get("volumes", {}).get("preferredStorageClassName")
                 if sc_name:
                     return sc_name
-            else:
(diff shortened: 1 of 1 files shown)


vsibirsk (approved):

RoniKishner:
/verified
`smoke`

openshift-virtualization-qe-bot-2:

PR #3491, opened 2026-01-18
[IUO] Add test to test assigning default storage class

hmeir:
##### Short description:
Starting from 4.19 - ODF team doesn't set ocs-virt storage class as the default virt-sc while fresh-installing ocs-virt.
After discussing with storage team, we are fine with this behaviour and this will be the default.

##### More details:
In order to make sure it doesn't happen again, we need to add a test to verify this behaviour.

The test should run after test_cnv_installation, and:
1. Get the default storageclass from the matrix.
2. If not virt-default, set it as default storageclass and make sure all the others are not default.
3. Make sure that the all golden images resources are imported successfully.

To make sure the expected sc is default post-installation, to make sure this behaviour is consistent.
If not, it will harm our production testing(install lane + smoke + console test for each relealse.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-64256

## Summary by CodeRabbit

* **Tests**
  * Added tests to verify the cluster's default storage class is set correctly during OpenShift Virtualization installation and that boot sources are reimported after changes.
* **Chores**
  * Utilities updated to allow configurable verification retries and to programmatically mark a storage class as the default.

### DIFF
--- a/tests/install_upgrade_operators/product_install/conftest.py
+++ b/tests/install_upgrade_operators/product_install/conftest.py
@@ -8,7 +8,9 @@ from ocp_resources.hostpath_provisioner import HostPathProvisioner
 from ocp_resources.hyperconverged import HyperConverged
 from ocp_resources.installplan import InstallPlan
 from ocp_resources.persistent_volume import PersistentVolume
+from ocp_resources.storage_class import StorageClass
 from pytest_testconfig import py_config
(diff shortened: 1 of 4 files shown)


hmeir:
deplicate of https://github.com/RedHatQE/openshift-virtualization-tests/pull/3504
closing

PR #3512, opened 2026-01-19
Add tox verify-bugs-are-open-gh to add this check as a check run

rnetser:
##### Short description:
To allow open bugs check to be a GH check run, integrated with myk-org/github-webhook-server#961 - tox was updated with an explicit command
Manual cherrypick of https://github.com/RedHatQE/openshift-virtualization-tests/pull/3048

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -73,7 +73,7 @@ setenv =
 deps =
     python-utility-scripts
 commands =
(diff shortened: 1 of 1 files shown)


rnetser:
/verified
`tox -e verify-bugs-are-open-gh`

PR #3584, opened 2026-01-24
net,bgp: exclude external FRR from BGP router nodes

servolkov:
Sometimes connectivity between UDN VM and external provider network
does not work. The current setup connects external FRR pod to the
cluster nodes network via localnet, and the basic condition is already
complied: external FRR pod and UDN VM are living on different cluster
nodes to avoid connectivity issue. However, it is not enough: the node
where external FRR pod resides should be also excluded from BGP
routing.

**Rationale**: the external FRR pod receives BGP advertisements from all
cluster nodes, including the node it is currently hosted on.
If the local node is used as a BGP nexthop, ECMP may select it,
causing traffic to hairpin via br-ex instead of exiting via eno1.
This breaks return traffic.

Removing the local node from routing guarantees consistent
egress via remote cluster nodes.

## Summary by CodeRabbit

* **Tests**
  * Improved BGP test setup to exclude the node running the external FRR component, yielding more accurate node selection during convergence checks.
  * Removed expected-failure markers and related imports from BGP tests, reflecting increased test reliability and fewer intentional xfail cases.

### DIFF
--- a/tests/network/bgp/conftest.py
+++ b/tests/network/bgp/conftest.py
@@ -88,11 +88,15 @@ def nad_localnet(
 
 @pytest.fixture(scope="module")
 def frr_configmap(
-    workers: list[Node], cnv_tests_utilities_namespace: Namespace, admin_client: DynamicClient
+    workers: list[Node],
+    cnv_tests_utilities_namespace: Namespace,
+    admin_client: DynamicClient,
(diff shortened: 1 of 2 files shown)


openshift-virtualization-qe-bot-5:
Cherry-picked PR net,bgp: exclude external FRR from BGP router nodes into cnv-4.21

PR #3636, opened 2026-01-28
Fix: Handle multi-document YAML in generate_unique_icsp_idms_file

rlobillo:
##### Short description:
When YAML files contain multiple ImageContentSourcePolicy or ImageDigestMirrorSet documents, all were assigned the same name (iib-{version}), causing OpenShift resource conflicts.

This fix adds an index to each document name (iib-{version}-0, iib-{version}-1, etc.) ensuring unique resource names within the same file while maintaining backwards compatibility with the delete_existing_icsp_idms function which uses name prefix matching.

##### More details:
Changes:
- Updated generate_unique_icsp_idms_file() to use yaml.safe_load_all() and iterate through all documents with index counter
- Changed from yaml.dump() to yaml.dump_all() to preserve all documents
- Updated tests to verify unique naming for multiple documents

Fix upgrade when oc catalog mirror generates more than 1 ICSP.

N/A
N/A
##### jira-ticket:
https://issues.redhat.com/browse/CNV-78256

## Release Notes

* **New Features**
  * Added support for multi-document YAML file handling with index-based naming for each document.

* **Tests**
  * Updated test cases to validate multi-document YAML processing and metadata renaming.

### DIFF
--- a/utilities/operator.py
+++ b/utilities/operator.py
@@ -90,12 +90,18 @@ def generate_icsp_idms_file(folder_name, command, is_idms_file, cnv_version=None
 
 
 def generate_unique_icsp_idms_file(file_name, version_string):
-    # update the metadata.name value to generate unique ICSP/IDMS
     with open(file_name) as fd:
-        file_yaml = yaml.safe_load(fd.read())
-    file_yaml["metadata"]["name"] = f"iib-{version_string}"
+        documents = list(yaml.safe_load_all(fd.read()))
+
+    document_index = 0
+    for document in documents:
+        if document and document.get("kind") in ["ImageContentSourcePolicy", "ImageDigestMirrorSet"]:
+            document["metadata"]["name"] = f"iib-{version_string}-{document_index}"
+            document_index += 1
+
     with open(file_name, "w") as current_mirror_file:
-        yaml.dump(file_yaml, current_mirror_file)
+        yaml.dump_all(documents, current_mirror_file, default_flow_style=False)
+
     new_file_name = file_name.replace(file_name, f"{file_name.replace('.yaml', '')}{version_string}.yaml")
     os.rename(file_name, new_file_name)
     return new_file_name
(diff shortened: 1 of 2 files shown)


utilities/operator.py
```diff
@@ -90,12 +90,18 @@ def generate_icsp_idms_file(folder_name, command, is_idms_file, cnv_version=None
...
 def generate_unique_icsp_idms_file(file_name, version_string):
-    # update the metadata.name value to generate unique ICSP/IDMS
     with open(file_name) as fd:
-        file_yaml = yaml.safe_load(fd.read())
-    file_yaml["metadata"]["name"] = f"iib-{version_string}"
+        documents = list(yaml.safe_load_all(fd.read()))
+
+    document_index = 0
+    for document in documents:
+        if document and document.get("kind") in ["ImageContentSourcePolicy", "ImageDigestMirrorSet"]:
```
rnetser: why do you need the `if document` check here?

rlobillo: Files generated by `oc adm catalog mirror` frequently end with `---` and that would lead to an exception if we don't include that protection:


```
>>> yaml_content='''
... kind: test
... metadata:
...     name: test
... ---
... '''
>>> import yaml
>>> documents = list(yaml.safe_load_all(yaml_content))
>>> documents
[{'kind': 'test', 'metadata': {'name': 'test'}}, None]
```

rnetser: ack, you can resolve
please move this test to tier1

utilities/operator.py
```diff
@@ -90,12 +90,18 @@ def generate_icsp_idms_file(folder_name, command, is_idms_file, cnv_version=None
 
 
 def generate_unique_icsp_idms_file(file_name, version_string):
-    # update the metadata.name value to generate unique ICSP/IDMS
     with open(file_name) as fd:
```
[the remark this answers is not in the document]
rlobillo: fully tested as suggested

PR #3706, opened 2026-02-02
tests, net, udn, ip-spec: Add an external connectivity test

EdDev:

Test that a VM with an explicit IP address specified is reaching an external IP address.

> **Note**: The public external IP needs to be changed for all network tests (potentially beyond).
> It is not safe to use an IP which is not controlled by the CI setup.

##### jira-ticket:

https://issues.redhat.com/browse/CNV-76572

## Summary by CodeRabbit

* **Tests**
  * Enabled external connectivity validation for virtual machines with explicit IP address specifications to ensure proper network access to external services.


servolkov:
> Please correct the PR header connectivit -> connectivity

@EdDev commit message as well

PR #3722, opened 2026-02-03
Set explicit API Version for HyperConverged

nunnatsa:
##### Short description:
As a preparation for the planned introduction of the `v1` API of the `HyperConverged` resource, explicitly set its API version to `v1beta1`.

##### More details:
The planned `v1` API of the HyperConverged resource, is going to restructure fields, moving them around, deprecating several fields and so on.

The current implementation takes the latest version of the API, so once the new version is introduced, the client will try to read or write the `HyperConverged` resource as `v1`, while the code is using its `v1beta1` API. This will break any test uses this type.

To avoid that, and give the developers some time to adjust, this PR explicitly sets the API version of the `HyperConverged` type to `v1beta1`, when fetching it from Kubernetes.

That way, the client will explicitly read and write the `HyperConverged` resource as `v1beta1`, and with the planned conversion webhook, the addition of the new `v1` API version will be transparent.

See above. Without this PR, the tests will fail once HCO v1 API will be introduced.

n/a

None

##### jira-ticket:
https://issues.redhat.com/browse/CNV-78901

## Summary by CodeRabbit

* **Bug Fixes**
  * HyperConverged resource lookups now use the v1beta1 API version, improving reliability of resource discovery and reducing inconsistent "not found" results across environments.

### DIFF
--- a/utilities/infra.py
+++ b/utilities/infra.py
@@ -653,6 +653,7 @@ def get_hyperconverged_resource(client, hco_ns_name):
         namespace=hco_ns_name,
         name=hco_name,
     )
+    hco.api_version = hco.ApiVersion.V1BETA1
     if hco.exists:
         return hco
     raise ResourceNotFoundError(f"Hyperconverged: {hco_name} not found in {hco_ns_name}")


dshchedr (dismissed):

vsibirsk (dismissed):

utilities/infra.py
```diff
@@ -653,6 +653,7 @@ def get_hyperconverged_resource(client, hco_ns_name):
         namespace=hco_ns_name,
         name=hco_name,
     )
+    hco.api_version = NamespacedResource.ApiVersion.V1BETA1
```
rnetser:
```suggestion
    hco.api_version = hco.ApiVersion.V1BETA1
```
can use the object, reduces the need of another import.

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  **Analysis Requirements:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

</details>

PR #3798, opened 2026-02-10
storage, ipv6: Adjust gating tests

azhivovk:
The following storage gating tests use CirrOS VMs:
- test_clone_from_fs_to_block_using_dv_template
- test_create_vm_with_cloned_data_volume_permissions_for_pods_positive

By default, VirtualMachineForTests assumes ssh=true. However, on IPv6 single-stack clusters, SSH remains disabled because guest VMs are not automatically
assigned a masquerade IPv6 address.
Since CirrOS lacks support for the Guest Agent and Cloud-Init, the standard networking fixes cannot be applied.

On this change, we set ssh=false for such VMs,
as SSH connectivity is not a requirement for these test cases

## Summary by CodeRabbit

* **Tests**
  * Updated test infrastructure to streamline VM initialization and testing procedures by modifying SSH connectivity verification behavior during test execution.

### DIFF
--- a/tests/storage/cdi_clone/test_clone.py
+++ b/tests/storage/cdi_clone/test_clone.py
@@ -50,6 +50,7 @@ def create_vm_from_clone_dv_template(
         os_flavor=Images.Cirros.OS_FLAVOR,
         client=client,
         memory_guest=Images.Cirros.DEFAULT_MEMORY_SIZE,
+        ssh=False,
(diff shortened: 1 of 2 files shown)


jpeimer (changes requested):

azhivovk:
The changed tests will be using fedora VMs, closing this PR

PR #3844, opened 2026-02-12
[4.20] net,traffic-generator: add `container` parameter to pod.execute

servolkov:
In `PodTcpClient` `pod.execute` must contain container parameter explicitly to avoid collisions.

Original change: https://github.com/RedHatQE/openshift-virtualization-tests/pull/2887

### DIFF
--- a/libs/net/traffic_generator.py
+++ b/libs/net/traffic_generator.py
@@ -143,7 +143,7 @@ def _is_process_running(vm: BaseVirtualMachine, cmd: str) -> bool:
 class PodTcpClient(BaseTcpClient):
     """Represents a TCP client that connects to a server to test network performance.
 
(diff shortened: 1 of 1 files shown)


servolkov:
> how was the verification successful? 

@rnetser about how it was successful - if we need I can investigate, but honestly I don't want to spend a lot time here, I will try to catch it when I am more available.  

> anything we can learn from this to the future?

from my perspective it is worth not to cherry-pick several commits at once to avoid collisions and make the process clearer in terms of review and verification. On the other hand, I have experienced myself and have seen from colleagues when, in order to successfully cherry-pick a certain commit, you need to add a part of another commit (just a part). In this case, we need to clearly describe it in the description and conduct a full review like it is not a just backport.

PR #3948, opened 2026-02-23
fix: use admin client for node object to avoid unprivileged 403 errors

geetikakay:
VMs created with unprivileged_client pass that client down to the Node object via vmi.node. Node is cluster-scoped so unprivileged users can't query node status causing 403 Forbidden in wait_for_node_status.


RoniKishner (approved):

geetikakay:
/verified 
I have verified these changes locally

PR #4003, opened 2026-02-27
Guard cnv_current_version against missing csv spec.version

geetikakay:
##### Short description:
spec.version is not a required field in the ClusterServiceVersion schema. If absent, the fixture silently returns None causing unexpected behavior in multiple downstream tests.
identified during verifying https://github.com/RedHatQE/openshift-virtualization-tests/pull/3984 that we always assume csv has .spec.version , although it exist but since it's optional if it's not there , We should fail early and at right place

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation to ensure CSV specification version information is properly defined, raising an error if the required version field is missing rather than silently accepting empty values.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -956,7 +956,10 @@ def cnv_current_version(installing_cnv, csv_scope_session):
     if installing_cnv:
         return CNV_NOT_INSTALLED
     if csv_scope_session:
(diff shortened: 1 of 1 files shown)


dshchedr (approved):

openshift-virtualization-qe-bot-2:

PR #4029, opened 2026-03-02
Fix tests marking and collection order

rnetser:
##### Short description:
re-order `add_tier2_marker` and `mark_tests_by_team` to allow correct tier 2 tests collection using EXCLUDE_MARKER_FROM_TIER2_MARKER with updated teams markers

## Summary by CodeRabbit

* **Chores**
  * Adjusted internal test configuration to change the order in which test markers are applied, improving test categorization and reliability during collection.

**Note:** This is an internal test infrastructure change; there are no user-facing changes in this release.


rnetser:
/verified 



remove chaos marker from  a test and make sure it is not collected as tier2

PR #4056, opened 2026-03-03
net, stuntime, std: VM stuntime measurement during live migration

Anatw:
##### Short description:
Introduce the STD for VM stuntime measurement during live migration on Linux bridge and OVN localnet secondary networks, for both IPv4 and IPv6. These tests provide baseline measurements and a framework
for regression detection.

Baseline and Threshold:
- Global threshold - setting the threshold to the minimum of (max_observed × 4) and 5 seconds — i.e. 4× the worst observed value, capped at 5 seconds.
- The threshold will be hardcoded based on 10-run baselines, on a BM cluster.

Measurement Methodology:
- Ping command: ICMP ping at 100ms intervals with UNIX timestamps (`ping -D -O -i 0.1` for IPv4, `ping -6 -D -O -i 0.1` for IPv6).
- Stuntime calculation: Stuntime is the largest gap between any two consecutive successful replies in the ping log - i.e., the maximum time difference between timestamps of successive successful packets. The boundaries (last success before loss, first success after recovery) are the pair of packets that define that gap.
Stuntime = (Timestamp of first success after) - (Timestamp of last success before)
- Alternatives rejected: tcping (introducing a new dependency), iperf3 (unnecessary complexity), and curl (unnecessary complexity - requires an active web server inside the VM).

IP Family:
- IPv4 and IPv6 are measured in separate migrations to avoid interactions between ARP (IPv4) and NDP (IPv6) recovery paths.
- ip_family is a parametrize dimension with pytest.mark.ipv4 and pytest.mark.ipv6 applied per value, allowing selective runs: IPv4-only (`pytest -m ipv4`), IPv6-only (`pytest -m ipv6`), or both (no `-m` flag needed).
- Both ipv4 and ipv6 markers restrict execution to BM clusters (dual-stack lane).
- Total scenarios: 24 (2 CNI types × 3 migration paths × 2 ping initiators × 2 IP families).
- IPv4 and IPv6 are covered in separate subtests under each scenario.

The tests are ordered so that each test execution ends where the next test execution starts. Also, there is no need to continue execution if a previous test has failed.

##### jira-ticket:
https://issues.redhat.com/browse/CNV-78677

## Summary by CodeRabbit

* **Tests**
  * Added two test classes covering Linux Bridge and OVN LocalNet stuntime measurements.
  * Each class includes a single, fully documented test method (parameters, preconditions, steps, expected outcomes).
  * Both test methods are present in the codebase but intentionally disabled from automated test collection.

### DIFF
--- a/tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
+++ b/tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
@@ -1,0 +1,170 @@
+"""
+Linux bridge migration stuntime measurement tests over live migration.
+
+Tests measure the connectivity gap (stuntime) during VM live migration on Linux bridge
+secondary network, for both IPv4 and IPv6, for regression detection.
+Stuntime is defined as the connectivity gap from last successful reply before loss
+to first successful reply after recovery.
+
+Stuntime is measured using ICMP ping from client to server in 0.1s intervals, using ping -D so each
+log line includes a timestamp for gap calculation.
+The under-test VMs are configured on a Linux bridge secondary network, with a single interface,
+on which IPv4/IPv6 static addresses will be defined according to the environment the test runs on.
+
+Client - The connectivity initiator VM that runs continuous ping toward the server VM.
+Server - The connectivity listener VM that receives the ping and responds.
+
+STP Reference:
+https://github.com/RedHatQE/openshift-virtualization-tests-design-docs/blob/main/stps/sig-network/stuntime_measurement.md
+"""
+
+import pytest
+
+__test__ = False
+
+"""
+Parametrize:
+    - ip_family:
+        - ipv4 [Markers: ipv4]
+        - ipv6 [Markers: ipv6]
+
+Preconditions:
+    - Shared under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+    - Shared under-test client VM on Linux bridge secondary network, for that same IP family,
+      initially running on the same node as the server VM.
+"""
+
+
+@pytest.mark.incremental
+class TestMigrationStuntime:
+    @pytest.mark.polarion("CNV-15252")
+    def test_client_migrates_off_server_node(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the client
+        VM migrates from the node hosting the server VM into a different node.
+
+        Preconditions:
+            - Under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on Linux bridge secondary network, for that same IP family,
+              running on the same node as the server VM.
+            - Ping initiated from the client to the server.
+
+        Steps:
+            1. Initiate live migration of the client VM to a node different from the node hosting the server VM
+               and wait for migration completion.
+            2. Stop the continuous ping.
+            3. Compute stuntime from the ping results.
+
+        Expected:
+            - Measured stuntime does not exceed the global threshold.
+        """
+
+    @pytest.mark.polarion("CNV-15253")
+    def test_client_migrates_between_non_server_nodes(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the client VM migrates between nodes
+        while the client and server VMs remain on different nodes.
+
+        Preconditions:
+            - Under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on Linux bridge secondary network, for that same IP family,
+              running on a worker node other than the node hosting the server VM.
+            - Ping initiated from the client to the server.
+
+        Steps:
+            1. Initiate live migration of the client VM to a node different from the node hosting the server VM
+               and wait for migration completion.
+            2. Stop the continuous ping.
+            3. Compute stuntime from the ping results.
+
+        Expected:
+            - Measured stuntime does not exceed the global threshold.
+        """
+
+    @pytest.mark.polarion("CNV-15254")
+    def test_client_migrates_to_server_node(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the client VM migrates
+        from a node other than the node hosting the server VM onto the node hosting the server VM.
+
+        Preconditions:
+            - Under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on Linux bridge secondary network, for that same IP family,
+              running on a worker node other than the node hosting the server VM.
+            - Ping initiated from the client to the server.
+
+        Steps:
+            1. Initiate live migration of the client VM to the node hosting the server VM
+               and wait for migration completion.
+            2. Stop the continuous ping.
+            3. Compute stuntime from the ping results.
+
+        Expected:
+            - Measured stuntime does not exceed the global threshold.
+        """
+
+    @pytest.mark.polarion("CNV-15255")
+    def test_server_migrates_off_client_node(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the server
+        VM migrates from the node hosting the client VM into a different node.
+
+        Preconditions:
+            - Under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on Linux bridge secondary network, for that same IP family,
+              running on the same node as the server VM.
+            - Ping initiated from the client to the server.
+
+        Steps:
+            1. Initiate live migration of the server VM to a node different from the node hosting the client VM
+               and wait for migration completion.
+            2. Stop the continuous ping.
+            3. Compute stuntime from the ping results.
+
+        Expected:
+            - Measured stuntime does not exceed the global threshold.
+        """
+
+    @pytest.mark.polarion("CNV-15256")
+    def test_server_migrates_between_non_client_nodes(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the server VM migrates between nodes
+        while the client and server VMs remain on different nodes.
+
+        Preconditions:
+            - Under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on Linux bridge secondary network, for that same IP family,
+              running on a worker node other than the node hosting the server VM (before and after migration).
+            - Ping initiated from the client to the server.
+
+        Steps:
+            1. Initiate live migration of the server VM to a node different from the node hosting the client VM
+               and wait for migration completion.
+            2. Stop the continuous ping.
+            3. Compute stuntime from the ping results.
+
+        Expected:
+            - Measured stuntime does not exceed the global threshold.
+        """
+
+    @pytest.mark.polarion("CNV-15257")
+    def test_server_migrates_to_client_node(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the server VM migrates from a node
+        other than the node hosting the client VM onto the node hosting the client VM.
+
+        Preconditions:
+            - Under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on Linux bridge secondary network, for that same IP family,
+              running on a worker node other than the node hosting the server VM.
+            - Ping initiated from the client to the server.
+
+        Steps:
+            1. Initiate live migration of the server VM to the node hosting the client VM
+               and wait for migration completion.
+            2. Stop the continuous ping.
+            3. Compute stuntime from the ping results.
+
+        Expected:
+            - Measured stuntime does not exceed the global threshold.
+        """
--- a/tests/network/localnet/migration_stuntime/test_migration_stuntime.py
+++ b/tests/network/localnet/migration_stuntime/test_migration_stuntime.py
@@ -1,0 +1,170 @@
+"""
+OVN localnet (OVS bridge) migration stuntime measurement tests over live migration.
+
+Tests measure the connectivity gap (stuntime) during VM live migration on OVN localnet
+secondary network, for both IPv4 and IPv6, for regression detection.
+Stuntime is defined as the connectivity gap from last successful reply before loss
+to first successful reply after recovery.
+
+Stuntime is measured using ICMP ping from client to server in 0.1s intervals, using ping -D so each
+log line includes a timestamp for gap calculation.
+The under-test VMs are configured on an OVN localnet secondary network, with a single interface,
+on which IPv4/IPv6 static addresses will be defined according to the environment the test runs on.
+
+Client - The connectivity initiator VM that runs continuous ping toward the server VM.
+Server - The connectivity listener VM that receives the ping and responds.
+
+STP Reference:
+https://github.com/RedHatQE/openshift-virtualization-tests-design-docs/blob/main/stps/sig-network/stuntime_measurement.md
+"""
+
+import pytest
+
+__test__ = False
+
+"""
+Parametrize:
+    - ip_family:
+        - ipv4 [Markers: ipv4]
+        - ipv6 [Markers: ipv6]
+
+Preconditions:
+    - Shared under-test server VM on OVN localnet secondary network, for the IP family from ip_family parametrization.
+    - Shared under-test client VM on OVN localnet secondary network, for that same IP family,
+      initially running on the same node as the server VM.
+"""
+
+
+@pytest.mark.incremental
+class TestMigrationStuntime:
+    @pytest.mark.polarion("CNV-15258")
+    def test_client_migrates_off_server_node(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the client
+        VM migrates from the node hosting the server VM into a different node.
+
+        Preconditions:
+            - Under-test server VM on OVN localnet secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on OVN localnet secondary network, for that same IP family,
+              running on the same node as the server VM.
+            - Ping initiated from the client to the server.
+
+        Steps:
+            1. Initiate live migration of the client VM to a node different from the node hosting the server VM
+               and wait for migration completion.
+            2. Stop the continuous ping.
+            3. Compute stuntime from the ping results.
+
+        Expected:
+            - Measured stuntime does not exceed the global threshold.
+        """
+
+    @pytest.mark.polarion("CNV-15259")
+    def test_client_migrates_between_non_server_nodes(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the client VM migrates between nodes
+        while the client and server VMs remain on different nodes.
+
+        Preconditions:
+            - Under-test server VM on OVN localnet secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on OVN localnet secondary network, for that same IP family,
+              running on a worker node other than the node hosting the server VM.
+            - Ping initiated from the client to the server.
+
+        Steps:
+            1. Initiate live migration of the client VM to a node different from the node hosting the server VM
+               and wait for migration completion.
+            2. Stop the continuous ping.
+            3. Compute stuntime from the ping results.
+
+        Expected:
+            - Measured stuntime does not exceed the global threshold.
+        """
+
+    @pytest.mark.polarion("CNV-15260")
+    def test_client_migrates_to_server_node(self):
+        """
+        Test that measured stuntime does not exceed the global threshold when the client VM migrates
+        from a node other than the node hosting the server VM onto the node hosting the server VM.
+
+        Preconditions:
+            - Under-test server VM on OVN localnet secondary network, for the IP family from ip_family parametrization.
+            - Under-test client VM on OVN localnet secondary network, for that same IP family,
+              running on a worker node other than the node hosting the server VM.
+            - Ping initiated from the client to the server.
+
+        Steps:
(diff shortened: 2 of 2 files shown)


tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,75 @@
...
+    @pytest.mark.polarion("CNV-00001")
+    def test_stuntime(self):
+        """
+        Test that measured stuntime during live migration does not exceed the per-scenario threshold.
+
+        Markers:
+            - ipv4
+
+        Parametrize:
+            - migration_path: [static_to_different, different_to_static, between_different]
```
yossisegev: If I wasn't already familiar with the STP, I wouldn't understand what is said here.
Please add a short explanation so the path will be clear. Thanks

yossisegev (changes requested):

tests/network/stuntime/test_stuntime_measurement.py
yossisegev: Much clearer now, thank you.
Please resolve.

yossisegev (approved):
Thank you

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,79 @@
...
+
+class TestStuntimeLinuxBridge:
+    """Stuntime measurement on Linux bridge secondary network."""
+
+    @pytest.mark.polarion("CNV-00001")
+    def test_stuntime(self):
+        """
+        Test that measured stuntime during live migration does not exceed the per-scenario threshold.
+
+        Markers:
```
servolkov: nit
I would add markers during implementation (sounds like implementations detail to me), don't know why we need it in STP.  Since we don't have many examples, it is up to you. If you decide to leave it, consider moving it under class level not repeating for each test.

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,79 @@
+"""
```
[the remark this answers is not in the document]
servolkov: I would consider again a directory like `non-functional`, I just want to explicitly separate functional testing (our regular ones) from tests like these.

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,79 @@
...
+            - Running Fedora static-VM on Linux bridge secondary network, running on worker1.
+            - Ping running at 100 ms intervals (ping -D -O -i 0.1) from initiator VM to peer.
+
+        Steps:
+            1. Initiate live migration of the migrated VM along the specified path.
+            2. Parse ping output for connectivity gap (last success before loss to first success after recovery).
+            3. Compare measured stuntime against per-scenario threshold.
+
+        Expected:
+            - Measured stuntime does not exceed the scenario threshold (min(max×4, 5s)).
```
servolkov: not sure we need this detail in STD: `min(max×4, 5s)`, but I may be wrong.

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,79 @@
...
+            - ipv4
+
+        Parametrize:
+            - migration_path: [static_to_different, different_to_static, between_different]
+              (relative to static VM's node: from static's node to different; from different to
+              static's node; between two nodes both different from static's node). See STP II.1.
+            - ping_initiator: [migrated_vm, static_vm]
+
+        Preconditions:
+            - Running Fedora migrated-VM on Linux bridge secondary network, running on worker1.
```
servolkov: 1) why we need to point that this is Fedora? I think "Running VM..." is enough. Per my understanding tests are not VM-specific. 
2) `migrated-VM` does not mean the VM migrated in precondition, right? I read a few times to figure that in reality the VM is not migrated, it is just a VM that will be migrated in tests. If so, please, consider writing it more accurately, like: `Running VM for migration...` or `Running VM for migration ("migrated_vm")...` or some sort of. Probably, it is worth to name it `migration_vm`, not `migrated`

servolkov (approved):

azhivovk (approved):

Anatw:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/e58defdf311f3366359d6cab53d8374ffd0e6a36..24a6b1253f9b802d205a1fee29f2d36723d4a2fb):
Update the STD to align with changes in the STP: add IPv6 to stuntime test scope. The STP, STD, and test file are updated accordingly (scenario count: 24). The commit message was updated to reflect the IPv6 addition.

yossisegev (changes requested):
Thank you @Anatw

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,94 @@
...
+               - between_different (migrate between two nodes, both different from static VM's node).
+            - ping_initiator: VM from which the ping command is launched toward the peer.
+              Values:
+               - migrated_vm (ping from the VM for migration).
+               - static_vm (ping from the VM that stays on its node).
+
+        Preconditions:
+            - Running VM for migration on OVN localnet secondary network, running on worker1.
+            - Running static VM on OVN localnet secondary network, running on worker1.
+            - Ping running at 100 ms intervals from ping_initiator VM to peer (ping or ping6 depending on ip_family).
```
azhivovk: https://github.com/RedHatQE/openshift-virtualization-tests/pull/4056/changes#r2923601421

azhivovk (changes requested):

yossisegev (approved):

azhivovk (approved):

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,94 @@
...
+            - ping_initiator: VM from which the ping command is launched toward the peer.
+              Values:
+               - migrated_vm (ping from the VM for migration).
+               - static_vm (ping from the VM that stays on its node).
+
+        Preconditions:
+            - Running VM for migration on Linux bridge secondary network, running on worker1.
+            - Running static VM on Linux bridge secondary network, running on worker1.
+            - Ping running at 100 ms intervals from ping_initiator VM to peer.
+            - Predefined stuntime threshold to test against (per-scenario, derived from BM baseline runs).
```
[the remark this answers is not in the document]
EdDev: The threshold exists, but at this stage we have not measured it.
I do not think the value must be defined now, it can be left as a value to be filled later.

azhivovk (approved):

EdDev (changes requested):
Thanks.

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,102 @@
...
+               - remote_to_co_located (migrate from a remote node to peer's node).
+               - remote_to_remote (migrate between two remote nodes).
+            - ping_initiator: VM from which the ping command is launched toward the peer.
+              Values:
+               - migrated_vm (ping from the VM for migration toward the peer).
+               - peer_vm (ping from the peer toward the VM for migration).
+
+        Preconditions:
+            - Running VM for migration on Linux bridge secondary network, running on worker1.
+            - Running peer VM on Linux bridge secondary network, running on worker1.
```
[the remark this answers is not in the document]
Anatw: 1. We don't have many previous examples, but in the existing ones I see a "under-test VM." In an earlier comment you suggested the client-server terminology. I'm not sure how these approaches combine. I'm rephrasing as client-server and waiting to hear if more refactoring is needed.
2. Ack, rephrased.
3. Ack.

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,47 @@
...
+    Markers:
+        - ipv4
+        - ipv6
+        (applied per ip_family parametrize value for selective runs)
+
+    Parametrize:
+        - cni_type: [linux_bridge, ovn_localnet]
+        - ip_family: [ipv4, ipv6]
+        - migration_path: [co_located_to_remote, remote_to_co_located, remote_to_remote]
+        - connectivity_initiator: [client, server]
```
EdDev: Does not seem right to me. In some other thread I explained that the initiator is always the client.

Anatw:
Change: align client/server with migrating_vm_role.

tests/network/l2_bridge/test_migration_stuntime.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Test that measured stuntime during live migration on Linux bridge does not exceed the single global threshold.
+    Stuntime is the connectivity gap from last successful reply before loss to first successful reply after recovery.
+
+    The same VMs under test are shared across all parametrized runs.
+
+    Parametrize:
+        - ip_family:
+            - ipv4 [Markers: ipv4]
+            - ipv6 [Markers: ipv6]
```
azhivovk: Why do we need `ip_family` param? 
Usually we run the tests according to existing cluster network stack, shouldn't we also apply this approach here?

tests/network/stuntime/test_stuntime_measurement.py
EdDev: I think you tried to resolve it by adding a "migration" role. I guess it can be discussed on the new version and resolved here.

EdDev (changes requested):
The usage of a new module and embedding into an existing one is questionable, I suggest you check if new fixtures are needed or not and decide accordance to that. The minimum is a module as this is a different test category as @servolkov mentioned. (there is an open thread about this)

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
```diff
@@ -0,0 +1,151 @@
...
+https://github.com/RedHatQE/openshift-virtualization-tests-design-docs/blob/main/stps/sig-network/stuntime_measurement.md
+"""
+
+import pytest
+
+__test__ = False
+
+
+@pytest.mark.polarion("CNV-00001")
+def test_client_migrates_same_to_different():
```
azhivovk: The name might be confusing since it might not be clear what is same and different
Maybe rename it to something like: `test_client_migrates_off_server_node`

EdDev (commented):
Thank you, this looks like a good direction.

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
Anatw: Sound good, thanks you both.

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
```diff
@@ -0,0 +1,158 @@
+"""
+Linux bridge migration stuntime measurement tests over live migration.
+
+Tests measure the connectivity gap (stuntime) during VM live migration on Linux bridge
+secondary network, for both IPv4 and IPv6, for regression detection.
+Stuntime is defined as the connectivity gap from last successful reply before loss
+to first successful reply after recovery.
+
```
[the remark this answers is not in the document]
Anatw: Ack, added definitions.

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,47 @@
...
+
+    Preconditions:
+        - Running under-test VMs on secondary network (CNI per cni_type). Initial placement per migration_path.
```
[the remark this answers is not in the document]
Anatw: Resolving obsolete comments.

tests/network/stuntime/test_stuntime_measurement.py
```diff
@@ -0,0 +1,94 @@
...
+    @pytest.mark.polarion("CNV-00001")
+    def test_stuntime(self):
+        """
+        Test that measured stuntime during live migration does not exceed the per-scenario threshold.
+
+        Parametrize:
+            - ip_family: IP family used for connectivity downtime measurements.
+              Values:
+               - ipv4 (pytest.mark.ipv4, ping -D -O -i 0.1).
+               - ipv6 (pytest.mark.ipv6, ping -6 -D -O -i 0.1).
```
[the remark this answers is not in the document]
Anatw: Resolving obsolete comments.

yossisegev (approved):
Thank you, especially for tearing down the matrix in favor of explicit, separate test cases. This brings much clarity to this plan.

azhivovk (approved):
Thanks

EdDev (changes requested):

rnetser (commented):
stopped my review after test_client_migrates_between_non_server_nodes

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
```diff
@@ -0,0 +1,169 @@
...
+        - Under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+        - Under-test client VM on Linux bridge secondary network, for that same IP family,
+          running on a worker node other than the node hosting the server VM.
+        - Ping initiated from the client to the server.
+
+    Steps:
+        1. Initiate live migration of the client VM to a node different from the node hosting the server VM
+           and wait for migration completion.
+        2. Stop the continuous ping.
+        3. Compute stuntime from the ping results.
```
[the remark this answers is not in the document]
yossisegev: We actually had a discussion about it within the team.
Initially @Anatw did apply parametization, but it ended with a multi-dimensional matrix which was almost impossible to understand.
So we came to the conclusion that because this is actually a test-design (even though it is the base for the implementation), it can and should depict the test cases explicitly and clearly; then, when actually implementing, it can be improved by parametization and other means that pytest enables.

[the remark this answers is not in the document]
Anatw: Thanks @yossisegev for the clear summary. We intentionally keep explicit separate STD scenarios here for readability/reviewability, and plan to introduce parametrization with ids at implementation stage where it improves usability without reintroducing a hard-to-read matrix. Hope this works for you, @rnetser.

EdDev (changes requested):
The inter test adjusting is wasteful, please fix it by enforcing order.

EdDev (approved):
Thank you.

There is one small comment, other than that it looks good to me.

azhivovk (approved):

EdDev (approved):
Thank you.

/approve

rnetser (dismissed):

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
```diff
@@ -0,0 +1,158 @@
...
+
+@pytest.mark.polarion("CNV-00001")
+def test_client_migrates_same_to_different():
+    """
+    Test that measured stuntime does not exceed the global threshold when the client
+    VM migrates from the node hosting the server VM into a different node.
+
+    Preconditions:
+        - Under-test server VM on Linux bridge secondary network.
+        - Under-test client VM on Linux bridge secondary network, running on the same node as the server VM.
```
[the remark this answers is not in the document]
servolkov: > Also, when failing a single test in the start/middle (assuming you do set it to run in sequence), it should fail all the other tests as they are no longer going to check the expected 

I don't see this logic is applicable for the current test suite. Threshold measurement is independent for each scenario, as I see. Why `test_client_migrates_off_server_node` should affect `test_client_migrates_between_non_server_nodes`, for example? The second test may pass even if the threshold check failed for the first one.

Moreover, I think it's even harmful: if one of the measurements fails, I prefer to know whether other tests pass or fail, we should not loose coverage.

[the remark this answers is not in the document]
servolkov: I understand, but I am highlighting the risk. In fact, tests in this test suite are independent (logically). And I suspect the suite will be launched rarely. In case of early failure, we loose coverage for a long time for other tests that became dependent due to implementation. Roughly, we can sit for a month or more without proper results. If we are considering it's okay, no more to discuss then.

Anatw:
/verified

```
uv run pytest -s --skip-deprecated-api-test --pdbcls=IPython.terminal.debugger:TerminalPdb --tc-format=python --cluster-sanity-skip-storage-check tests/network/localnet/migration_stuntime/test_migration_stuntime.py tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py --collect-only

==== no tests collected (1 deselected) in 0.07s ====
```

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  **Analysis Requirements:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

</details>

Anatw:
Change: add Polarion ID.

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
```diff
@@ -0,0 +1,170 @@
...
+        - ipv6 [Markers: ipv6]
+
+Preconditions:
+    - Shared under-test server VM on Linux bridge secondary network, for the IP family from ip_family parametrization.
+    - Shared under-test client VM on Linux bridge secondary network, for that same IP family,
+      initially running on the same node as the server VM.
+"""
+
+
+@pytest.mark.incremental
```
rnetser, line 38: with force pushing and losing history - was this in the approved version?
any reason to use incremental here? there is not logical dependnecy between tests - i.e `test_client_migrates_off_server_node` may fail and   `test_client_migrates_between_non_server_nodes` should still be executed.

@EdDev

EdDev: Yes, it was raised.
Yes, it was controversial, @servolkov did not like it, I asked for it, thread existed, thread is missing (was it resolved by mistake?).


Frankly, I’m getting annoyed by this.
If you think it is worth to double the time of this test group in order to keep test independence, go with that. I find it absurd, but I lost interest to invest in convincing.

EdDev: Here is the threat which was lost: https://github.com/RedHatQE/openshift-virtualization-tests/pull/4056#discussion_r2994755886

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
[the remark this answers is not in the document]
EdDev: From my perspective, if one fails, the contract is broken and it needs to be re-evaluated. We decided for simplicity and maintainability to expose one value for the stuntime.

I do not know what rarely translates to, and the only risk I see is ignoring execution failures.
I can claim that increasing test complexity and time is also raising the risk of flakes (double the migrations, double the risk).

Said all that, I’m taking a step back and letting the experts decide.
I raised all my reasoning, feel free to agree on this with the root approver. @rnetser also raised this topic, so you may be in agreement.

EdDev:
From my perspective this change is good.
I am waiting for another reviewer to check the polarion IDs correctness before approving.

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
Anatw: Yes, this was in the approved version. The only change I did in the last push was the Polarion Ids.
The incremental addition was discussed in this [thread](https://github.com/RedHatQE/openshift-virtualization-tests/pull/4056#discussion_r2994755886), added in [this change](https://github.com/RedHatQE/openshift-virtualization-tests/pull/4056#issuecomment-4154062719).

Anatw:
/verified

```
uv run pytest -s --skip-deprecated-api-test --pdbcls=IPython.terminal.debugger:TerminalPdb --tc-format=python --cluster-sanity-skip-storage-check tests/network/localnet/migration_stuntime/test_migration_stuntime.py tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py --collect-only

==== no tests collected (1 deselected) in 0.07s ====
```

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  **Analysis Requirements:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

</details>

tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py
rnetser: > Yes, it was raised. Yes, it was controversial, @servolkov did not like it, I asked for it, thread existed, thread is missing (was it resolved by mistake?).
> 
> Frankly, I’m getting annoyed by this. If you think it is worth to double the time of this test group in order to keep test independence, go with that. I find it absurd, but I lost interest to invest in convincing.


these are 2 different things; tests can be grouped for re-use but not have this dependency.
if this was apporved, will not block but please revisit when writing the tests.

servolkov (approved):

EdDev (approved):
Thanks.

Raising @servolkov vote
/approve

Anatw:
/verified

```
uv run pytest -s --skip-deprecated-api-test --pdbcls=IPython.terminal.debugger:TerminalPdb --tc-format=python --cluster-sanity-skip-storage-check tests/network/localnet/migration_stuntime/test_migration_stuntime.py tests/network/l2_bridge/migration_stuntime/test_migration_stuntime.py --collect-only

==== no tests collected (1 deselected) in 0.07s ====
```

openshift-virtualization-qe-bot-4:

PR #4145, opened 2026-03-12
[4.21] Storage: replace artifactory with DataSource in test_golden_image

kshvaika:
##### Short description:
Manual cherry-pick: https://github.com/RedHatQE/openshift-virtualization-tests/pull/4095

### DIFF
--- a/tests/storage/utils.py
+++ b/tests/storage/utils.py
@@ -144,10 +144,7 @@ def upload_image(token, data, asynchronous=False, client=None):
     try:
         with open(data, "rb") as fd:
             fd_data = fd.read()
-    except OSError as error:
-        LOGGER.error(
-            f"Failed to read upload image (type={type(data).__name__}); treating input as raw data. error={error}"
-        )
+    except OSError | IOError:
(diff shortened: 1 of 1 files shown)


tests/storage/utils.py
```diff
@@ -476,3 +477,12 @@ def create_windows_directory(windows_vm: VirtualMachineForTests, directory_path:
         windows_vm=windows_vm,
         directory_path=directory_path,
     )
+
+
+def get_dv_size_from_datasource(data_source: DataSource) -> str | int | None:
+    source_dict = data_source.source.instance.to_dict()
+    source_spec_dict = source_dict["spec"]
+    dv_size = source_spec_dict.get("resources", {}).get("requests", {}).get("storage") or source_dict.get(
```
OhadRevah: I have a question about this one, for the tests that using this util, is it ok that they will get None here and use the default of the create_dv function?
If so, so there is no need to use this function and just use the default of create_dv, if not then assert that the dv_size is not None and return a value.
WDYT?

PR #4171, opened 2026-03-16
storage, infra, ipv6: Adjust import http tests

azhivovk:
Storage gating tests in `tests/storage/cdi_import/test_import_http.py` require the following changes for running on IPv6 single-stack clusters:
- Short hostname internal-http.cnv-tests-utilities went through the external HTTP proxy (not in no_proxy); FQDN .svc.cluster.local matches and bypasses it.
- openssl s_client - IPv6 addresses need bracket syntax [::1]:443
- Use Alpine v3.20.1 from artifactory for all architectures

These changes are followed by update of the image used in test to v1.2.0:
- TLS cert SAN - old cert baked into the image only had the short name; needed to regenerate with FQDN included
- Makefile update to use Alpine v3.20.1 from artifactory for all architectures
- Add IPv6 listening sockets to Nginx server blocks

test_unpack_compressed is dropped as it's covered in Tier1

## Summary by CodeRabbit

* **Chores**
  * Updated internal HTTP test image tag to v1.2.0
  * Added Artifactory-backed Alpine image references for multiple architectures

* **Bug Fixes**
  * Certificate retrieval now formats IPv6 addresses with brackets for host:port
  * Internal service address recognition updated to use fully-qualified in-cluster DNS

* **Tests**
  * Storage/import tests switched to Alpine images and removed a failing compressed-unpack case

### DIFF
--- a/libs/infra/images.py
+++ b/libs/infra/images.py
@@ -23,6 +23,8 @@ class Cirros:
 @dataclass
 class Alpine:
     QCOW2_IMG: str | None = None
+    QCOW2_IMG_VERSIONED: str | None = None
+    RAW_IMG_XZ: str | None = None
     DIR: str = f"{BASE_IMAGES_DIR}/alpine-images"
     DEFAULT_DV_SIZE: str = "1Gi"
     DEFAULT_MEMORY_SIZE: str = "128M"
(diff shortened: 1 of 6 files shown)


azhivovk:
/wip
mismath of cirros version between arm and amd in code - need to resolve it and change the makefile in image accordingly

libs/infra/images.py
```diff
@@ -23,6 +23,8 @@ class Cirros:
 @dataclass
 class Alpine:
     QCOW2_IMG: str | None = None
+    QCOW2_IMG_A: str | None = None
```
jpeimer: Please let's give it a more explanatory name...
For other images, we used the same constants names for Artifactory and the internal http server, but it seems like it would be easier to distinguish them. So we could more easily understand where each image comes from.

PR #4564, opened 2026-04-23
Switching to Claude Code convention

geetikakay:
@ is a Cursor convention and does not work in Claude Code. The markdown link redirect is more portable across tools.


rnetser (changes requested):
check claude `/memory`
you will see that the file is loaded
```
 L AGENTS.md              @-imported

```

PR #4589, opened 2026-04-27
docs: replace CLAUDE.md @import with a symlink to AGENTS.md

EdDev:
CLAUDE.md used an @AGENTS.md import directive to pull in project rules,
adding a wrapper with no extra content. Replacing it with a symlink
removes the indirection — both Claude and other AI tools now read the
same file directly.

##### jira-ticket:
NONE

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,7 +1,1 @@
-# CLAUDE.md
-
-Project guidance for AI coding assistants.
-
-> **Using a different AI tool?** See `AGENTS.md` for the tool-agnostic project rules.
-
-@AGENTS.md
+AGENTS.md


rnetser:
The current `CLAUDE.md` with `@AGENTS.md` already works — Claude Code resolves the `@` include directive and inlines the full content of `AGENTS.md`. This is Claude's native include syntax.

Replacing it with a symlink is a lateral move:
- **Symlinks can cause issues** with some CI tools, editors, and GitHub features (e.g., GitHub renders symlinks as plain text showing the target path, not the content).
- The current approach is safer and well-documented.
- The symlink also lacks a trailing newline.

I'd suggest keeping the current `@AGENTS.md` approach.

EdDev:
> The current `CLAUDE.md` with `@AGENTS.md` already works — Claude Code resolves the `@` include directive and inlines the full content of `AGENTS.md`. This is Claude's native include syntax.
> 
> Replacing it with a symlink is a lateral move:
> 
> * **Symlinks can cause issues** with some CI tools, editors, and GitHub features (e.g., GitHub renders symlinks as plain text showing the target path, not the content).
> * The current approach is safer and well-documented.
> * The symlink also lacks a trailing newline.
> 
> I'd suggest keeping the current `@AGENTS.md` approach.

The current approach requires to have this file managed. The intention to reduce to the minimum the AI tooling vendor lock, therefore avoiding claude specific instructions.
Keeping the vendor file editable is also a problem, as future changes may wrongly edit it with specific content.
Agents may override the file (e.g. claude agent with the `/init` instruction), diverging from the project preference.

IMO all 3 reasoning against it are very weak, however, I am not going to fight over this one.

EdDev:
There has been no agreement on this one, closing.

PR #4778, opened 2026-05-07
net, evpn: implement VM reboot connectivity test

servolkov:

Verify TCP connectivity to all EVPN endpoints (VM-to-VM, stretched L2, routed L3) is re-established after restarting the target VM.

##### jira-ticket: https://redhat.atlassian.net/browse/CNV-87077

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/tests/network/bgp/evpn/libevpn.py
+++ b/tests/network/bgp/evpn/libevpn.py
@@ -1,15 +1,23 @@
 import contextlib
+import ipaddress
 import logging
 import shlex
 from collections.abc import Generator
 from dataclasses import dataclass
 
 from ocp_resources.pod import Pod
+from pytest import Subtests
 from timeout_sampler import retry
 
 from libs.net.cluster import ipv4_supported_cluster, ipv6_supported_cluster
 from libs.net.ip import filter_link_local_addresses, random_ipv4_address, random_ipv6_address
-from libs.net.traffic_generator import IPERF_SERVER_PORT, PodTcpClient, TcpServer
+from libs.net.traffic_generator import (
+    IPERF_SERVER_PORT,
+    PodTcpClient,
+    TcpServer,
+    active_tcp_connections,
+    is_tcp_connection,
+)
 from libs.net.vmspec import lookup_iface_status, lookup_primary_network
 from libs.vm.vm import BaseVirtualMachine
 from tests.network.libs.bgp import CLUSTER_FRR_ASN, EXTERNAL_FRR_ASN, NET_TOOLS_CONTAINER_NAME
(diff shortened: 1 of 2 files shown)


tests/network/bgp/evpn/libevpn.py
```diff
@@ -230,3 +238,41 @@ def _evpn_workloads_connection(
...
         ) as tcp_client:
             yield tcp_client, tcp_server
+
+
+def assert_evpn_workloads_connectivity(
+    target_vm: BaseVirtualMachine,
+    ref_vm: BaseVirtualMachine,
+    l2_endpoint: EvpnEndpoint,
+    l3_endpoint: EvpnEndpoint,
+    subtests: Any,
```
EdDev: You should use `from pytest import SubTests`.

EdDev (changes requested):
Thanks!

Only one small request inline.

tests/network/bgp/evpn/libevpn.py
servolkov: overlooked, thanks. Updated.

EdDev (approved):
Thanks!

I guess this is temporary ack as you'll need to rebase.

/approve

servolkov:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/ed9366351551f0a4851b1adc1bbae1e800f80f47..3393ca53f8517b25d4bdaaf8be58d4915bdaa178): rebase on main

servolkov:
/verified

```
$ uv run pytest --cluster-sanity-skip-nodes-check --cluster-sanity-skip-storage-check --skip-deprecated-api-test --cluster-sanity-skip-check tests/network/bgp/evpn/ -k reboot
...
TEST: test_connectivity_after_udn_vm_cold_reboot STATUS: PASSED
```

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  **Analysis Requirements:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

</details>

tests/network/bgp/evpn/libevpn.py
```diff
@@ -1,15 +1,23 @@
 import contextlib
+import ipaddress
```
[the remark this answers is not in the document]
servolkov: done

tests/network/bgp/evpn/libevpn.py
```diff
@@ -361,3 +369,41 @@ def _evpn_workloads_connection(
...
+
+    with evpn_workloads_active_connections(endpoint=l2_endpoint, vm=target_vm) as l2_connections:
+        for l2_client, l2_server in l2_connections:
+            with subtests.test(f"stretched-L2 IPv{ipaddress.ip_address(l2_client.server_ip).version}"):
+                assert is_tcp_connection(server=l2_server, client=l2_client)
+
+    with evpn_workloads_active_connections(endpoint=l3_endpoint, vm=target_vm) as l3_connections:
+        for l3_client, l3_server in l3_connections:
+            with subtests.test(f"routed-L3 IPv{ipaddress.ip_address(l3_client.server_ip).version}"):
+                assert is_tcp_connection(server=l3_server, client=l3_client)
```
jpeimer, line 409: Shouldn't one test verify just one thing?

jpeimer (changes requested):

EdDev (approved):
Thank you!

/approve

tests/network/bgp/evpn/libevpn.py
servolkov: in net SIG we treat tests as end-to-end user scenarios. A test should represent the complete user scenario.

Our STD explicitly expects that: `New connections are established after the cold reboot.` The current implementation reflects this exact user flow. It is logical, self-documenting, and keeps the 'Act' (reboot) and the 'Assert' (connectivity checks) right next to each other.

We could certainly split this up:

- fully separate tests: this would require executing a cold reboot for every single network path check. It is highly resource-consuming, strains the infrastructure, and artificially inflates our test execution time.

- we could use a class fixture to reboot once and run dedicated test methods for each check. However, this breaks the user scenario narrative. It hides the 'Action' inside a setup fixture, making the code less self-descriptive. It also adds unnecessary architectural complexity by introducing shared state between tests.

yossisegev (approved):
Thank you Sergei

tests/network/bgp/evpn/libevpn.py
servolkov: @jpeimer please, let me know if you still have a question, or I can resolve the thread.

EdDev: @servolkov explanation is accurate.
This is testing one thing, which is `assert_evpn_workloads_connectivity`. The check includes several components, but they are not relevant as individuals to be considered a test.

I think we have provided all the needed information and see no reason to block this PR on this topic.
I would consider the root approver ack on this PR as a sign that we can resolve this thread.

jpeimer: Sorry for the late answer, coming back from a sick leave.
The Jira linked to this PR didn't have much info about what needs to be done, and didn't have the link to the mentioned STP (I could eventually find it in the other card).

So if network SIG is happy with this approach, and @EdDev approves it, I also don't see a reason to block this PR. 
Good job!

jpeimer (approved):

tests/network/bgp/evpn/libevpn.py
servolkov: thanks @jpeimer. I mentioned STD - it is a few lines above the test implementation (docstring with Preconditions, Steps and Expected)

Resolving.

yossisegev: > I would consider the root approver ack on this PR as a sign that we can resolve this thread.

Let's be careful with this approach. We claim that peer reviews are required even if root approval was already given, but this quite contradicts it. We don't want the reviews to be just an official rubber-stamp.

PR #4772, opened 2026-05-07
net: Introduce dual-stream linux bridge tests

azhivovk:
##### Short description:
Dual-stream clusters run worker nodes with a mix of RHCOS 9 and RHCOS 10.
Live migration across OS versions can expose compatibility issues: kernel bridge drivers, OVN flow tables, and binding plugins all behave differently between RHCOS releases, that would only manifest during cross-OS migration and not in homogeneous OS nodes clusters. 

##### What this PR does / why we need it:
This PR adds the shared infrastructure layer needed by all network migration tests in this series, plus Linux bridge connectivity tests.                        
The shared layer registers the mixed_os_nodes marker to gate tests to compatible clusters, and introduces helpers to direct a running VM to a specific OS node before triggering migration.

The Linux bridge tests verify that an active TCP connection over a secondary Linux bridge network survives live migration of the server VM in both directions (RHCOS 9 -> 10 and back). 
The client VM remains on RHCOS 9 throughout, so the test isolates the impact of the server-side OS change on the forwarding path.

##### Special notes for reviewer:
This is the first PR in a series. Localnet, primary network, and primary UDN are covered in separate PRs that depend on the same infra commits.  

## Summary by CodeRabbit

* **New Features**
  * Added testing framework for virtual machine migrations between RHEL9 and RHEL10 worker nodes.
  * Enhanced L2 bridge network configuration and testing across mixed-OS clusters.

* **Tests**
  * Implemented comprehensive test fixtures for L2 bridge integration testing.
  * Added automated network connectivity verification during VM migrations across operating system versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RedHatQE/openshift-virtualization-tests/pull/4772)

PR #4817, opened 2026-05-12
net, evpn: add source provider migration test

servolkov:
  Add source provider migration test for EVPN.

  - Add VNI-aware teardown for L2 endpoints: clean up VLAN/VNI bridge mappings,
    allowing re-deployment with different parameters in the same test session.
  - Add `macAddress` field to `Interface` dataclass for explicit MAC assignment.
  - Add explicit MAC support to `deploy_evpn_l2_endpoint`.
  - Add `test_source_provider_migration`: emulates migration of an external L2
    workload into the cluster as a CUDN VM, preserving its IP and MAC addresses.
    The old endpoint is torn down, a VM takes over the original identity, then a
    new L2 endpoint is deployed with different IPs to verify connectivity.

  - `test_source_provider_migration` uses `@pytest.mark.order("last")` because it
    tears down the shared L2 endpoint, which would break other tests if run earlier.
  - IPv6 static IP fix: `random_ipv6_address()` generates addresses with a random 7th hextet, 
    but OVN IPAM caps /64 allocation to 65536 addresses. 
    Replaced with a low-offset hardcoded address. Only affects IPs requested via 
    `network.kubevirt.io/addresses`; external netns endpoints are unaffected.
    The known bug: https://redhat.atlassian.net/browse/OCPBUGS-86810
  - The new L2 endpoint is created as a test step rather than a fixture. This is a 
    known limitation — currently only one L2 endpoint can exist at a time. A follow-up
    will reorganise fixtures to separate VLAN lifecycle from endpoints, enabling
    independent concurrent endpoints.

  ##### jira-ticket: https://redhat.atlassian.net/browse/CNV-89431

## Summary by CodeRabbit

* **New Features**
  * VM interfaces can include an explicit MAC address; L2 endpoints accept and apply a provided MAC.

* **Tests**
  * EVPN connectivity tests now perform a full source-provider migration end-to-end and run last.
  * IPv4/IPv6 endpoint addressing is deterministic for repeatable runs.
  * Tests inject endpoint MACs into VM templates and add a fixture to provision VMs attached to external L2 endpoints.

* **Bug Fixes**
  * L2 endpoint teardown now fully cleans bridge/VNI state.

### DIFF
--- a/libs/vm/spec.py
+++ b/libs/vm/spec.py
@@ -71,6 +71,7 @@ class Disk:
 @dataclass
 class Interface:
     name: str
+    macAddress: str | None = None  # noqa: N815
     masquerade: dict[Any, Any] | None = None
     bridge: dict[Any, Any] | None = None
     sriov: dict[Any, Any] | None = None
--- a/libs/vm/spec.py
+++ b/libs/vm/spec.py
@@ -77,6 +77,7 @@ class Interface:
     passtBinding: dict[Any, Any] | None = None  # noqa: N815
     binding: NetBinding | None = None
     state: str | None = None
+    macAddress: str | None = None  # noqa: N815
 
 
 @dataclass
--- a/tests/network/bgp/evpn/conftest.py
+++ b/tests/network/bgp/evpn/conftest.py
@@ -11,11 +11,14 @@ from ocp_resources.node import Node
 from ocp_resources.resource import ResourceEditor
 from ocp_resources.vtep import VTEP
 
-from libs.net.ip import random_ipv4_address, random_ipv6_address
+from libs.net.ip import random_ipv4_address
 from libs.net.traffic_generator import TcpServer
-from libs.net.udn import UDN_BINDING_DEFAULT_PLUGIN_NAME, create_udn_namespace
+from libs.net.udn import UDN_BINDING_DEFAULT_PLUGIN_NAME, create_udn_namespace, udn_primary_network
+from libs.vm.factory import base_vmspec, fedora_vm
+from libs.vm.spec import Devices
 from libs.vm.vm import BaseVirtualMachine
 from tests.network.bgp.evpn.libevpn import (
+    CUDN_EVPN_SUBNET_IPV6,
     EndpointTcpClient,
     EvpnEndpoint,
     cudn_evpn_subnets,
@@ -42,7 +45,10 @@ EVPN_ADVERTISE_LABEL: Final[dict] = {"advertise": "evpn"}
 APP_EVPN_CUDN_LABEL: Final[dict] = {**EVPN_ADVERTISE_LABEL, "app": "cudn-evpn"}
 CUDN_EVPN_BGP_LABEL: Final[dict] = {"cudn-bgp": "evpn"}
 EXTERNAL_L2_ENDPOINT_IPV4: Final[str] = f"{random_ipv4_address(net_seed=5, host_address=250)}/24"
-EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = f"{random_ipv6_address(net_seed=5, host_address=250)}/64"
+EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = (
+    f"{ipaddress.ip_network(CUDN_EVPN_SUBNET_IPV6, strict=False).network_address + 250}/64"
+)
+EXTERNAL_L2_ENDPOINT_MAC: Final[str] = "02:00:05:00:fa:00"
 EXTERNAL_L3_ENDPOINT_IPV4: Final[str] = "192.168.100.100/24"
 EXTERNAL_L3_ENDPOINT_IPV6: Final[str] = "fd01:1234:5678::64/64"
 EXTERNAL_L3_GATEWAY_IPV4: Final[str] = "192.168.100.1/24"
@@ -233,6 +239,7 @@ def external_l2_endpoint(
         pod=frr_external_pod.pod,
         vni=EVPN_MAC_VRF_VNI,
         endpoint_ips=[EXTERNAL_L2_ENDPOINT_IPV4, EXTERNAL_L2_ENDPOINT_IPV6],
+        mac_address=EXTERNAL_L2_ENDPOINT_MAC,
     )
     yield endpoint
     teardown_evpn_l2_endpoint(pod=frr_external_pod.pod, vni=EVPN_MAC_VRF_VNI)
@@ -269,3 +276,25 @@ def evpn_routed_l3_active_connections(
 ) -> Generator[list[tuple[EndpointTcpClient, TcpServer]]]:
     with evpn_workloads_active_connections(endpoint=external_l3_endpoint, vm=vm_evpn_target) as connections:
         yield connections
+
+
+@pytest.fixture()
+def vm_source_provider(
+    external_l2_endpoint: EvpnEndpoint,
+    namespace_evpn: Namespace,
+    cudn_evpn_layer2: libcudn.ClusterUserDefinedNetwork,
+    admin_client: DynamicClient,
+) -> Generator[BaseVirtualMachine]:
+    spec = base_vmspec()
+    iface, network = udn_primary_network(name="udn-primary", binding=UDN_BINDING_DEFAULT_PLUGIN_NAME)
+    iface.macAddress = external_l2_endpoint.mac_address
+    spec.template.spec.domain.devices = Devices(interfaces=[iface])
+    spec.template.spec.networks = [network]
+
+    with fedora_vm(namespace=namespace_evpn.name, name="vm-source-provider", client=admin_client, spec=spec) as vm:
+        vm.update_template_annotations(
+            template_annotations={
+                "network.kubevirt.io/addresses": json.dumps({"udn-primary": external_l2_endpoint.ip_addresses})
+            }
+        )
+        yield vm
--- a/tests/network/bgp/evpn/conftest.py
+++ b/tests/network/bgp/evpn/conftest.py
@@ -235,7 +235,7 @@ def external_l2_endpoint(
         endpoint_ips=[EXTERNAL_L2_ENDPOINT_IPV4, EXTERNAL_L2_ENDPOINT_IPV6],
     )
     yield endpoint
-    teardown_evpn_l2_endpoint(pod=frr_external_pod.pod)
+    teardown_evpn_l2_endpoint(pod=frr_external_pod.pod, vni=EVPN_MAC_VRF_VNI)
 
 
 @pytest.fixture(scope="module")
--- a/tests/network/bgp/evpn/conftest.py
+++ b/tests/network/bgp/evpn/conftest.py
@@ -20,6 +20,7 @@ from libs.vm.spec import Devices, Metadata
 from libs.vm.vm import BaseVirtualMachine
 from tests.network.bgp.evpn.libevpn import (
     CUDN_EVPN_SUBNET_IPV6,
+    EVPN_CUDN_NET_SEED,
     EndpointTcpClient,
     EvpnEndpoint,
     cudn_evpn_subnets,
@@ -45,7 +46,7 @@ from tests.network.libs.vm_factory import udn_vm
 EVPN_ADVERTISE_LABEL: Final[dict] = {"advertise": "evpn"}
 APP_EVPN_CUDN_LABEL: Final[dict] = {**EVPN_ADVERTISE_LABEL, "app": "cudn-evpn"}
 CUDN_EVPN_BGP_LABEL: Final[dict] = {"cudn-bgp": "evpn"}
-EXTERNAL_L2_ENDPOINT_IPV4: Final[str] = f"{random_ipv4_address(net_seed=5, host_address=250)}/24"
+EXTERNAL_L2_ENDPOINT_IPV4: Final[str] = f"{random_ipv4_address(net_seed=EVPN_CUDN_NET_SEED, host_address=250)}/24"
 EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = f"{ipaddress.ip_network(CUDN_EVPN_SUBNET_IPV6, strict=False)[250]}/64"
 EXTERNAL_L2_ENDPOINT_MAC: Final[str] = "02:00:05:00:fa:00"
 EXTERNAL_L3_ENDPOINT_IPV4: Final[str] = "192.168.100.100/24"
--- a/tests/network/bgp/evpn/conftest.py
+++ b/tests/network/bgp/evpn/conftest.py
@@ -11,11 +11,15 @@ from ocp_resources.node import Node
 from ocp_resources.resource import ResourceEditor
 from ocp_resources.vtep import VTEP
 
-from libs.net.ip import random_ipv4_address, random_ipv6_address
+from libs.net.ip import random_ipv4_address
 from libs.net.traffic_generator import TcpServer
-from libs.net.udn import UDN_BINDING_DEFAULT_PLUGIN_NAME, create_udn_namespace
+from libs.net.udn import UDN_BINDING_DEFAULT_PLUGIN_NAME, create_udn_namespace, udn_primary_network
+from libs.vm.affinity import new_pod_anti_affinity
+from libs.vm.factory import base_vmspec, fedora_vm
+from libs.vm.spec import Devices, Metadata
 from libs.vm.vm import BaseVirtualMachine
 from tests.network.bgp.evpn.libevpn import (
+    CUDN_EVPN_SUBNET_IPV6,
     EndpointTcpClient,
     EvpnEndpoint,
     cudn_evpn_subnets,
@@ -42,7 +46,8 @@ EVPN_ADVERTISE_LABEL: Final[dict] = {"advertise": "evpn"}
 APP_EVPN_CUDN_LABEL: Final[dict] = {**EVPN_ADVERTISE_LABEL, "app": "cudn-evpn"}
 CUDN_EVPN_BGP_LABEL: Final[dict] = {"cudn-bgp": "evpn"}
 EXTERNAL_L2_ENDPOINT_IPV4: Final[str] = f"{random_ipv4_address(net_seed=5, host_address=250)}/24"
-EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = f"{random_ipv6_address(net_seed=5, host_address=250)}/64"
+EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = f"{ipaddress.ip_network(CUDN_EVPN_SUBNET_IPV6, strict=False)[250]}/64"
+EXTERNAL_L2_ENDPOINT_MAC: Final[str] = "02:00:05:00:fa:00"
 EXTERNAL_L3_ENDPOINT_IPV4: Final[str] = "192.168.100.100/24"
 EXTERNAL_L3_ENDPOINT_IPV6: Final[str] = "fd01:1234:5678::64/64"
 EXTERNAL_L3_GATEWAY_IPV4: Final[str] = "192.168.100.1/24"
@@ -233,6 +238,7 @@ def external_l2_endpoint(
         pod=frr_external_pod.pod,
         vni=EVPN_MAC_VRF_VNI,
         endpoint_ips=[EXTERNAL_L2_ENDPOINT_IPV4, EXTERNAL_L2_ENDPOINT_IPV6],
+        mac_address=EXTERNAL_L2_ENDPOINT_MAC,
     )
     yield endpoint
     teardown_evpn_l2_endpoint(pod=frr_external_pod.pod, vni=EVPN_MAC_VRF_VNI)
@@ -269,3 +275,30 @@ def evpn_routed_l3_active_connections(
 ) -> Generator[list[tuple[EndpointTcpClient, TcpServer]]]:
     with evpn_workloads_active_connections(endpoint=external_l3_endpoint, vm=vm_evpn_target) as connections:
         yield connections
+
+
+@pytest.fixture()
+def vm_source_provider(
+    external_l2_endpoint: EvpnEndpoint,
+    namespace_evpn: Namespace,
+    cudn_evpn_layer2: libcudn.ClusterUserDefinedNetwork,
+    admin_client: DynamicClient,
+    frr_external_pod: ExternalFrrPodInfo,
+) -> Generator[BaseVirtualMachine]:
+    spec = base_vmspec()
+    network_name = "udn-network"
+    iface, network = udn_primary_network(name=network_name, binding=UDN_BINDING_DEFAULT_PLUGIN_NAME)
+    iface.macAddress = external_l2_endpoint.mac_address
+    spec.template.spec.domain.devices = Devices(interfaces=[iface])
+    spec.template.spec.networks = [network]
+    spec.template.metadata = Metadata(
+        labels=EXTERNAL_FRR_POD_LABEL,
+        annotations={"network.kubevirt.io/addresses": json.dumps({network_name: external_l2_endpoint.ip_addresses})},
+    )
+    label, *_ = EXTERNAL_FRR_POD_LABEL.items()
+    spec.template.spec.affinity = new_pod_anti_affinity(
+        label=label, namespaces=[frr_external_pod.pod.namespace, namespace_evpn.name]
+    )
+
+    with fedora_vm(namespace=namespace_evpn.name, name="vm-source-provider", client=admin_client, spec=spec) as vm:
+        yield vm
--- a/tests/network/bgp/evpn/libevpn.py
+++ b/tests/network/bgp/evpn/libevpn.py
@@ -169,11 +169,20 @@ def deploy_evpn_l2_endpoint(
     return EvpnEndpoint(pod=pod, ip_addresses=bare_ips, netns_name=_L2_ENDPOINT_NETNS)
 
 
-def teardown_evpn_l2_endpoint(pod: Pod) -> None:
-    """Removes the EVPN L2 endpoint (netns, veth) from the FRR pod."""
+def teardown_evpn_l2_endpoint(pod: Pod, vni: int) -> None:
+    """Removes the EVPN L2 endpoint (netns, veth, VLAN/VNI mappings) from the FRR pod.
+
+    Args:
+        pod: The FRR pod hosting the endpoint.
+        vni: MAC-VRF VNI used during deployment.
+    """
     for cmd in [
         f"ip netns delete {_L2_ENDPOINT_NETNS}",
         f"ip link delete {_L2_VETH_POD_SIDE}",
+        f"bridge vlan del dev {_VXLAN_NAME} vid {_L2_VID} tunnel_info id {vni}",
+        f"bridge vni del dev {_VXLAN_NAME} vni {vni}",
+        f"bridge vlan del dev {_VXLAN_NAME} vid {_L2_VID}",
+        f"bridge vlan del dev {_BRIDGE_NAME} vid {_L2_VID} self",
     ]:
         pod.execute(command=shlex.split(cmd), container=NET_TOOLS_CONTAINER_NAME, ignore_rc=True)
 
--- a/tests/network/bgp/evpn/libevpn.py
+++ b/tests/network/bgp/evpn/libevpn.py
@@ -51,6 +51,7 @@ class EvpnEndpoint:
     pod: Pod
     ip_addresses: list[str]
     netns_name: str
+    mac_address: str | None = None
 
 
 class EndpointTcpClient(PodTcpClient):
@@ -143,6 +144,7 @@ def deploy_evpn_l2_endpoint(
     pod: Pod,
     vni: int,
     endpoint_ips: list[str],
+    mac_address: str | None = None,
 ) -> EvpnEndpoint:
     """Creates a stretched L2 endpoint on the shared SVD bridge.
 
@@ -155,18 +157,19 @@ def deploy_evpn_l2_endpoint(
         pod: The FRR pod hosting the endpoint.
         vni: MAC-VRF VNI (must match CUDN's macVRF VNI).
         endpoint_ips: IPs with prefix length (e.g. ["10.0.5.250/24", "fd00::fa/64"]).
+        mac_address: Explicit MAC for the endpoint interface (locally-administered).
 
     Returns:
         EvpnEndpoint.
     """
-    commands = _build_l2_endpoint_commands(vni=vni, endpoint_ips=endpoint_ips)
+    commands = _build_l2_endpoint_commands(vni=vni, endpoint_ips=endpoint_ips, mac_address=mac_address)
     for command in commands:
         pod.execute(command=shlex.split(command), container=NET_TOOLS_CONTAINER_NAME)
 
     bare_ips = [ip.split("/")[0] for ip in endpoint_ips]
     LOGGER.info(f"EVPN L2 endpoint deployed: {bare_ips} in namespace {_L2_ENDPOINT_NETNS}")
 
-    return EvpnEndpoint(pod=pod, ip_addresses=bare_ips, netns_name=_L2_ENDPOINT_NETNS)
+    return EvpnEndpoint(pod=pod, ip_addresses=bare_ips, netns_name=_L2_ENDPOINT_NETNS, mac_address=mac_address)
 
 
 def teardown_evpn_l2_endpoint(pod: Pod, vni: int) -> None:
@@ -189,7 +192,12 @@ def teardown_evpn_l2_endpoint(pod: Pod, vni: int) -> None:
     LOGGER.info(f"EVPN L2 endpoint removed: namespace={_L2_ENDPOINT_NETNS}")
 
 
-def _build_l2_endpoint_commands(vni: int, endpoint_ips: list[str]) -> list[str]:
+def _build_l2_endpoint_commands(
+    vni: int,
+    endpoint_ips: list[str],
+    mac_address: str | None = None,
+) -> list[str]:
+    mac_cmd = [f"ip netns exec {_L2_ENDPOINT_NETNS} ip link set dev {_L2_VETH_EP_SIDE} address {mac_address}"]
     return [
         f"bridge vlan add dev {_BRIDGE_NAME} vid {_L2_VID} self",
         f"bridge vlan add dev {_VXLAN_NAME} vid {_L2_VID}",
@@ -202,6 +210,7 @@ def _build_l2_endpoint_commands(vni: int, endpoint_ips: list[str]) -> list[str]:
         f"ip netns add {_L2_ENDPOINT_NETNS}",
         f"ip link set {_L2_VETH_EP_SIDE} netns {_L2_ENDPOINT_NETNS}",
         *(f"ip netns exec {_L2_ENDPOINT_NETNS} ip addr add {ip} dev {_L2_VETH_EP_SIDE}" for ip in endpoint_ips),
+        *(mac_cmd if mac_address else []),
         f"ip netns exec {_L2_ENDPOINT_NETNS} ip link set {_L2_VETH_EP_SIDE} up",
         f"ip netns exec {_L2_ENDPOINT_NETNS} ip link set lo up",
     ]
--- a/tests/network/bgp/evpn/libevpn.py
+++ b/tests/network/bgp/evpn/libevpn.py
@@ -24,8 +24,9 @@ from tests.network.libs.bgp import CLUSTER_FRR_ASN, EXTERNAL_FRR_ASN, NET_TOOLS_
 
 LOGGER = logging.getLogger(__name__)
 
-CUDN_EVPN_SUBNET_IPV4: str = f"{random_ipv4_address(net_seed=5, host_address=0)}/24"
-CUDN_EVPN_SUBNET_IPV6: str = f"{random_ipv6_address(net_seed=5, host_address=0)}/64"
+EVPN_CUDN_NET_SEED: int = 5
+CUDN_EVPN_SUBNET_IPV4: str = f"{random_ipv4_address(net_seed=EVPN_CUDN_NET_SEED, host_address=0)}/24"
+CUDN_EVPN_SUBNET_IPV6: str = f"{random_ipv6_address(net_seed=EVPN_CUDN_NET_SEED, host_address=0)}/64"
 
 _BRIDGE_NAME: str = "br0"
 _VXLAN_NAME: str = "vxlan0"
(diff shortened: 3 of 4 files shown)


openshift-virtualization-qe-bot-5:

## Welcome! 🎉

This pull request will be automatically processed with the following features:

### 🔄 Automatic Actions
* **Reviewer Assignment**: Reviewers are automatically assigned based on the OWNERS file in the repository root
* **Size Labeling**: PR size labels (XS, S, M, L, XL, XXL) are automatically applied based on changes
* **Issue Creation**: A tracking issue is created for this PR and will be closed when the PR is merged or closed
* **Branch Labeling**: Branch-specific labels are applied to track the target branch
* **Auto-verification**: Auto-verified users have their PRs automatically marked as verified
* **Labels**: Enabled categories: `branch`, `can-be-merged`, `cherry-pick`, `has-conflicts`, `hold`, `needs-rebase`, `size`, `verified`, `wip`

### 📋 Available Commands

#### PR Status Management
* `/wip` - Mark PR as work in progress (adds WIP: prefix to title)
* `/wip cancel` - Remove work in progress status
* `/hold` - Block PR merging (approvers only)
* `/hold cancel` - Unblock PR merging
* `/verified` - Mark PR as verified
* `/verified cancel` - Remove verification status
* `/reprocess` - Trigger complete PR workflow reprocessing (useful if webhook failed or configuration changed)
* `/regenerate-welcome` - Regenerate this welcome message

#### Review & Approval
* `/lgtm` - Approve changes (looks good to me)
* `/approve` - Approve PR (approvers only)
* `/assign-reviewers` - Assign reviewers based on OWNERS file
* `/assign-reviewer @username` - Assign specific reviewer
* `/check-can-merge` - Check if PR meets merge requirements

#### Testing & Validation
 * `/retest tox` - Run Python test suite with tox
 * `/retest build-container` - Rebuild and test container image
 * `/retest verify-bugs-are-open` - verify-bugs-are-open
 * `/retest all` - Run all available tests


#### Container Operations
* `/build-and-push-container` - Build and push container image (tagged with PR number)
  * Supports additional build arguments: `/build-and-push-container --build-arg KEY=value`

#### Cherry-pick Operations
* `/cherry-pick <branch>` - Schedule cherry-pick to target branch when PR is merged
  * Multiple branches: `/cherry-pick branch1 branch2 branch3`

#### Label Management
* `/<label-name>` - Add a label to the PR
* `/<label-name> cancel` - Remove a label from the PR

### ✅ Merge Requirements

This PR will be automatically approved when the following conditions are met:

1. **Approval**: `/approve` from at least one approver
2. **LGTM Count**: Minimum 2 `/lgtm` from reviewers
3. **Status Checks**: All required status checks must pass
4. **No Blockers**: No wip, hold, has-conflicts labels and PR must be mergeable (no conflicts)
5. **Verified**: PR must be marked as verified

### 📊 Review Process

<details>
<summary><strong>Approvers and Reviewers</strong></summary>


**Approvers:**
 * EdDev
 * dshchedr
 * myakove
 * rnetser
 * vsibirsk


**Reviewers:**
 * Anatw
 * EdDev
 * RoniKishner
 * azhivovk
 * dshchedr
 * frenzyfriday
 * nirdothan
 * orelmisan
 * rnetser
 * servolkov
 * vsibirsk
 * yossisegev


</details>

<details>
<summary><strong>Available Labels</strong></summary>

 * hold
 * verified
 * wip
 * lgtm
 * approve

</details>

<details>
<summary><strong>AI Features</strong></summary>

* **Cherry-Pick Conflict Resolution**: Enabled (claude/claude-opus-4-6[1m])

</details>

### 💡 Tips

* **WIP Status**: Use `/wip` when your PR is not ready for review
* **Verification**: The verified label is removed on new commits unless the push is detected as a clean rebase
* **Cherry-picking**: Cherry-pick labels are processed when the PR is merged
* **Container Builds**: Container images are automatically tagged with the PR number
* **Permission Levels**: Some commands require approver permissions
* **Auto-verified Users**: Certain users have automatic verification and merge privileges

For more information, please refer to the project documentation or contact the maintainers.

tests/network/bgp/evpn/test_evpn_connectivity.py
```diff
@@ -20,11 +20,20 @@
...
+from libs.net.ip import random_ipv4_address, random_ipv6_address
 from libs.net.traffic_generator import active_tcp_connections, is_tcp_connection
 from libs.net.vmspec import lookup_primary_network
-from tests.network.bgp.evpn.libevpn import evpn_workloads_active_connections
+from tests.network.bgp.evpn.libevpn import (
+    assert_evpn_workloads_connectivity,
+    deploy_evpn_l2_endpoint,
+    evpn_workloads_active_connections,
+    teardown_evpn_l2_endpoint,
+)
```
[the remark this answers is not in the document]
servolkov: this is wip, once https://github.com/RedHatQE/openshift-virtualization-tests/pull/4778 is merged, the PR will be rebased. Now it is for review only.

tests/network/bgp/evpn/conftest.py
```diff
@@ -42,7 +44,8 @@
 APP_EVPN_CUDN_LABEL: Final[dict] = {**EVPN_ADVERTISE_LABEL, "app": "cudn-evpn"}
 CUDN_EVPN_BGP_LABEL: Final[dict] = {"cudn-bgp": "evpn"}
 EXTERNAL_L2_ENDPOINT_IPV4: Final[str] = f"{random_ipv4_address(net_seed=5, host_address=250)}/24"
-EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = f"{random_ipv6_address(net_seed=5, host_address=250)}/64"
+EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = "fd00:1234:5678::fa/64"
+EXTERNAL_L2_ENDPOINT_MAC: Final[str] = "02:00:05:00:fa:00"
```
yossisegev: Is it safe that the MAC and IP are hard-coded? Aren't we risking to have addresses collisions?

yossisegev (changes requested):

tests/network/bgp/evpn/conftest.py
servolkov: not sure what kind of risk you are describing. Could you elaborate on what collisions you mean?

tests/network/bgp/evpn/test_evpn_connectivity.py
```diff
@@ -181,6 +198,23 @@ def test_source_provider_migration():
     Expected:
     - New connections are established after new UDN VM deployment.
     """
-
-
-test_source_provider_migration.__test__ = False
+    mac_vrf_vni = cudn_evpn_layer2.instance.spec.network.evpn.macVRF.vni
+
+    teardown_evpn_l2_endpoint(pod=frr_external_pod.pod, vni=mac_vrf_vni)
```
Anatw, line 218: This feels wrong - is the teardown fails - the test fails - but this scenario doesn't test the teardown of the l2 endpoint. I think it should live in a fixture called by the test as the setup.

tests/network/bgp/evpn/test_evpn_connectivity.py
```diff
@@ -163,7 +172,15 @@ def test_connectivity_after_udn_vm_cold_reboot():
 
 
 @pytest.mark.polarion("CNV-15233")
-def test_source_provider_migration():
+@pytest.mark.order("last")
```
Anatw, line 189: This feels like a hack, why can't the fixtures creating the setup clean it after yield. This places wrong order responsibility as I understand it, but maybe I'm missing something.

tests/network/bgp/evpn/conftest.py
```diff
@@ -42,7 +44,8 @@
 APP_EVPN_CUDN_LABEL: Final[dict] = {**EVPN_ADVERTISE_LABEL, "app": "cudn-evpn"}
 CUDN_EVPN_BGP_LABEL: Final[dict] = {"cudn-bgp": "evpn"}
 EXTERNAL_L2_ENDPOINT_IPV4: Final[str] = f"{random_ipv4_address(net_seed=5, host_address=250)}/24"
-EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = f"{random_ipv6_address(net_seed=5, host_address=250)}/64"
+EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = "fd00:1234:5678::fa/64"
```
Anatw: Why not using _random_ipv6_address_() here like we did for the IPv4?

tests/network/bgp/evpn/conftest.py
yossisegev: @servolkov the risk I am talking about is if any of these addresses are already assigned somewhere around the network.
You probably remember we had these occasional conflicts with IPv4, that's why we replaced 10.200.x.y with 172.16.x.y in PR like https://github.com/RedHatQE/openshift-virtualization-tests/pull/2479.
So if you know for sure that these are unique addresses that could not be conflicted then say it here, and I'll dismiss this comment, but otherwise we probably need to generate addresses from "safe" ranges.

tests/network/bgp/evpn/test_evpn_connectivity.py
```diff
@@ -163,7 +172,15 @@ def test_connectivity_after_udn_vm_cold_reboot():
...
+def test_source_provider_migration(
```
Anatw, line 190: Suggestion - move this test to its own module, restructure the setup into a fixture chain, and the test body becomes just the assertion. This solves the double teardown of the L2 endpoint, removes the @pytest.mark.order("last") dependency, and ensures the new L2 endpoint (new_l2_endpoint) gets proper fixture-based cleanup instead of being left behind.

tests/network/bgp/evpn/conftest.py
```diff
@@ -269,3 +273,25 @@ def evpn_routed_l3_active_connections(
 ) -> Generator[list[tuple[EndpointTcpClient, TcpServer]]]:
     with evpn_workloads_active_connections(endpoint=external_l3_endpoint, vm=vm_evpn_target) as connections:
         yield connections
+
+
+@pytest.fixture()
+def vm_source_provider(
```
Anatw, line 282: vm_source_provider inlines the same spec-building logic that udn_vm already provides — the only addition is the MAC address. Consider adding an optional mac_address parameter to udn_vm and calling it from the fixture, same pattern as vm_evpn_target.

If you do decide to keep both udn_vm and this fixture - both of them hardcode "udn-primary" - it's worth extracting to a constant or a default parameter on udn_primary_network so they can't silently diverge.

tests/network/bgp/evpn/conftest.py
yossisegev: @servolkov If what I said above is still not clear then let's continue in a quick F2F to avoid the back-and-forth.

tests/network/bgp/evpn/test_evpn_connectivity.py
servolkov: Restructuring into a fixture chain doesn't simplify things — the STD defines teardown/deploy/reconnect as sequential test steps, not setup. Wrapping each step in a fixture just obscures the workflow. Explain clearly what you don't understand regarding this scenario. Now it looks quite straightforward per my understanding. 

The only "not good" thing - deploying new l2 endpoint as a test step. But I described in the description why so - will be fixed in a follow up. Let me know your thoughts.

tests/network/bgp/evpn/conftest.py
yossisegev: @servolkov You didn't answer, so I don't know if you keep this because you are sure it is safe, or because you didn't address it yet.

servolkov: The MAC and IPs are safe from collisions here:

- MAC 02:00:05:00:fa:00 has the locally-administered bit set (02: prefix) — it's explicitly designated for local/test use.
- IPs — both IPv4 and IPv6 are within the CUDN subnet, CUDN subnet is generated one form our helpers. Moreover, both CUDN and emulated endpoints are isolated by design, we can use any IP there.

servolkov: >@servolkov You didn't answer, so I don't know if you keep this because you are sure it is safe, or because you didn't address it yet.

sorry, I did not submit my comments, see the answer above.

tests/network/bgp/evpn/conftest.py
servolkov: I added the explanation of this change to the PR description (see `Special notes for reviewer:` section). Please, let me know if it's clear.

tests/network/bgp/evpn/conftest.py
servolkov: Agree that it looks like duplication, but `mac_address` is a low-level VM spec detail that doesn't belong in udn_vm's API signature. udn_vm is about creating a VM with a UDN primary network — not about fine-tuning interface hardware addresses. Adding it as parameter would bloat the function signature for a single use case.

The current approach — build the spec manually, set `iface.macAddress`, pass to `fedora_vm` — is ~5 lines and makes the MAC handling explicit and isolated. This is the right way how to setup mac address properly for now without refactoring a general helper. `udn_vm` needs refactoring to accept a pre-built spec or interface customization, that's a separate effort. BTW I have a plan to refactor this helper as a follow-up.

WDYT?

tests/network/bgp/evpn/test_evpn_connectivity.py
servolkov: The teardown of the old endpoint is a test step, not setup (see STD).

The test has a specific sequential workflow: these steps are interdependent and tell the migration story. If you split step 1 into a fixture, you'd create a fixture chain that mirrors the test steps 1:1 but obscures the workflow.

If teardown itself fails — yes, the test fails, but that's correct: if you can't remove the old endpoint, the "migration" scenario is broken and the test should fail.

tests/network/bgp/evpn/test_evpn_connectivity.py
servolkov: The test belongs in `test_evpn_connectivity.py`— it tests EVPN connectivity in a migration scenario. Its precondition is that L2 EVPN connectivity works, which is what the other tests in this module verify. The order("last") marker cleanly expresses the one constraint this test has. Per my vision this marker is very simple and harmless solution. Otherwise, as a precondition for this test we have to repeat all basic connectivity scenarios.

tests/network/bgp/evpn/conftest.py
yossisegev: > IPs — both IPv4 and IPv6 are within the CUDN subnet, CUDN subnet is generated one form our helpers. 

In that case , I think it would be safer and better practice to import `_IPV6_ADDRESS_SUBNET_PREFIX_VMI` from [libs/net/ip.py](https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/libs/net/ip.py#L11) (and remove its '_', prefix as it will no longer be used only in that module).

tests/network/bgp/evpn/conftest.py
Anatw: I read the explanation in the PR description — the OVN IPAM /64 limitation makes sense.
That said, random_ipv6_address() works fine everywhere else (localnet, sriov, l2_bridge, connectivity, stuntime) because those set the IP via cloud-init directly on the interface — no IPAM involved. The issue is specific to the OVN IPAM path (network.kubevirt.io/addresses), not the helper itself.
So this is a workaround for the IPAM context. If another test needs OVN-allocated IPv6 in the future, it'll hit the same issue and someone will have to rediscover why. How about adding a comment explaining why this is hardcoded, so that future readers won't think this was an oversight?

tests/network/bgp/evpn/conftest.py
Anatw: Makes sense, thanks for the explanation.
I think a constant for "udn-primary"  is needed - to make sure the two hardcoded places don't diverge.

tests/network/bgp/evpn/conftest.py
servolkov: makes sense, but I did it a little differently, take a look.

tests/network/bgp/evpn/conftest.py
servolkov: >I think a constant for "udn-primary" is needed - to make sure the two hardcoded places don't diverge.

Again, I don't want to touch general helper now with inappropriate changes. Even creating a constant causes the helper change, that in reality we don't need. Lets wait for the `udn_vm` follow-up refactoring. Having a few lines of duplication not high cost IMO. 

If you disagree, lets wait for other opinions.

servolkov:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/e9d6b87ae4aa0e57cd81005d2f06dc4a7ce8288c..e97b3ea46de6728254da163c0927ddb94edb70cf): get IPv6 from range, [thread](https://github.com/RedHatQE/openshift-virtualization-tests/pull/4817#discussion_r3226261731)

tests/network/bgp/evpn/conftest.py
```diff
@@ -42,7 +45,10 @@
 APP_EVPN_CUDN_LABEL: Final[dict] = {**EVPN_ADVERTISE_LABEL, "app": "cudn-evpn"}
 CUDN_EVPN_BGP_LABEL: Final[dict] = {"cudn-bgp": "evpn"}
 EXTERNAL_L2_ENDPOINT_IPV4: Final[str] = f"{random_ipv4_address(net_seed=5, host_address=250)}/24"
-EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = f"{random_ipv6_address(net_seed=5, host_address=250)}/64"
+EXTERNAL_L2_ENDPOINT_IPV6: Final[str] = (
+    f"{ipaddress.ip_network(CUDN_EVPN_SUBNET_IPV6, strict=False).network_address + 250}/64"
```
yossisegev: Why is the 250 addition to the network prefix?

tests/network/bgp/evpn/test_evpn_connectivity.py
Anatw: Ack, thank you for the explanation.

tests/network/bgp/evpn/test_evpn_connectivity.py
Anatw: I understand the constraint, but order("last") is implicit — it doesn't say what this test depends on. Consider using @pytest.mark.dependency(depends=[...]) instead - it makes the coupling explicit and has another major advantages of skipping this scenario if the test it depends on didn't pass.

tests/network/bgp/evpn/test_evpn_connectivity.py
Anatw: Ack about leaving the test in the current module.
The pytest.mark.dependency suggestion from my other comment covers the ordering concern.
For the new L2 endpoint as a test step - ack, let's address it in the follow-up.

tests/network/bgp/evpn/conftest.py
Anatw: I hear you, but this PR introduces the second copy of "udn-primary", which today only exists in udn_vm. The natural time to extract a constant is when the duplication is introduced.
That said, not against doing it as a follow-up. Let's see what others think.

tests/network/bgp/evpn/conftest.py
servolkov: I don't understand the question. I preserve the same host address as it was in previous implementation. In general it can be any. IPV4 has same host address. What I am missing?

tests/network/bgp/evpn/test_evpn_connectivity.py
servolkov: > it doesn't say what this test depends on.

this test doesn't depend on any specific test's outcome, it depends on running after all of them,
because it tears down the shared endpoint. The real constraint is "run after everything else," not "run after test X." `order("last")` says exactly that.

tests/network/bgp/evpn/test_evpn_connectivity.py
servolkov: >The pytest.mark.dependency suggestion from my other comment covers the ordering concern.

It does not because it does not depend on a single test. Lets continue [here](https://github.com/RedHatQE/openshift-virtualization-tests/pull/4817/changes#r3227350171).

tests/network/bgp/evpn/conftest.py
EdDev:
```suggestion
    f"{ipaddress.ip_network(CUDN_EVPN_SUBNET_IPV6, strict=False)[250]}/64"
```

openshift-virtualization-qe-bot-3:
/retest all

_Auto-triggered: Files in this PR were modified by merged PR #4772._

<details>
<summary>Overlapping files</summary>

libs/vm/spec.py

</details>

tests/network/bgp/evpn/conftest.py
yossisegev: After having an offline discussion about this, I understand that this is safe and there's no risk of exceeding the allowed network address range.
Please reaolve.

tests/network/bgp/evpn/conftest.py
yossisegev: IIUC, `udn-primary` is just an arbitrary name for an interface (and its corresponding network) in a VM, right?
In that case, I don't think there's a need to hard-code it.
If at all, then I would store it locally in a var here in the function, because it referenced both for the network/interface and in the `addresses` annotation, which should specify, IIUC, the interface name. In this case it would make sense to do something like
```
udn_primary_name = "udn-primary"
...
iface, network = udn_primary_network(name=udn_primary_name, binding=UDN_BINDING_DEFAULT_PLUGIN_NAME)
...
            template_annotations={
                "network.kubevirt.io/addresses": json.dumps({udn_primary_name: external_l2_endpoint.ip_addresses})
            }
```

tests/network/bgp/evpn/test_evpn_connectivity.py
yossisegev: I agree with @servolkov reasoning here, because the need for this marker is clear from the PR description. However, maybe if you add a short documentation to the test that explains why it must be the last (just like you did in the PR message) then it would help clarify why this explicit ordering is valid.

yossisegev (approved):

tests/network/bgp/evpn/conftest.py
```diff
@@ -269,3 +276,25 @@ def evpn_routed_l3_active_connections(
...
+    admin_client: DynamicClient,
+) -> Generator[BaseVirtualMachine]:
+    spec = base_vmspec()
+    iface, network = udn_primary_network(name="udn-primary", binding=UDN_BINDING_DEFAULT_PLUGIN_NAME)
+    iface.macAddress = external_l2_endpoint.mac_address
+    spec.template.spec.domain.devices = Devices(interfaces=[iface])
+    spec.template.spec.networks = [network]
+
+    with fedora_vm(namespace=namespace_evpn.name, name="vm-source-provider", client=admin_client, spec=spec) as vm:
+        vm.update_template_annotations(
```
EdDev: Why do you need to patch it instead of creating the VM with it in the first place?
You can pass annotations through the `spec.template.metadata`.

tests/network/bgp/evpn/test_evpn_connectivity.py
EdDev: - I agree it is not great.
- It seems the best options I could think of at the moment. I do not have any better elegant alternative.
- As @yossisegev mentioned, it may be useful to document on why this must be executed last in this module. In the commit message, this is a must IMO. You should explain well this speciall case, so future us will be able to understand what happened here.

tests/network/bgp/evpn/test_evpn_connectivity.py
EdDev: @Anatw , the discussion AFAIU moved to the other thread, can this one be resolved?

EdDev (dismissed):
Thank you.

I had some little things and I think it will be useful to document the less trivial part with the test order.

tests/network/bgp/evpn/conftest.py
servolkov: indeed, fixed.

tests/network/bgp/evpn/conftest.py
servolkov: changed

tests/network/bgp/evpn/conftest.py
servolkov: @yossisegev our discussion with @Anatw is about different thing. To resolve this thread:

1) @yossisegev your suggestion applied, done
2) @Anatw I changed the network name not to bother you with the same naming as you see in `udn_vm`.

tests/network/bgp/evpn/test_evpn_connectivity.py
servolkov: It is explained already in PR description (that is supposed to be a commit message). Commit message updated accordingly. Having a comment in the code - I am not sure we need it. The line is simple to understand in terms of WHAT. If WHY is needed - git commit message will tell the story.

rnetser: @servolkov 
> Its precondition is that L2 EVPN connectivity works, which is what the other tests in this module verify.

you already have the preconditions; normally preconditions are translated to setup of the test.
as in this case, the preconditions rely on previous tests, it should be noted in the test's precondition. this will also help clarify why `@pytest.mark.order("last")` is used

servolkov: > the preconditions rely on previous tests, it should be noted in the test's precondition

do you mean I have to add a real step to STD like: "all (or some of them) previous tests passed"? If yes, I don't like it. `last` marker is just an implementation detail of `TCP connectivity exists between the connectivity reference VM and the external L2 and L3 endpoints`. Based on your suggestion, every test in the class marked with `incremental` should contain an STD step: "check the previous test passed".

But probably I misunderstood you.

rnetser: i mean that the prerequisites need to include what is exepcted as the test setup. in this case, based on you comment it is
> Its precondition is that L2 EVPN connectivity works

so this should be a precondition (and next to it you should add that this precondition is the outcome of any preceeding tests)

and if tests rely on previous tests results, then yes, you should use `incremental` (or dependency where applicable)

servolkov: > so this should be a precondition

but it is, I don't get you. See STD. 

Precondition:
`- TCP connectivity exists between the connectivity reference VM and the external L2 and L3 endpoints.`

Implementation:
`@pytest.mark.order("last")`

Please, tell me directly what I have to fix per your vision.

rnetser: as the precondition is the success of any previous test it should be noted next to the precondition, for example:

```
Precondition:
- TCP connectivity exists between the connectivity reference VM and the external L2 and L3 endpoints. Precondition is verified in preceeding tests.
```

bw - if someone runs only this test, the precondition is never really checked.
if this is indeed a precondition for the test, it should have a dependecy on another test.
setting `last` only means that the test will be executed last in this module but if it will fail on the precondition, it will require checking the code to understandthis hidden dependency.
i.e jut seeing this test and previous test fail, does not imply the connection between the 2

servolkov: the commentary has been added to test steps.

libs/vm/spec.py
```diff
@@ -71,6 +71,7 @@ class Disk:
 @dataclass
 class Interface:
     name: str
+    macAddress: str | None = None  # noqa: N815
     masquerade: dict[Any, Any] | None = None
```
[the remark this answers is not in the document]
servolkov: does not affect any usage, I will consider the fix if I need to push smth else in this PR.

libs/vm/spec.py
```diff
@@ -71,6 +71,7 @@ class Disk:
 @dataclass
 class Interface:
     name: str
+    macAddress: str | None = None  # noqa: N815
```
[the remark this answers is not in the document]
servolkov: and what is your suggestion then?

tests/network/bgp/evpn/conftest.py
```diff
@@ -269,3 +275,30 @@ def evpn_routed_l3_active_connections(
...
+    spec.template.spec.domain.devices = Devices(interfaces=[iface])
+    spec.template.spec.networks = [network]
+    spec.template.metadata = Metadata(
+        labels=EXTERNAL_FRR_POD_LABEL,
+        annotations={"network.kubevirt.io/addresses": json.dumps({network_name: external_l2_endpoint.ip_addresses})},
+    )
+    label, *_ = EXTERNAL_FRR_POD_LABEL.items()
+    spec.template.spec.affinity = new_pod_anti_affinity(
+        label=label, namespaces=[frr_external_pod.pod.namespace, namespace_evpn.name]
+    )
```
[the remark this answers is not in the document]
servolkov: `label, *_ = EXTERNAL_FRR_POD_LABEL.items()` unpacks the dict items and assigns the first tuple ('role', 'frr-external') to label. That's exactly what new_pod_anti_affinity expects — its signature is label: 
  tuple[str, str] and it unpacks with (key, value) = label.

tests/network/bgp/evpn/test_evpn_connectivity.py
```diff
@@ -185,6 +202,7 @@ def test_source_provider_migration():
     - External Source Provider L2 and L3 endpoints.
     - Running connectivity reference VM with a primary EVPN-enabled CUDN.
     - TCP connectivity exists between the connectivity reference VM and the external L2 and L3 endpoints.
+      Precondition is verified in preceeding tests.
```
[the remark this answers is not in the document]
servolkov: done

servolkov:
/verified

```
$ uv run pytest --cluster-sanity-skip-nodes-check --cluster-sanity-skip-storage-check --skip-deprecated-api-test --cluster-sanity-skip-check tests/network/bgp/evpn
...
collected 8 items / 1 deselected / 7 selected
...
7 passed, 1 deselected, 119 warnings
```

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  **Analysis Requirements:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

</details>

libs/vm/spec.py
```diff
@@ -77,6 +77,7 @@ class Interface:
     passtBinding: dict[Any, Any] | None = None  # noqa: N815
     binding: NetBinding | None = None
     state: str | None = None
+    macAddress: str | None = None  # noqa: N815
```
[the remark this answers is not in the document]
servolkov: done

tests/network/bgp/evpn/test_evpn_connectivity.py
Anatw: Yes, please resolve.

tests/network/bgp/evpn/conftest.py
Anatw: Great, thank you!

tests/network/bgp/evpn/test_evpn_connectivity.py
Anatw: Thanks, the added docstring note makes this clearer.
Regarding @rnetser's point about the hidden dependency - this is what I was getting at with the _@pytest.mark.dependency_ suggestion. If you depend on a specific preceding scenario, mark it explicitly; if you depend on all of them passing, mark the class as _incremental_. Either way, the coupling becomes visible at runtime instead of just in the docstring. If you agree I think this could be done as a follow-up as well, since the docstring covers this concern in case we have a failure.

Anatw (approved):
Thank you, Sergei.

tests/network/bgp/evpn/conftest.py
servolkov: resolving due to approval

yossisegev (approved):

openshift-virtualization-qe-bot-5:
No pre-commit.ci configured for this repository

servolkov:
Rebase to get proper labels

tests/network/bgp/evpn/conftest.py
```diff
@@ -269,3 +275,30 @@ def evpn_routed_l3_active_connections(
...
+    namespace_evpn: Namespace,
+    cudn_evpn_layer2: libcudn.ClusterUserDefinedNetwork,
+    admin_client: DynamicClient,
+    frr_external_pod: ExternalFrrPodInfo,
+) -> Generator[BaseVirtualMachine]:
+    spec = base_vmspec()
+    network_name = "udn-network"
+    iface, network = udn_primary_network(name=network_name, binding=UDN_BINDING_DEFAULT_PLUGIN_NAME)
+    iface.macAddress = external_l2_endpoint.mac_address
+    spec.template.spec.domain.devices = Devices(interfaces=[iface])
```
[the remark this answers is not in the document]
servolkov: no issue here, there are no boot disks or smth else.

servolkov:
/verified 

[this](https://github.com/RedHatQE/openshift-virtualization-tests/pull/4817#issuecomment-4554231301)

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  **Analysis Requirements:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

</details>

libs/vm/spec.py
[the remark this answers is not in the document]
servolkov: done

yossisegev (approved):

Anatw (approved):

openshift-virtualization-qe-bot-3:
/retest all

_Auto-triggered: Files in this PR were modified by merged PR #5041._

<details>
<summary>Overlapping files</summary>

tests/network/bgp/evpn/test_evpn_connectivity.py

</details>

openshift-virtualization-qe-bot-3:
/retest all

_Auto-triggered: Files in this PR were modified by merged PR #4954._

<details>
<summary>Overlapping files</summary>

tests/network/bgp/evpn/libevpn.py

</details>

openshift-virtualization-qe-bot-3:
/retest all

_Auto-triggered: Files in this PR were modified by merged PR #5014._

<details>
<summary>Overlapping files</summary>

libs/vm/spec.py

</details>

tests/network/bgp/evpn/libevpn.py
```diff
@@ -155,32 +157,47 @@ def deploy_evpn_l2_endpoint(
...
     LOGGER.info(f"EVPN L2 endpoint removed: namespace={_L2_ENDPOINT_NETNS}")
 
 
-def _build_l2_endpoint_commands(vni: int, endpoint_ips: list[str]) -> list[str]:
+def _build_l2_endpoint_commands(
+    vni: int,
+    endpoint_ips: list[str],
+    mac_address: str | None = None,
+) -> list[str]:
+    mac_cmd = [f"ip netns exec {_L2_ENDPOINT_NETNS} ip link set dev {_L2_VETH_EP_SIDE} address {mac_address}"]
```
qinqon: Just inline this with the rest, it's weird to have the fstring here and the conditional down.

tests/network/bgp/evpn/test_evpn_connectivity.py
```diff
@@ -20,11 +20,20 @@
...
+from tests.network.bgp.evpn.libevpn import (
+    assert_evpn_workloads_connectivity,
+    deploy_evpn_l2_endpoint,
+    evpn_workloads_active_connections,
+    teardown_evpn_l2_endpoint,
+)
 from utilities.virt import migrate_vm_and_verify
 
+_L2_ENDPOINT_IPV4: str = f"{random_ipv4_address(net_seed=5, host_address=249)}/24"
+_L2_ENDPOINT_IPV6: str = f"{random_ipv6_address(net_seed=5, host_address=249)}/64"
```
qinqon: Can we jus constants fro net_seed and host_address ? so we can explain why we use those at their docstring, btw I think there are other parts of the code that use net_seed=5 maybe we should re-use them with different host_address ?

tests/network/bgp/evpn/test_evpn_connectivity.py
```diff
@@ -176,7 +185,15 @@ def test_connectivity_after_udn_vm_cold_reboot(
 
 
 @pytest.mark.polarion("CNV-15233")
-def test_source_provider_migration():
+@pytest.mark.order("last")
```
qinqon, line 189: Isn't this problematic ? this means one test depends on other test, how this work if we filter to run just this test ?

qinqon (changes requested):

tests/network/bgp/evpn/libevpn.py
servolkov: it is not beautiful enough, but I don't mind, fixed.

tests/network/bgp/evpn/test_evpn_connectivity.py
servolkov: Added constant for seed. host_address might be any (as well as seed, but seed is shared between CUDN and L2 endpoint). So, having constant for host_address does not make any sense.

tests/network/bgp/evpn/test_evpn_connectivity.py
servolkov: It was discussed in [this thread](https://github.com/RedHatQE/openshift-virtualization-tests/pull/4817/changes#r3227350171), take a look. There is no issue if this test is the only one to run. Also please check the PR description, there is an explanation there why so. Let me know if it's unclear.

servolkov:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/pull/4817/changes/b827d09f389303c6c3fed89351556d674225fe66): 
1) const for EVPN CUDN NET SEED to reuse in L2 enpoints
2) mac cmd inline

servolkov:
/verified

[this](https://github.com/RedHatQE/openshift-virtualization-tests/pull/4817#issuecomment-4554231301)

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  **Analysis Requirements:**

         never imports B directly.

  **Output rules:**

  **Your deliverable:**

  **Test Execution Plan**

  **Guidelines:**

</details>

qinqon (approved):

tests/network/bgp/evpn/conftest.py
```diff
@@ -269,3 +276,30 @@ def evpn_routed_l3_active_connections(
...
+    spec.template.spec.networks = [network]
+    spec.template.metadata = Metadata(
+        labels=EXTERNAL_FRR_POD_LABEL,
+        annotations={"network.kubevirt.io/addresses": json.dumps({network_name: external_l2_endpoint.ip_addresses})},
+    )
+    label, *_ = EXTERNAL_FRR_POD_LABEL.items()
+    spec.template.spec.affinity = new_pod_anti_affinity(
+        label=label, namespaces=[frr_external_pod.pod.namespace, namespace_evpn.name]
+    )
```
maiqueb, line 302: we are setting this anti-affinity rule so the vm source provider doesn't land on the node where the external FRR pod is running, right ?

maiqueb (approved):
tks

tests/network/bgp/evpn/conftest.py
servolkov: right

Anatw (approved):
Thank you

openshift-virtualization-qe-bot-2:

openshift-virtualization-qe-bot-5:
Cherry-picked PR net, evpn: add source provider migration test into cnv-4.22: https://github.com/RedHatQE/openshift-virtualization-tests/pull/5206

PR #5230, opened 2026-06-15
fix: pytest 9.1.0 compatibility — nmstate marker and config.inicfg

rnetser:
Fixes two pytest 9.1.0 compatibility issues that block PR #5216 (lock file maintenance):

1. **`nmstate` marker registration** — pytest 9.1.0 fixes a 9.0 regression where `--strict-markers` in `addopts` was silently ignored (#14442). `utilities/unittests/pytest.ini` has `--strict-markers` but did not register the `nmstate` marker, causing `test_adds_nmstate_marker` to fail.

2. **`config.inicfg` deprecation** — `config.inicfg` is deprecated in pytest 9.1 and will be removed in pytest 10 (#13946). Replaced with `config.getini("testpaths")` in `utilities/data_collector.py`.

Fixes tox failure in #5216

- Adds `nmstate` marker to `utilities/unittests/pytest.ini`
- Replaces `config.inicfg.get("testpaths")` with `config.getini(name="testpaths")` (returns `list[str]` directly)
- Removes the `ConfigValue` workaround test that was needed for pytest 9.0

##### jira-ticket:

Assisted-by: Claude <noreply@anthropic.com>


rnetser:
Closing — changes will be pushed to the renovate PR #5216 branch directly.

PR #5285, opened 2026-06-18
Handle missing cpu-arch config gracefully in pytest fixture

yossisegev:
The current implementation of checking the `cpu-arch` argument only supports a single value, however for multi-arch tests it is required to pass a list of architectures (`--cpu-arch=amd64,arm64`)."
The fixture now uses dict.get() instead of direct indexing to handle missing cpu_arch gracefully.
`cpu_arch` is not set for multi-arch cluster runs now (by design), and this fix handles the KeyError failure.

Assisted-by: Claude Code <noreply@anthropic.com>

## Summary of changes

* **New Features**
  * Added architecture-aware selection for Fedora container images and container image digests, ensuring VM disk/volume provisioning and images match the resolved VM architecture (including multi-arch behavior).

* **Bug Fixes**
  * Improved cluster CPU architecture reporting to use the cluster configuration directly and avoid missing/undefined values.

* **Tests**
  * Updated fixtures to handle optional CPU architecture configuration safely when the key is absent.

PR #5349, opened 2026-06-23
net, sriov: memory hotplug tests

azhivovk:
Add SR-IOV VM memory hot-plug tests
During SR-IOV live migration, the NIC is hot-unplugged on
the source and re-attached on the target. A virt-handler
memory miscalculation caused re-attachment to fail, which
for the user manifested as loss of network connectivity.

Tests verify memory hot-plug completes without connectivity
loss and that guest memory is increased as expected.

The scenario is motivated by observed NIC loss after live migration triggered by memory hot-plug on SR-IOV VMs
https://redhat.atlassian.net/browse/CNV-69778
https://redhat.atlassian.net/browse/CNV-69974

New tests will fail in CI due to `CNV-91360: duplicate podnetinfo volume injected prevents virt-launcher pod creation` as many secondary network tests, but pass when net resources injector is disabled

## Summary by CodeRabbit

* **New Features**
  * Added the ability to update a running VM’s guest memory, applying changes with automatic live migration when applicable.
  * Extended VM memory spec to support an optional max-guest memory value.
  * Enhanced SR-IOV VM provisioning to support guest and max-guest memory scenarios.

* **Bug Fixes**
  * Strengthened SR-IOV memory hot-plug coverage by adding deterministic SR-IOV VM setup, waiting for guest-reported memory updates, validating SR-IOV interface readiness, and confirming TCP connectivity between the hot-plug VM and a reference VM.

### DIFF
--- a/libs/vm/spec.py
+++ b/libs/vm/spec.py
@@ -50,6 +50,7 @@ class CPU:
 @dataclass
 class Memory:
     guest: str
+    maxGuest: str | None = None  # noqa: N815
 
 
 @dataclass
--- a/libs/vm/vm.py
+++ b/libs/vm/vm.py
@@ -18,6 +18,7 @@ from libs.vm.spec import (
     ContainerDisk,
     Devices,
     Disk,
+    Memory,
     Metadata,
     Network,
     SpecDisk,
@@ -128,6 +129,22 @@ class BaseVirtualMachine(VirtualMachine):
         self._spec.template.spec.networks = networks
         serialized = [asdict(obj=net, dict_factory=self._filter_out_none_values) for net in networks]
         ResourceEditor(patches={self: {"spec": {"template": {"spec": {"networks": serialized}}}}}).update()
+
+    def set_guest_memory(self, memory_guest: str) -> None:
+        """Set the guest memory and update the VM spec on the cluster.
+
+        On a running VM this triggers an automatic live migration.
+
+        Args:
+            memory_guest: New guest memory value (e.g. "5Gi").
+        """
+        if self._spec.template.spec.domain.memory:
+            self._spec.template.spec.domain.memory.guest = memory_guest
+        else:
+            self._spec.template.spec.domain.memory = Memory(guest=memory_guest)
+        ResourceEditor(
+            patches={self: {"spec": {"template": {"spec": {"domain": {"memory": {"guest": memory_guest}}}}}}}
+        ).update()
 
     def set_template_affinity(self, affinity: Affinity | None) -> None:
         """Replace the VM template affinity.
--- a/tests/network/sriov/libsriov.py
+++ b/tests/network/sriov/libsriov.py
@@ -1,7 +1,14 @@
 """SR-IOV library constants and utilities."""
 
+from kubernetes.dynamic import DynamicClient
+
 from libs.net.cluster import ipv4_supported_cluster, ipv6_supported_cluster
 from libs.net.ip import random_ipv4_address, random_ipv6_address
+from libs.vm.factory import base_vmspec, fedora_vm
+from libs.vm.spec import CloudInitNoCloud, Interface, Memory, Multus, Network
+from libs.vm.vm import BaseVirtualMachine, add_volume_disk, cloudinitdisk_storage
+from tests.network.libs import cloudinit
+from tests.network.libs.cloudinit import MatchSelector
 from utilities.constants.networking import SRIOV
 from utilities.infra import get_node_selector_dict
 from utilities.network import compose_cloud_init_data_dict, sriov_network_dict
@@ -9,6 +16,60 @@ from utilities.virt import VirtualMachineForTests, fedora_vm_body
 
 VM_SRIOV_IFACE_NAME = "sriov1"
 MTU_9000 = 9000
+
+
+def base_sriov_vm(
+    namespace: str,
+    name: str,
+    client: DynamicClient,
+    sriov_network_name: str,
+    sriov_mac: str,
+    addresses: list[str],
+    memory_guest: str = "1Gi",
+    memory_max_guest: str | None = None,
+) -> BaseVirtualMachine:
+    """Create a Fedora VM with an SR-IOV secondary interface using BaseVirtualMachine.
+
+    The SR-IOV VF is matched by MAC address and renamed to sriov1 via
+    cloud-init set-name.
+
+    Args:
+        namespace: Namespace to deploy the VM in.
+        name: VM name prefix (a unique suffix is appended automatically).
+        client: Kubernetes dynamic client.
+        sriov_network_name: Name of the SR-IOV NetworkAttachmentDefinition.
+        sriov_mac: MAC address to assign to the SR-IOV interface.
+        addresses: CIDR addresses to assign to the SR-IOV interface
+            (e.g. ["172.16.0.1/24", "fd00::1/64"]).
+        memory_guest: Initial guest memory (e.g. "1Gi"). Defaults to "1Gi".
+        memory_max_guest: Maximum guest memory for hot-plug. None disables hot-plug.
+
+    Returns:
+        BaseVirtualMachine configured with an SR-IOV secondary interface.
+    """
+    spec = base_vmspec()
+    spec.template.spec.domain.memory = Memory(guest=memory_guest, maxGuest=memory_max_guest)
+    spec.template.spec.domain.devices.interfaces = [  # type: ignore[union-attr]
+        Interface(name=sriov_network_name, sriov={}, macAddress=sriov_mac),
+    ]
+    spec.template.spec.networks = [
+        Network(name=sriov_network_name, multus=Multus(networkName=f"{namespace}/{sriov_network_name}")),
+    ]
+
+    ethernets: dict[str, cloudinit.EthernetDevice] = {}
+    ethernets["sriov"] = cloudinit.EthernetDevice(
+        addresses=addresses,
+        match=MatchSelector(macaddress=sriov_mac),
+        set_name=VM_SRIOV_IFACE_NAME,
+    )
+    disk, volume = cloudinitdisk_storage(
+        data=CloudInitNoCloud(
+            networkData=cloudinit.asyaml(no_cloud=cloudinit.NetworkData(ethernets=ethernets)),
+            userData=cloudinit.format_cloud_config(userdata=cloudinit.UserData(users=[])),
+        )
+    )
+    spec.template.spec = add_volume_disk(vmi_spec=spec.template.spec, volume=volume, disk=disk)
+    return fedora_vm(namespace=namespace, name=name, client=client, spec=spec)
 
 
 def vm_sriov_mac(mac_suffix_index):
--- a/tests/network/sriov/memory_hotplug/conftest.py
+++ b/tests/network/sriov/memory_hotplug/conftest.py
@@ -1,0 +1,78 @@
+import ipaddress
+from collections.abc import Generator
+from typing import TYPE_CHECKING, Final
+
+import pytest
+
+from libs.net.ip import random_cidr_addresses_by_family
+from libs.vm.oper import run_vm
+from tests.network.sriov.libsriov import base_sriov_vm, vm_sriov_mac
+
+if TYPE_CHECKING:
+    from kubernetes.dynamic import DynamicClient
+    from ocp_resources.namespace import Namespace
+    from ocp_resources.sriov_network import SriovNetwork
+
+    from libs.vm.vm import BaseVirtualMachine
+
+_NET_SEED: Final[int] = 0
+_HOTPLUG_VM_HOST_ADDRESS: Final[int] = 10
+_REF_VM_HOST_ADDRESS: Final[int] = 11
+_INITIAL_MEMORY: Final[str] = "1Gi"
+_MAX_GUEST_MEMORY: Final[str] = "4Gi"
+_SRIOV_CLOUD_INIT_PROFILE: Final[str] = "sriov"
+
+
+@pytest.fixture(scope="class")
+def sriov_vm_with_max_guest(
+    index_number: Generator[int],
+    unprivileged_client: DynamicClient,
+    namespace: Namespace,
+    sriov_network: SriovNetwork,
+) -> Generator[BaseVirtualMachine]:
+    sriov_iface_addresses = random_cidr_addresses_by_family(net_seed=_NET_SEED, host_address=_HOTPLUG_VM_HOST_ADDRESS)
+    with base_sriov_vm(
+        namespace=namespace.name,
+        name="sriov-memory-hotplug-vm",
+        client=unprivileged_client,
+        sriov_network_name=sriov_network.name,
+        sriov_mac=vm_sriov_mac(mac_suffix_index=next(index_number)),
+        addresses=sriov_iface_addresses,
+        memory_guest=_INITIAL_MEMORY,
+        memory_max_guest=_MAX_GUEST_MEMORY,
+    ) as vm:
+        addresses = vm.cloud_init_network_data.ethernets[_SRIOV_CLOUD_INIT_PROFILE].addresses or []
+        run_vm(
+            vm=vm,
+            ip_addresses_by_spec_net_name={
+                sriov_network.name: [str(ipaddress.ip_interface(addr).ip) for addr in addresses]
+            },
+        )
+        yield vm
+
+
+@pytest.fixture(scope="class")
+def sriov_ref_vm(
+    index_number: Generator[int],
+    unprivileged_client: DynamicClient,
+    namespace: Namespace,
+    sriov_network: SriovNetwork,
+) -> Generator[BaseVirtualMachine]:
+    sriov_iface_addresses = random_cidr_addresses_by_family(net_seed=_NET_SEED, host_address=_REF_VM_HOST_ADDRESS)
+    with base_sriov_vm(
+        namespace=namespace.name,
+        name="sriov-ref-vm",
+        client=unprivileged_client,
+        sriov_network_name=sriov_network.name,
+        sriov_mac=vm_sriov_mac(mac_suffix_index=next(index_number)),
+        addresses=sriov_iface_addresses,
+        memory_guest=_INITIAL_MEMORY,
+    ) as vm:
+        addresses = vm.cloud_init_network_data.ethernets[_SRIOV_CLOUD_INIT_PROFILE].addresses or []
+        run_vm(
+            vm=vm,
+            ip_addresses_by_spec_net_name={
+                sriov_network.name: [str(ipaddress.ip_interface(addr).ip) for addr in addresses]
+            },
+        )
+        yield vm
--- a/tests/network/sriov/memory_hotplug/lib_helpers.py
+++ b/tests/network/sriov/memory_hotplug/lib_helpers.py
@@ -1,0 +1,27 @@
+from kubernetes.utils.quantity import parse_quantity
+from timeout_sampler import retry
+
+from libs.vm.vm import BaseVirtualMachine
+from utilities.constants.timeouts import TIMEOUT_5MIN, TIMEOUT_5SEC
+
+
+def hotplug_memory_and_wait(vm: BaseVirtualMachine, memory_guest: str) -> None:
+    """Hot-plug memory on a VM and wait for the guest OS to report the new amount.
+
+    Sets the new guest memory value on the VM, then polls the VMI status
+    guestCurrent field until it reaches the expected value — confirming the
+    live migration triggered by the hot-plug has completed and the guest OS
+    has received the memory.
+
+    Args:
+        vm: The virtual machine to hot-plug memory on.
+        memory_guest: New guest memory value (e.g. "2Gi").
+    """
+    vm.set_guest_memory(memory_guest=memory_guest)
+    _wait_for_guest_memory_in_vmi_status(vm=vm, memory_guest=memory_guest)
+
+
+@retry(wait_timeout=TIMEOUT_5MIN, sleep=TIMEOUT_5SEC)
+def _wait_for_guest_memory_in_vmi_status(vm: BaseVirtualMachine, memory_guest: str) -> bool:
+    current = vm.vmi.instance.status.memory.guestCurrent
+    return bool(current) and parse_quantity(str(current)) == parse_quantity(memory_guest)
--- a/tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
+++ b/tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
@@ -4,10 +4,21 @@ SR-IOV VM memory hot-plug tests.
 Jira: https://redhat.atlassian.net/browse/CNV-69778 # <skip-jira-utils-check>
 """
 
+from typing import Final
+
 import pytest
+
+from libs.net.ip import filter_link_local_addresses
+from libs.net.vmspec import lookup_iface_status
+from tests.network.libs.connectivity import poll_tcp_connectivity
+from tests.network.sriov.memory_hotplug.lib_helpers import hotplug_memory_and_wait
+from utilities.constants.timeouts import TIMEOUT_2MIN
+
+HOTPLUG_MEMORY_SIZE: Final[str] = "3Gi"
 
 
 @pytest.mark.sriov
+@pytest.mark.usefixtures("sriov_ref_vm")
 @pytest.mark.incremental
 class TestSriovMemoryHotplug:
     """
@@ -21,7 +32,11 @@ class TestSriovMemoryHotplug:
     """
 
     @pytest.mark.polarion("CNV-16278")
-    def test_vmi_status(self):
+    def test_vmi_status(
+        self,
+        sriov_vm_with_max_guest,
+        sriov_network,
+    ):
         """
         Test that after memory hot-plug on SR-IOV VM, the VMI status reflects the
         new memory and the SR-IOV interface is present in the VMI status.
@@ -37,11 +52,24 @@ class TestSriovMemoryHotplug:
             - VMI status of the under-test VM shows actual guest memory
               increased to 3Gi
             - SR-IOV interface is present in the VMI status of the under-test
-              VM with guest-agent as an info source
+              VM with guest-agent and domain as an info source
         """
+        hotplug_memory_and_wait(vm=sriov_vm_with_max_guest, memory_guest=HOTPLUG_MEMORY_SIZE)
+        lookup_iface_status(
+            vm=sriov_vm_with_max_guest,
+            iface_name=sriov_network.name,
+            timeout=TIMEOUT_2MIN,
+            predicate=lambda iface: all(source in iface["infoSource"] for source in ("domain", "guest-agent")),
+        )
 
     @pytest.mark.polarion("CNV-16279")
-    def test_connectivity(self):
+    def test_connectivity(
+        self,
+        subtests,
+        sriov_vm_with_max_guest,
+        sriov_ref_vm,
+        sriov_network,
+    ):
         """
         Test that SR-IOV connectivity is preserved after memory hot-plug.
 
@@ -57,6 +85,14 @@ class TestSriovMemoryHotplug:
             - TCP connectivity between the under-test VM and the reference VM
               over the SR-IOV interface is eventually preserved after memory hotplug
         """
-
-
-__test__ = False
+        ref_iface = lookup_iface_status(vm=sriov_ref_vm, iface_name=sriov_network.name)
+        under_test_iface = lookup_iface_status(vm=sriov_vm_with_max_guest, iface_name=sriov_network.name)
+        for server_ip in filter_link_local_addresses(ip_addresses=ref_iface.ipAddresses):
+            with subtests.test(msg=f"IPv{server_ip.version}: {server_ip}"):
+                poll_tcp_connectivity(
+                    client_vm=sriov_vm_with_max_guest,
+                    server_vm=sriov_ref_vm,
+                    server_ip=str(server_ip),
+                    client_bind_dev=under_test_iface.interfaceName,
+                    server_bind_dev=ref_iface.interfaceName,
+                )


azhivovk:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/c80a6d26120be2786f09641375fd4c0d8c38ba64..46a3996239852c9b588a17c4b2d84feb1d149f71): drop unused affinity param

tests/network/sriov/memory_hotplug/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+        name="sriov-memory-hotplug-vm",
+        client=unprivileged_client,
+        sriov_network_name=sriov_network.name,
+        sriov_mac=vm_sriov_mac(mac_suffix_index=next(index_number)),
+        net_seed=_NET_SEED,
+        host_address=_HOTPLUG_VM_HOST_ADDRESS,
+        memory_guest=_INITIAL_MEMORY,
+        memory_max_guest=_MAX_GUEST_MEMORY,
+    ) as vm:
+        yield vm
```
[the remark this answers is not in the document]
azhivovk: This flow is according to agreed STD, which follows the existing pattern of test flow
We shouldn't do 2 memory hot-plugs, 1 is enough for the current tests
The setup is shared between the 2 tests

tests/network/sriov/memory_hotplug/lib_helpers.py
```diff
@@ -0,0 +1,27 @@
...
+        memory_guest: New guest memory value (e.g. "2Gi").
+    """
+    vm.set_guest_memory(memory_guest=memory_guest)
+    _wait_for_guest_memory_in_vmi_status(vm=vm, memory_guest=memory_guest)
+
+
+@retry(wait_timeout=TIMEOUT_10MIN, sleep=TIMEOUT_5SEC, exceptions_dict={})
+def _wait_for_guest_memory_in_vmi_status(vm: BaseVirtualMachine, memory_guest: str) -> bool:
+    current = vm.vmi.instance.status.memory.guestCurrent
+    return bool(current) and parse_quantity(str(current)) >= parse_quantity(memory_guest)
```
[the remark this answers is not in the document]
azhivovk: `exceptions_dict={}` empty dict does not mean it expects all types of exceptions?

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
```diff
@@ -0,0 +1,96 @@
...
+            - A running reference VM with an SR-IOV interface on the same
+              SR-IOV network
+
+        Steps:
+            1. Poll TCP connection from the under-test VM to the reference VM over the SR-IOV interface
+
+        Expected:
+            - TCP connectivity between the under-test VM and the reference VM
+              over the SR-IOV interface is eventually preserved after memory hotplug
+        """
```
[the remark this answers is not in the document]
azhivovk: You're repeating youself https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#discussion_r3461063981
We also have incremental mark - if hp fails we won't test connectivity

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
```diff
@@ -0,0 +1,96 @@
...
+        iface = lookup_iface_status(vm=sriov_ref_vm, iface_name=sriov_network.name)
+        for server_ip in filter_link_local_addresses(ip_addresses=iface.ipAddresses):
+            with subtests.test(msg=f"IPv{server_ip.version}: {server_ip}"):
+                poll_tcp_connectivity(
+                    client_vm=sriov_vm_with_memory_hotplug,
+                    server_vm=sriov_ref_vm,
+                    server_ip=str(server_ip),
+                    client_bind_dev=VM_SRIOV_GUEST_DEV,
+                    server_bind_dev=VM_SRIOV_GUEST_DEV,
+                )
```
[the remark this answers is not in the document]
azhivovk: We'll have addresses in the var 
If not, we'll fail in setup in lookup

azhivovk:
@coderabbitai I addressed your review comments, please revisit

libs/vm/vm.py
```diff
@@ -129,6 +130,24 @@ def set_networks(self, networks: list[Network]) -> None:
...
+        Args:
+            memory_guest: New guest memory value (e.g. "5Gi").
+        """
+        if self._spec.template.spec.domain.memory:
+            self._spec.template.spec.domain.memory.guest = memory_guest
+        else:
+            self._spec.template.spec.domain.memory = Memory(guest=memory_guest)
+        ResourceEditor(
+            patches={self: {"spec": {"template": {"spec": {"domain": {"memory": {"guest": memory_guest}}}}}}}
+        ).update()
```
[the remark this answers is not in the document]
azhivovk: This is the convention in class
I'll keep it as is

azhivovk:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/46a3996239852c9b588a17c4b2d84feb1d149f71..d9225175d224ea5e034b80f77194d05f3df43cae): apply bot's suggestions

azhivovk:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/d9225175d224ea5e034b80f77194d05f3df43cae..ef7da7a6aa9edad06baa007e74a6bc2d5dc2eb47):
- Address bot's [suggestion](https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#discussion_r3466165611) 
- Drop `>` and expect the exact memory size we've defined

tests/network/sriov/memory_hotplug/lib_helpers.py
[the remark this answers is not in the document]
azhivovk: I dropped the exceptions_dict param to fetch all types of exceptions

libs/vm/spec.py
```diff
@@ -48,6 +48,7 @@ class CPU:
 @dataclass
 class Memory:
     guest: str
+    maxGuest: str | None = None  # noqa: N815
```
nirdothan, line 53: The commit message describes the code instead of why this test is needed.
The reason that it is needed is because of a bug in the memory calculation that we used to have. 
In short: virt-handler miscalculated guest memory in mem hotplug and libvirt wanted more. In turn it tried to lock additional memory pages, but since it's running in an unprivileged pod, it failed to do so.
The symptom was that during migration (in SRIOV we hot-unplug in the source and then hotplug in the target) hotplug failed. For the user it manifested into loss of network connectivity.


[CNV-69778](https://redhat.atlassian.net/browse/CNV-69778) — After workload migration the SR-IOV interface failed to attach (ON_QA)
[CNV-73148](https://redhat.atlassian.net/browse/CNV-73148) — SRIOV NIC reattachment fails after memory HP migration [v4.20] (Closed)

EDIT: And so we need to verify that we can perform memory hotplug in SRIOV without connectivity loss (and also verify the memory is increased as the user expects).

libs/vm/vm.py
```diff
@@ -129,6 +130,24 @@ def set_networks(self, networks: list[Network]) -> None:
         serialized = [asdict(obj=net, dict_factory=self._filter_out_none_values) for net in networks]
         ResourceEditor(patches={self: {"spec": {"template": {"spec": {"networks": serialized}}}}}).update()
 
+    def set_guest_memory(self, memory_guest: str) -> None:
+        """Update the guest memory on a running VM.
+
+        Updates the in-memory spec first so the object stays consistent with
+        the cluster without requiring a re-fetch after the patch. On a running
+        VM this triggers an automatic live migration.
```
nirdothan: > in-memory spec first so the object stays consistent with
        the cluster without requiring a re-fetch after the patch.

Isn't this too technical for consumers? They're just trying to change the VM guest memory. Please consider simplifying the text. 
The migration comment is good though.

tests/network/sriov/memory_hotplug/conftest.py
```diff
@@ -0,0 +1,83 @@
...
+_NET_SEED: Final[int] = 0
+_HOTPLUG_VM_HOST_ADDRESS: Final[int] = 10
+_REF_VM_HOST_ADDRESS: Final[int] = 11
+_INITIAL_MEMORY: Final[str] = "1Gi"
+_MAX_GUEST_MEMORY: Final[str] = "4Gi"
+_SRIOV_CLOUD_INIT_PROFILE: Final[str] = "sriov"
+
+
+@pytest.fixture(scope="class")
+def sriov_vm_for_memory_hotplug(
```
nirdothan: Do we really care what this VM is for?
From the code below I see that all it does is add an IP to `base_sriov_vm`, hence please consider something like  `sriov_vm_with_ip`.

tests/network/sriov/memory_hotplug/conftest.py
```diff
@@ -0,0 +1,83 @@
...
+        sriov_mac=vm_sriov_mac(mac_suffix_index=next(index_number)),
+        addresses=sriov_iface_addresses,
+        memory_guest=_INITIAL_MEMORY,
+        memory_max_guest=_MAX_GUEST_MEMORY,
+    ) as vm:
+        yield vm
+
+
+@pytest.fixture(scope="class")
+def sriov_ref_vm(
```
nirdothan, line 55: If the only difference between this fixture and the previous are the memory parameters passed to `base_sriov_vm` then we don't need 2 fixtures, we can have one fixture with parametrized memory values (with defaults).

tests/network/sriov/memory_hotplug/lib_helpers.py
```diff
@@ -0,0 +1,27 @@
...
+        memory_guest: New guest memory value (e.g. "2Gi").
+    """
+    vm.set_guest_memory(memory_guest=memory_guest)
+    _wait_for_guest_memory_in_vmi_status(vm=vm, memory_guest=memory_guest)
+
+
+@retry(wait_timeout=TIMEOUT_10MIN, sleep=TIMEOUT_5SEC)
+def _wait_for_guest_memory_in_vmi_status(vm: BaseVirtualMachine, memory_guest: str) -> bool:
+    current = vm.vmi.instance.status.memory.guestCurrent
+    return bool(current) and parse_quantity(str(current)) == parse_quantity(memory_guest)
```
nirdothan, line 27: Just verifying `bool(current)` will be false if current differs from `None`?

tests/network/sriov/memory_hotplug/lib_helpers.py
```diff
@@ -0,0 +1,27 @@
+from kubernetes.utils.quantity import parse_quantity
+from timeout_sampler import retry
+
+from libs.vm.vm import BaseVirtualMachine
+from utilities.constants import TIMEOUT_5SEC, TIMEOUT_10MIN
+
+
+def hotplug_memory_and_wait(vm: BaseVirtualMachine, memory_guest: str) -> None:
```
nirdothan, line 8: At first I thought that this should be a VM method but since it **must** be used on a running VM perhaps it shouldn't useless we already have other "hot" methods there. 
Perhaps we can split this function into 2:
One a VM method that sets memory, and a separate one in this test that just waits for guest momery.

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
```diff
@@ -39,9 +53,21 @@ def test_vmi_status(self):
             - SR-IOV interface is present in the VMI status of the under-test
               VM with guest-agent as an info source
         """
+        hotplug_memory_and_wait(vm=sriov_vm_for_memory_hotplug, memory_guest=_HOTPLUG_MEMORY)
+        lookup_iface_status(
+            vm=sriov_vm_for_memory_hotplug,
+            iface_name=sriov_network.name,
+            predicate=lambda iface: "guest-agent" in iface["infoSource"],
```
nirdothan: We need to validate infoSource `domain` because it indicates that the hot plug took place. See upstream [1].

[1] https://github.com/kubevirt/kubevirt/blob/a3a194cc10b919052fd20105122ab09cfc9ed0ed/tests/network/sriov.go#L403

tests/network/sriov/libsriov.py
```diff
@@ -11,6 +18,66 @@
...
+        memory_guest: Initial guest memory (e.g. "1Gi"). Defaults to "1Gi".
+        memory_max_guest: Maximum guest memory for hot-plug. None disables hot-plug.
+
+    Returns:
+        BaseVirtualMachine configured with an SR-IOV secondary interface.
+    """
+    spec = base_vmspec()
+    spec.template.spec.domain.memory = Memory(guest=memory_guest, maxGuest=memory_max_guest)
+    spec.template.spec.domain.devices.interfaces = [  # type: ignore[union-attr]
+        Interface(name="default", masquerade={}),
```
nirdothan: Why do we need the masquerade interface here? We're only checking SR-IOV and another interface can accidentally lead to false positive in connectivity test.

azhivovk:
Change: reabse upon main

libs/vm/vm.py
azhivovk: It's a similar docstring to the existing `set_networks`

tests/network/sriov/memory_hotplug/conftest.py
azhivovk: I remember that we agreed we won't parametrize VM fixtures so each used VM fixture has a clear and readable config
@servolkov WDYT?

tests/network/sriov/memory_hotplug/conftest.py
azhivovk: Renamed to `sriov_vm_with_max_guest` to better describe the VM config

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
azhivovk: Done
Also added to the STD docstring

tests/network/sriov/memory_hotplug/lib_helpers.py
azhivovk: True when current differs from None (i.e. the field is set), and False when current is None (field not yet available)

tests/network/sriov/memory_hotplug/lib_helpers.py
azhivovk: I followed the similar pattern from `update_nad_references` helper
https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/libs/net/vmspec.py#L309
We change VM spec and then we wait for it to be applied

azhivovk:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/51445dcd8552cdc464eb90a48a0b6391214299c3..7b7c8b29cb1338c87b2fcf31dab8d6e1a3741ed9): apply changes after CR
- Rename VM fixture
- Add domain to predicate post HP
- Edit commit message

libs/vm/spec.py
azhivovk: Edited the commit message, thanks

openshift-virtualization-qe-bot-6:
New container for quay.io/openshift-cnv/openshift-virtualization-tests:pr-5349 published

tests/network/sriov/libsriov.py
azhivovk: Removed, thanks

azhivovk:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/7b7c8b29cb1338c87b2fcf31dab8d6e1a3741ed9..dd76704cf62e63df1a033312a45908719e1e18f4): 
- Block new tests due to a new bug
- Remove masquerade interface

azhivovk:
Change: edit commit message

openshift-virtualization-qe-bot-6:
**Clean rebase detected** — no code changes compared to previous head (`dd76704`).

azhivovk:
Change: edit commit message

openshift-virtualization-qe-bot-6:
**Clean rebase detected** — no code changes compared to previous head (`4a3e5d8`).
The following labels were preserved: `commented-coderabbitai[bot]`.

openshift-virtualization-qe-bot-3:
/retest all

_Auto-triggered: Files in this PR were modified by merged PR #5285._

<details>
<summary>Overlapping files</summary>

libs/vm/vm.py

</details>

openshift-virtualization-qe-bot-3:
/retest all

_Auto-triggered: Files in this PR were modified by merged PR #5321._

<details>
<summary>Overlapping files</summary>

tests/network/sriov/libsriov.py

</details>

azhivovk:
Change: rebase and resolve conflicts

libs/vm/vm.py
nirdothan: I don't insist but saying that it's similar to somewhere else doesn't make it right.

azhivovk: Usually when adding new code to T2, we have existing new tests for reference, so we can rely on similar existing patterns
Another reason to keep it IMO is for users to be are aware of the method usage

tests/network/sriov/memory_hotplug/lib_helpers.py
nirdothan: Sorry that doesn't explain why it's better and I also don't find it similar. Memory is a direct property of the VM. NAD is more indented and is a property of one of the netwoks if secondary even exists.

azhivovk: I followed the same logic of editing spec and then wait for it to be applied
IMO if we only change some field in a VM and don't wait for it to apply, we can silently move to the next test step and not fail early when possible
I already implemented  your [request](https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#discussion_r3473874217): 
- `vm.set_guest_memory()` is setting the requested memory
- `_wait_for_guest_memory_in_vmi_status` waits for the new memory to be applied
- These steps are wrapped in a helper `hotplug_memory_and_wait` which is used in the test for clear and readable test step

Please suggest the change that you mean if I understood incorrectly

tests/network/sriov/memory_hotplug/lib_helpers.py
```diff
@@ -0,0 +1,27 @@
...
+    Args:
+        vm: The virtual machine to hot-plug memory on.
+        memory_guest: New guest memory value (e.g. "2Gi").
+    """
+    vm.set_guest_memory(memory_guest=memory_guest)
+    _wait_for_guest_memory_in_vmi_status(vm=vm, memory_guest=memory_guest)
+
+
+@retry(wait_timeout=TIMEOUT_10MIN, sleep=TIMEOUT_5SEC)
+def _wait_for_guest_memory_in_vmi_status(vm: BaseVirtualMachine, memory_guest: str) -> bool:
```
vsibirsk, line 25: nit: if something goes wrong with hotplug, you're stuck here for 10 min.
you can update it to check the relevant VMIMigration object and quick exit if for e.g. it's stuck (most common unschedulable)

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
```diff
@@ -37,11 +52,23 @@ def test_vmi_status(self):
             - VMI status of the under-test VM shows actual guest memory
               increased to 3Gi
             - SR-IOV interface is present in the VMI status of the under-test
-              VM with guest-agent as an info source
+              VM with guest-agent and domain as an info source
         """
+        hotplug_memory_and_wait(vm=sriov_vm_with_max_guest, memory_guest=_HOTPLUG_MEMORY)
```
vsibirsk: should be part of setup i believe

libs/vm/vm.py
servolkov: I agree with @nirdothan in this thread.

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
```diff
@@ -4,10 +4,21 @@
...
+from typing import Final
+
 import pytest
 
+from libs.net.ip import filter_link_local_addresses
+from libs.net.vmspec import lookup_iface_status
+from tests.network.libs.connectivity import poll_tcp_connectivity
+from tests.network.sriov.memory_hotplug.lib_helpers import hotplug_memory_and_wait
+
+_HOTPLUG_MEMORY: Final[str] = "3Gi"
```
servolkov: nit
`_HOTPLUG_MEMORY` -> `_HOTPLUG_MEMORY_SIZE`

tests/network/sriov/memory_hotplug/conftest.py
```diff
@@ -0,0 +1,83 @@
...
+        sriov_network_name=sriov_network.name,
+        sriov_mac=vm_sriov_mac(mac_suffix_index=next(index_number)),
+        addresses=sriov_iface_addresses,
+        memory_guest=_INITIAL_MEMORY,
+    ) as vm:
+        yield vm
+
+
+@pytest.fixture(scope="class")
+def running_sriov_vms(
```
servolkov: I was thinking about this pattern a bit more: `running_sriov_vms` changes the state of VMs created by `sriov_vm_with_max_guest` and `sriov_ref_vm`, then tests request those same fixtures directly.  A reader looking at `sriov_vm_with_max_guest` has no idea whether it yields a running VM or a stopped one. You have to trace the `usefixtures("running_sriov_vms")` dependency to understand the actual state. I find it very messy. 

Fixtures should yield resources in their final usable state - not depend on a separate fixture to mutate them into readiness.

tests/network/sriov/memory_hotplug/conftest.py
servolkov: @azhivovk can't we re-use already available fixture like `sriov_vm1` as a reference VM? 

regarding memory parametrization: I don't clearly understand what you are talking about, we need 2 VMs as a precondition, parametrization does not help in this case.

servolkov (changes requested):

openshift-virtualization-qe-bot-3:
/retest all

_Auto-triggered: Files in this PR were modified by merged PR #5440._

<details>
<summary>Overlapping files</summary>

libs/vm/vm.py

</details>

tests/network/sriov/memory_hotplug/conftest.py
azhivovk: I can reuse `sriov_vm1` but I thought we are gradually moving to use `BaseVirtualMachine`
`sriov_vm1` is instance of `VirtualMachineForTests`
In current implementation we use `poll_tcp_connectivity` which uses `vm.console()` which `VirtualMachineForTests` doesn't have, and probably more adjustments are necessary


IIUC @nirdothan meant something like:

```
  def _sriov_vm(
      param: dict,
      index_number: Generator[int],
      unprivileged_client: DynamicClient,
      namespace: Namespace,
      sriov_network: SriovNetwork,
  ) -> Generator[BaseVirtualMachine]:
      with base_sriov_vm(
          namespace=namespace.name,
          name=param["name"],
          client=unprivileged_client,
          sriov_network_name=sriov_network.name,
          sriov_mac=vm_sriov_mac(mac_suffix_index=next(index_number)),
          addresses=random_cidr_addresses_by_family(net_seed=_NET_SEED, host_address=param["host_address"]),
          memory_guest=_INITIAL_MEMORY,
          memory_max_guest=param.get("memory_max_guest"),
      ) as vm:
          yield vm

  
  def sriov_vm_with_max_guest(index_number, unprivileged_client, namespace, sriov_network):
      yield from _sriov_vm(
          param={"name": "sriov-memory-hotplug-vm", "host_address": _HOTPLUG_VM_HOST_ADDRESS, "memory_max_guest": _MAX_GUEST_MEMORY},
          index_number=index_number,
          unprivileged_client=unprivileged_client,
          namespace=namespace,
          sriov_network=sriov_network,
      )


  def sriov_ref_vm(index_number, unprivileged_client, namespace, sriov_network):
      yield from _sriov_vm(
          param={"name": "sriov-ref-vm", "host_address": _REF_VM_HOST_ADDRESS},
          index_number=index_number,
          unprivileged_client=unprivileged_client,
          namespace=namespace,
          sriov_network=sriov_network,
      )
```

IMO:
It's too complex and unreadable and it's better to have 2 dedicated fixtures for each VM, even if the code seems duplicate

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
servolkov: per STD:
```
 Steps:
     1. Hot-plug memory to 3Gi on the under-test VM
```

IMO the test is clear: we hot plug memory and check the result of it.

tests/network/sriov/memory_hotplug/lib_helpers.py
azhivovk: We agreed on the STD that after we increase to the requested memory, we wait for it to appear

10-minute timeout is really a lot of time - I lowered it to 5, but OTOH we have 12-minute timeout when we wait for migration to finish like in `migrate_vm_and_verify`

When we perform interface hot-plug, it also triggers a live-migration and we don't monitor the VMIM status, but just wait for the interface to appear, like [here](https://github.com/RedHatQE/openshift-virtualization-tests/blob/main/tests/network/l2_bridge/libl2bridge.py#L62) 

I lowered the timeout to 5 min to reduce the wait on failure, keeping the implementation consistent with the STD

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
azhivovk: We agreed on the STD the memory HP is part of the test steps
I'll keep it as is, thanks

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
azhivovk: Done

tests/network/sriov/memory_hotplug/conftest.py
azhivovk: Moved the start and wait for agent to the dedicated VM fixtures

libs/vm/vm.py
azhivovk: I [disagree](https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#discussion_r3505240491)
Rephrased

tests/network/sriov/memory_hotplug/lib_helpers.py
azhivovk: Can I resolve?

azhivovk:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/d01af7b0e5c7acc97d769c14d8c418831e335d6a..8cc1c2ceffe704eb25f348fe9fb1e0e703f241f3):
- Rephrase `set_guest_memory` docstring
- Start each VM individually `sriov_ref_vm` `sriov_vm_with_max_guest`
- Rename constant `HOTPLUG_MEMORY_SIZE`
- Edit constants module import
- Add 2 minute timeout to the interface lookup post mem HP (30-second default may not be enough)
- Drop Jira blocker (tests pass with disabled network resources injector, and it's affecting SR-IOV and secondary network tests)

vsibirsk (dismissed):

tests/network/sriov/memory_hotplug/conftest.py
nirdothan: I prefer this ^ style (though not with a loosely typed param object), but I'm leaving it up to you and @servolkov to decide.

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
```diff
@@ -57,6 +85,14 @@ def test_connectivity(self):
...
               over the SR-IOV interface is eventually preserved after memory hotplug
         """
-
-
-__test__ = False
+        ref_iface = lookup_iface_status(vm=sriov_ref_vm, iface_name=sriov_network.name)
+        under_test_iface = lookup_iface_status(vm=sriov_vm_with_max_guest, iface_name=sriov_network.name)
+        for server_ip in filter_link_local_addresses(ip_addresses=ref_iface.ipAddresses):
+            with subtests.test(msg=f"IPv{server_ip.version}: {server_ip}"):
+                poll_tcp_connectivity(
```
servolkov, line 92: stop, `poll_tcp_connectivity` returns just bool, if there is no connectivity the test silently passes?

tests/network/sriov/memory_hotplug/conftest.py
servolkov: I am ok with what we have now

tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py
azhivovk: It won't silently pass
If we don't get the expected result of the client/server iperf process, `poll_tcp_connectivity` will raise an error (on setup and on the test)

Example of the helper when we fail the expected connectivity:
```
tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py::TestSriovMemoryHotplug::test_connectivity FAILED
_______________________________________________________________________________ TestSriovMemoryHotplug.test_connectivity ________________________________________________________________________________
contains 2 failed subtests
2026-07-06T16:26:39.531320+00:00 ocp_resources VirtualMachine INFO Delete VirtualMachine sriov-memory-hotplug-vm-21fc7d61dbc44734
Traceback (most recent call last):
  File "/openshift-virtualization-tests/tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py", line 92, in test_connectivity
    poll_tcp_connectivity(
    ~~~~~~~~~~~~~~~~~~~~~^
        client_vm=sriov_vm_with_max_guest,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        expect_connectivity=False
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/openshift-virtualization-tests/.venv/lib64/python3.14/site-packages/timeout_sampler/__init__.py", line 280, in wrapper
    for sample in TimeoutSampler(
                  ~~~~~~~~~~~~~~^
        func=func,
        ^^^^^^^^^^
    ...<7 lines>...
        **kwargs,
        ^^^^^^^^^
    ):
    ^
  File "/openshift-virtualization-tests/.venv/lib64/python3.14/site-packages/timeout_sampler/__init__.py", line 182, in __iter__
    raise TimeoutExpiredError(self._get_exception_log(exp=last_exp), last_exp=last_exp)
timeout_sampler.TimeoutExpiredError: Timed Out: 60
Function: tests.network.libs.connectivity.poll_tcp_connectivity  Kwargs: {'client_vm': <libs.vm.vm.BaseVirtualMachine object at 0x7f01907e8a50>, 'server_vm': <libs.vm.vm.BaseVirtualMachine object at 0x7f01908f7a10>, 'server_ip': 'fd00:1234:5678::577:b', 'client_bind_dev': 'eth0', 'server_bind_dev': 'sriov1', 'expect_connectivity': False}
Last exception: N/A.

TEST: TestSriovMemoryHotplug.test_connectivity STATUS: FAILED
```


`poll_tcp_connectivity` is working as expected - if we don't reach the expected connectivity/no connectivity, the helper will raise `TimeoutExpiredError`

servolkov: right, I overlooked `retry` decorator, resolve pls

servolkov (approved):

azhivovk:
/verified
Dual-stack BM with cnv-4.23 and disabled network-resources-injector:
`uv run pytest --verbose -rs -s -o log_cli=true --leftovers-collector --tb=native --skip-deprecated-api-test --cluster-sanity-skip-storage-check tests/network/sriov/memory_hotplug/test_sriov_memory_hotplug.py -m sriov`
Result: `2 passed, 1 deselected, 12 warnings, 2 subtests passed in 386.15s (0:06:26)`

This PR can be merged since the fix for `CNV-91360` (which affects many secondary network tests) will land downstream in the next few days

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  Do NOT submit a blocking review event (REQUEST_CHANGES/APPROVE).
  Post a single inline PR comment on Files Changed (non-blocking COMMENT flow).

  **Analysis Requirements:**

         never imports B directly.

  **Output rules:**

  **Your deliverable:**
  Your inline informational comment will be based on the following requirements:

  **Test Execution Plan**

  Use these formats:

  **Real test commands (MANDATORY when changes affect session/runtime code):**

  When the affected code runs at session/collection time (conftest fixtures, pytest plugins,
  config hooks, session-scoped setup) or modifies runtime behavior that unit tests mock away,
  you MUST include concrete `pytest` commands the PR author must run on a real cluster
  to verify the change works end-to-end. Include:
  - A command for the **error/fix path** (the scenario the PR fixes)
  - A command for the **happy path** (regression: the normal case still works)
  - Use lightweight tests (e.g., `--collect-only` for startup failures,
    a single small test for runtime behavior)
  If the PR only changes test logic (not utilities/libs/conftest), the affected test
  paths themselves serve as the real test commands — no separate section needed.

  Example output for a session-startup fix:
  ```
  **Real tests (cluster required)**
  Error path (the fix):
  `pytest tests/storage/.../test_foo.py --storage-class-matrix=nonexistent-sc --collect-only`
  Expected: ValueError with clear message, not IndexError

  Happy path (regression):
  `pytest tests/storage/.../test_foo.py --storage-class-matrix=<valid-sc> -k test_bar`
  Expected: session starts normally
  ```

  **Guidelines:**

  1. **YOUR ONLY DELIVERABLE**: Post one non-blocking inline comment containing the test execution plan on the first changed line
  3. **FALLBACK ONLY**: If inline comment API calls fail after retrying, post as a regular PR comment

</details>

tests/network/sriov/memory_hotplug/lib_helpers.py
```diff
@@ -0,0 +1,27 @@
+from kubernetes.utils.quantity import parse_quantity
+from timeout_sampler import retry
+
+from libs.vm.vm import BaseVirtualMachine
+from utilities.constants.timeouts import TIMEOUT_5MIN, TIMEOUT_5SEC
+
```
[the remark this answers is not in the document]
azhivovk: This is not a must

libs/vm/spec.py
```diff
@@ -49,6 +49,7 @@ class CPU:
...
```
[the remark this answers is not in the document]
azhivovk: Done https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#issuecomment-4895310578

nirdothan (approved):
Thank you!

libs/vm/vm.py
```diff
@@ -129,6 +130,22 @@ def set_networks(self, networks: list[Network]) -> None:
...
+        On a running VM this triggers an automatic live migration.
+
+        Args:
+            memory_guest: New guest memory value (e.g. "5Gi").
+        """
+        if self._spec.template.spec.domain.memory:
+            self._spec.template.spec.domain.memory.guest = memory_guest
+        else:
+            self._spec.template.spec.domain.memory = Memory(guest=memory_guest)
+        ResourceEditor(
```
rnetser, line 145: if the vm is not created, this will fail, right?
a user could call `set_guest_memory` without creating the vm

tests/network/sriov/memory_hotplug/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+    ) as vm:
+        vm.start()
+        vm.wait_for_agent_connected()
+        addresses = vm.cloud_init_network_data.ethernets[_SRIOV_CLOUD_INIT_PROFILE].addresses or []
+        wait_for_ifaces_status(
+            vm=vm,
+            ip_addresses_by_spec_net_name={
+                sriov_network.name: [str(ipaddress.ip_interface(addr).ip) for addr in addresses]
+            },
+        )
```
rnetser: the comment on this pattern of repeated code was added a few times in the past
if these steps are required to be checked when you create a vm, please extract to a helper and re-use

rnetser (changes requested):

tests/network/sriov/memory_hotplug/conftest.py
azhivovk: IMO it's better to keep it as is for clarity about the steps we perform on the VM
We don't always start VMs and wait for agent and IPs to appear, and here we see the explicit steps

libs/vm/vm.py
azhivovk: All methods in this class assume the VM is already created, or else `ResourceEditor.update()` will raise API error
Do I need to update the docstring?

tests/network/sriov/memory_hotplug/conftest.py
servolkov: I thinks a helper makes sense, however the path how we extract addresses may vary per our codebase, but we can handle it. I would create a follow-up ticket (@azhivovk could you create pls) and let this PR be merged.

azhivovk: https://redhat.atlassian.net/browse/CNV-92295

rnetser: @servolkov while in the past, when the team was new, a follow up was accepted. now a follow up is not the way to go.
when new code is introduced, it should be written following the repo's rules and not have follow ups
@azhivovk please fix

azhivovk: Added `start_vm_and_wait_for_sriov_ifaces` since this new code is repetitive
In a follow-up we can do similarly in existing tests, using more generic helper

azhivovk:
[Change after CR](https://github.com/RedHatQE/openshift-virtualization-tests/compare/8cc1c2ceffe704eb25f348fe9fb1e0e703f241f3..8d1319dca267341f52cc87ef844b8eaa4c9e4f95): extract post VM creation repetitive code checks to a helper

tests/network/sriov/libsriov.py
```diff
@@ -11,6 +18,60 @@
...
+    spec.template.spec.networks = [
+        Network(name=sriov_network_name, multus=Multus(networkName=f"{namespace}/{sriov_network_name}")),
+    ]
+
+    ethernets: dict[str, cloudinit.EthernetDevice] = {}
+    ethernets["sriov"] = cloudinit.EthernetDevice(
+        addresses=addresses,
+        match=MatchSelector(macaddress=sriov_mac),
+        set_name=VM_SRIOV_IFACE_NAME,
+    )
```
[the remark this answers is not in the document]
azhivovk: Redundant

nirdothan (approved):

azhivovk:
/verified
No change in code logic
https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#issuecomment-4895310578

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  Do NOT submit a blocking review event (REQUEST_CHANGES/APPROVE).
  Post a single inline PR comment on Files Changed (non-blocking COMMENT flow).

  **Analysis Requirements:**

         never imports B directly.

  **Output rules:**

  **Your deliverable:**
  Your inline informational comment will be based on the following requirements:

  **Test Execution Plan**

  Use these formats:

  **Real test commands (MANDATORY when changes affect session/runtime code):**

  When the affected code runs at session/collection time (conftest fixtures, pytest plugins,
  config hooks, session-scoped setup) or modifies runtime behavior that unit tests mock away,
  you MUST include concrete `pytest` commands the PR author must run on a real cluster
  to verify the change works end-to-end. Include:
  - A command for the **error/fix path** (the scenario the PR fixes)
  - A command for the **happy path** (regression: the normal case still works)
  - Use lightweight tests (e.g., `--collect-only` for startup failures,
    a single small test for runtime behavior)
  If the PR only changes test logic (not utilities/libs/conftest), the affected test
  paths themselves serve as the real test commands — no separate section needed.

  Example output for a session-startup fix:
  ```
  **Real tests (cluster required)**
  Error path (the fix):
  `pytest tests/storage/.../test_foo.py --storage-class-matrix=nonexistent-sc --collect-only`
  Expected: ValueError with clear message, not IndexError

  Happy path (regression):
  `pytest tests/storage/.../test_foo.py --storage-class-matrix=<valid-sc> -k test_bar`
  Expected: session starts normally
  ```

  **Guidelines:**

  1. **YOUR ONLY DELIVERABLE**: Post one non-blocking inline comment containing the test execution plan on the first changed line
  3. **FALLBACK ONLY**: If inline comment API calls fail after retrying, post as a regular PR comment

</details>

libs/vm/spec.py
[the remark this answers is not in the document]
azhivovk: Done https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#issuecomment-4905552319

tests/network/sriov/memory_hotplug/conftest.py
servolkov: > now a follow up is not the way to go.

@rnetser I disagree. Okay, we can call it not a follow-up rather "dedicated PR". It is normal to note that we repeat some pattern and handle it eventually. But refactoring existing tests to use a new helper is out of scope for this PR, and a separate PR is the proper way to introduce a new helper and fix tests. The project rules also say: "Keep PRs focused - each PR addresses ONE topic. Out-of-scope improvements go in a separate PR." 

The current helper does not solve the issue @rnetser raised. [This PR](https://github.com/RedHatQE/openshift-virtualization-tests/pull/5553) is about to help with the issue. Pls, take a look and let me know if it makes sense.

azhivovk: @rnetser can we merge this PR now or do we need to merge [this](https://github.com/RedHatQE/openshift-virtualization-tests/pull/5553) one first? ^

libs/vm/vm.py
rnetser: no, but i think that this approach should be re-visited for all functions. you cannot assume that the vmis running
you can resolve once @servolkov  is also aware of this

tests/network/sriov/memory_hotplug/conftest.py
rnetser: https://github.com/RedHatQE/openshift-virtualization-tests/pull/5553 should be merged first

tests/network/sriov/libsriov.py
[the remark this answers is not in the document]
openshift-virtualization-qe-bot-3: ⚠️ This thread was automatically unresolved because it was resolved without a substantive response. Please address the review comment and explain how it was resolved before resolving this thread again.

azhivovk:
Change: rebase upon main

azhivovk:
[Change](https://github.com/RedHatQE/openshift-virtualization-tests/compare/1a60a14..8186f1c): use `run_vm`

azhivovk:
/verified
Ran [this](https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#issuecomment-4895310578) - all passed

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  Do NOT submit a blocking review event (REQUEST_CHANGES/APPROVE).
  Post a single inline PR comment on Files Changed (non-blocking COMMENT flow).

  **Analysis Requirements:**

         never imports B directly.

  **Output rules:**

  **Your deliverable:**
  Your inline informational comment will be based on the following requirements:

  **Test Execution Plan**

  Use these formats:

  **Real test commands (MANDATORY when changes affect session/runtime code):**

  When the affected code runs at session/collection time (conftest fixtures, pytest plugins,
  config hooks, session-scoped setup) or modifies runtime behavior that unit tests mock away,
  you MUST include concrete `pytest` commands the PR author must run on a real cluster
  to verify the change works end-to-end. Include:
  - A command for the **error/fix path** (the scenario the PR fixes)
  - A command for the **happy path** (regression: the normal case still works)
  - Use lightweight tests (e.g., `--collect-only` for startup failures,
    a single small test for runtime behavior)
  If the PR only changes test logic (not utilities/libs/conftest), the affected test
  paths themselves serve as the real test commands — no separate section needed.

  Example output for a session-startup fix:
  ```
  **Real tests (cluster required)**
  Error path (the fix):
  `pytest tests/storage/.../test_foo.py --storage-class-matrix=nonexistent-sc --collect-only`
  Expected: ValueError with clear message, not IndexError

  Happy path (regression):
  `pytest tests/storage/.../test_foo.py --storage-class-matrix=<valid-sc> -k test_bar`
  Expected: session starts normally
  ```

  **Guidelines:**

  1. **YOUR ONLY DELIVERABLE**: Post one non-blocking inline comment containing the test execution plan on the first changed line
  3. **FALLBACK ONLY**: If inline comment API calls fail after retrying, post as a regular PR comment

</details>

libs/vm/spec.py
```diff
@@ -50,6 +50,7 @@ class CPU:
...
```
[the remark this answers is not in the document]
azhivovk: Done https://github.com/RedHatQE/openshift-virtualization-tests/pull/5349#issuecomment-4924832664

libs/vm/vm.py
servolkov: ack. However, `ResourceEditor.update()` will fail with an API error if the VM doesn't exist on the cluster - that's the expected fail-fast behavior, IMO we are good here.

servolkov (approved):

tests/network/sriov/libsriov.py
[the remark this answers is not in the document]
azhivovk: @coderabbitai Already addressed - this is redundant

libs/vm/vm.py
rnetser: either way it will fail fast, anyhow you can reoslve this

nirdothan (approved):

rnetser (approved):
/approve
/lgtm

openshift-virtualization-qe-bot-5:
Successfully removed PR tag: quay.io/openshift-cnv/openshift-virtualization-tests:pr-5349.

openshift-virtualization-qe-bot-5:

openshift-virtualization-qe-bot-5:
**Cherry-pick conflicts were resolved by AI**

Cherry-picked PR net, sriov: memory hotplug tests into cnv-4.22: https://github.com/RedHatQE/openshift-virtualization-tests/pull/5581
Conflicts were automatically resolved by AI (claude/claude-opus-4-6-1m).

@azhivovk **Manual verification is required** — please review the changes and test before merging.

PR #5739, opened 2026-07-22
feat: Replace code rabbit pull request description review to dedicated workflow

RoniKishner:
Replace code rabbit with a dedicated check.
 - This improves maintainability by separating the description review logic from the rest of the automation, making the code easier to read and evolve. 
 - Provides better control over the review process, helping prevent bugs and making it easier to add review future enhancements, such as validating the PR description against the actual code changes.
   
The current PR description is causing false positive on rare cases, and it also being silently ignored on big Pull Requests, for example:
 - https://github.com/RedHatQE/openshift-virtualization-tests/pull/5612
 - https://github.com/RedHatQE/openshift-virtualization-tests/pull/5737

An example - https://github.com/RoniKishner/openshift-virtualization-tests/pull/44

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -278,11 +278,6 @@ This is a test suite - internal APIs have NO backward compatibility requirements
 - **Mark PR as draft** when there are unresolved blockers, failing CI, or open design questions.
 - **NEVER** merge a PR with known unresolved issues — fix or document them first.
 - **DCO (Signed-off-by) REQUIRED** — all commits must include `Signed-off-by` trailer (enforced by CI).
-- **PR template sections REQUIRED** — the PR description MUST preserve these sections from `.github/pull_request_template.md` (even if left empty):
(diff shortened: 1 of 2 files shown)


rnetser (commented):
Thanks for looking into this but please keep this in CodeRabbit than add a Gemini workflow.

 Template validation under path_instructions is likely why we get false positives / silent skips on big PRs. Better fix: drop that block and move the rules into a pre_merge_checks.custom_checks entry (same pattern as the STP checks).
 We already have custom checks on this repo, so no need for a separate Action + GEMINI_API_KEY.

AGENTS.md
```diff
@@ -278,11 +278,6 @@ This is a test suite - internal APIs have NO backward compatibility requirements
 - **Mark PR as draft** when there are unresolved blockers, failing CI, or open design questions.
 - **NEVER** merge a PR with known unresolved issues — fix or document them first.
 - **DCO (Signed-off-by) REQUIRED** — all commits must include `Signed-off-by` trailer (enforced by CI).
-- **PR template sections REQUIRED** — the PR description MUST preserve these sections from `.github/pull_request_template.md` (even if left empty):
```
rnetser, line 281: should not be dropped ; used for local/dev agent review guidance

RoniKishner: Done

PR #6208, opened 2026-09-03
net, evpn: fix EVPN test infrastructure for OVN-K VTEP BGP advertisement

servolkov:

Fix the external FRR pod nexthop advertisement to be compatible with
the VTEP host IP BGP advertisement introduced in recent OVN-K versions,
and strengthen the overall EVPN test setup.

**Main fix — frr-external nexthop advertisement:**
Recent OVN-K versions introduced per-node VTEP host IP advertisement
via BGP through the ovnk-generated FRRConfiguration. With eBGP
next-hop-self, the external FRR pod was re-advertising these /32 host
routes to peer nodes with itself as the next-hop, installing host routes
in the main routing table that redirected all inter-node traffic
(including etcd raft) through the FRR pod, which dropped it.
Fixed by using `attribute-unchanged next-hop` so the original node IP
is preserved, making the host routes harmless.

FRR reference:
https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighbor-PEER-attribute-unchanged-as-path-next-hop-med

**EVPN setup improvements:**
- **use node-primary-ifaddr for node IP extraction**
Replace parsing of k8s.ovn.org/host-cidrs with k8s.ovn.org/node-primary-ifaddr.
OVN-K sets this annotation to the node's primary interface address per address
family, giving exactly one unambiguous entry per family.

- **ovn_local_gateway_mode yielded without waiting** 
Patching `routingViaHost: True` triggers an OVN-K DaemonSet rollout. 
Following the same pattern used in similar network operator patches 
across the repo (e.g. `cluster_network_resource_ra_enabled`, 
`multi_network_policy_enabled`), add `wait_for_consistent_resource_conditions` 
on the network ClusterOperator before yielding and after teardown. 
This is a safety measure to ensure the cluster is fully stable before and after 
applying the `routingViaHost: True` patch, consistent with how other fixtures 
handle network operator changes.

- **remove redundant as-path exclusion from evpn route-map**
OVN-K now generates `allowas-in origin` on nodes for both IPv4 unicast
and L2VPN EVPN address families, so cluster nodes accept routes where
their own ASN is the origin AS without needing the external FRR pod to
strip it first.
References:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6307
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6481

##### jira-ticket:
NONE

## Bug Fixes

- Improved BGP external-router configuration to preserve IPv4 next-hop information and correctly handle EVPN routes.
- Improved EVPN gateway-mode updates by confirming network components reach consistent states before and after changes.

## Tests

- Improved EVPN and VTEP validation by using each worker’s primary IPv4 network information for more reliable network checks.

### DIFF
--- a/tests/network/bgp/evpn/conftest.py
+++ b/tests/network/bgp/evpn/conftest.py
@@ -6,6 +6,7 @@ from typing import Final
 import ocp_resources.network_operator_openshift_io as openshift_no
 import pytest
 from kubernetes.dynamic import DynamicClient
+from ocp_resources.cluster_operator import ClusterOperator
 from ocp_resources.namespace import Namespace
 from ocp_resources.node import Node
 from ocp_resources.resource import ResourceEditor
@@ -45,6 +46,8 @@ from tests.network.libs.bgp import (
 )
 from tests.network.libs.label_selector import LabelSelector
 from tests.network.libs.vm_factory import udn_vm
+from utilities.constants.hco import DEFAULT_RESOURCE_CONDITIONS
+from utilities.infra import wait_for_consistent_resource_conditions
 
 EVPN_ADVERTISE_LABEL: Final[dict] = {"advertise": "evpn"}
 APP_EVPN_CUDN_LABEL: Final[dict] = {**EVPN_ADVERTISE_LABEL, "app": "cudn-evpn"}
@@ -62,6 +65,7 @@ EVPN_IP_VRF_VNI: Final[int] = 20102
 
 @pytest.fixture(scope="module")
 def ovn_local_gateway_mode(
+    admin_client: DynamicClient,
     network_operator: openshift_no.Network,
 ) -> Generator[None]:
     patch = {
@@ -76,7 +80,19 @@ def ovn_local_gateway_mode(
         }
     }
     with ResourceEditor(patches=patch):
+        wait_for_consistent_resource_conditions(
+            dynamic_client=admin_client,
+            resource_kind=ClusterOperator,
+            resource_name="network",
+            expected_conditions=DEFAULT_RESOURCE_CONDITIONS,
+        )
         yield
+    wait_for_consistent_resource_conditions(
+        dynamic_client=admin_client,
+        resource_kind=ClusterOperator,
+        resource_name="network",
+        expected_conditions=DEFAULT_RESOURCE_CONDITIONS,
+    )
 
 
 @pytest.fixture(scope="module")
--- a/tests/network/bgp/evpn/conftest.py
+++ b/tests/network/bgp/evpn/conftest.py
@@ -29,6 +29,7 @@ from tests.network.bgp.evpn.libevpn import (
     deploy_evpn_l3_endpoint,
     deploy_evpn_l3_vrf,
     evpn_workloads_active_connections,
+    node_primary_ipv4_interface,
     teardown_evpn_bridge,
     teardown_evpn_l2_endpoint,
     teardown_evpn_l3_endpoint,
@@ -88,9 +89,7 @@ def vtep(
     admin_client: DynamicClient,
     workers: list[Node],
 ) -> Generator[VTEP]:
-    host_cidrs = json.loads(workers[0].instance.metadata.annotations["k8s.ovn.org/host-cidrs"])
-    host_cidr = next(cidr for cidr in host_cidrs if "." in cidr)
-    vtep_cidr = str(ipaddress.ip_network(host_cidr, strict=False))
+    vtep_cidr = str(node_primary_ipv4_interface(workers[0]).network)
     with VTEP(
         name="evpn-vtep",
         cidrs=[vtep_cidr],
@@ -217,11 +216,7 @@ def evpn_bridge(
     frr_external_pod: ExternalFrrPodInfo,
     workers: list[Node],
 ) -> Generator[None]:
-    worker_ips = []
-    for worker in workers:
-        host_cidrs = json.loads(worker.instance.metadata.annotations["k8s.ovn.org/host-cidrs"])
-        host_ip = next(cidr.split("/")[0] for cidr in host_cidrs if "." in cidr)
-        worker_ips.append(host_ip)
+    worker_ips = [str(node_primary_ipv4_interface(worker).ip) for worker in workers]
 
     deploy_evpn_bridge(
         pod=frr_external_pod.pod,
--- a/tests/network/bgp/evpn/libevpn.py
+++ b/tests/network/bgp/evpn/libevpn.py
@@ -1,10 +1,12 @@
 import contextlib
 import ipaddress
+import json
 import logging
 import shlex
 import uuid
 from collections.abc import Generator
 from dataclasses import dataclass
+from typing import TYPE_CHECKING
 
 from ocp_resources.pod import Pod
 from pytest import Subtests
@@ -22,6 +24,9 @@ from libs.net.traffic_generator import (
 from libs.net.vmspec import lookup_iface_status, lookup_primary_network
 from libs.vm.vm import BaseVirtualMachine
 from tests.network.libs.bgp import CLUSTER_FRR_ASN, EXTERNAL_FRR_ASN, NET_TOOLS_CONTAINER_NAME
+
+if TYPE_CHECKING:
+    from ocp_resources.node import Node
 
 LOGGER = logging.getLogger(__name__)
 
@@ -468,3 +473,16 @@ def assert_evpn_workloads_connectivity(
         for l3_client, l3_server in l3_connections:
             with subtests.test(f"routed-L3 IPv{ipaddress.ip_address(l3_client.server_ip).version}"):
                 assert is_tcp_connection(server=l3_server, client=l3_client)
+
+
+def node_primary_ipv4_interface(node: Node) -> ipaddress.IPv4Interface:
+    """Return the primary IPv4 interface of a node.
+
+    Args:
+        node: The node to get the primary IPv4 interface of.
+
+    Returns:
+        The primary IPv4 interface of the node as an ipaddress.IPv4Interface object.
+    """
+    primary_ifaddr = json.loads(node.instance.metadata.annotations["k8s.ovn.org/node-primary-ifaddr"])
+    return ipaddress.IPv4Interface(address=primary_ifaddr["ipv4"])
--- a/tests/network/libs/bgp.py
+++ b/tests/network/libs/bgp.py
@@ -214,7 +214,7 @@ def generate_frr_conf(
     lines.extend([f" neighbor {ip} remote-as {CLUSTER_FRR_ASN}" for ip in nodes_ipv4_list])
     lines.append("")
 
-    # IPv4 unicast: advertise external subnet, activate neighbors
+    # IPv4 unicast: advertise external subnet to nodes
     lines.extend([
         " address-family ipv4 unicast",
         f"  network {external_subnet_ipv4}",
@@ -222,12 +222,11 @@ def generate_frr_conf(
     for ip in nodes_ipv4_list:
         lines.extend([
             f"  neighbor {ip} activate",
-            f"  neighbor {ip} next-hop-self",
-            f"  neighbor {ip} route-reflector-client",
+            f"  neighbor {ip} attribute-unchanged next-hop",
         ])
     lines.extend([" exit-address-family", ""])
 
-    # EVPN: activate neighbors with route-map to handle AS-path loop prevention
+    # EVPN: activate neighbors for L2VPN EVPN route exchange
     lines.append(" address-family l2vpn evpn")
     for ip in nodes_ipv4_list:
         lines.extend([
--- a/tests/network/libs/bgp.py
+++ b/tests/network/libs/bgp.py
@@ -194,16 +194,15 @@ def generate_frr_conf(
 
     evpn_route_map = "evpn-to-ocp"
 
-    # Route-map: strips cluster ASN from AS_PATH for eBGP EVPN re-advertisement
+    # Route-map: preserve next-hop for EVPN re-advertisement
     lines = [
         f"route-map {evpn_route_map} permit 10",
-        f" set as-path exclude {CLUSTER_FRR_ASN}",
         " set ip next-hop unchanged",
         "exit",
         "",
     ]
 
-    # BGP router and neighbor definitions
+    # BGP router and neighbor definitions (eBGP: external AS)
     lines.extend([
         f"router bgp {EXTERNAL_FRR_ASN}",
         " no bgp ebgp-requires-policy",


tests/network/bgp/evpn/libevpn.py
```diff
@@ -1,11 +1,13 @@
 import contextlib
 import ipaddress
+import json
 import logging
 import shlex
 import uuid
 from collections.abc import Generator
 from dataclasses import dataclass
 
+from ocp_resources.node import Node
```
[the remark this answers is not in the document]
servolkov: fixed

tests/network/bgp/evpn/conftest.py
```diff
@@ -88,9 +89,7 @@ def vtep(
     admin_client: DynamicClient,
     workers: list[Node],
 ) -> Generator[VTEP]:
-    host_cidrs = json.loads(workers[0].instance.metadata.annotations["k8s.ovn.org/host-cidrs"])
-    host_cidr = next(cidr for cidr in host_cidrs if "." in cidr)
-    vtep_cidr = str(ipaddress.ip_network(host_cidr, strict=False))
+    vtep_cidr = str(ipaddress.ip_network(node_primary_ipv4_cidr(workers[0]), strict=False))
```
qinqon: why do we convert to ip address then back to str ?

tests/network/bgp/evpn/libevpn.py
```diff
@@ -468,3 +473,14 @@ def assert_evpn_workloads_connectivity(
         for l3_client, l3_server in l3_connections:
             with subtests.test(f"routed-L3 IPv{ipaddress.ip_address(l3_client.server_ip).version}"):
                 assert is_tcp_connection(server=l3_server, client=l3_client)
+
+
+def node_primary_ipv4_cidr(node: Node) -> str:
```
qinqon: What about ipv6 or dual stack ? this is going to return ipv6 address with cidr < 32

qinqon: Also should we do something special for ipv6 ?

servolkov:
@coderabbit review

tests/network/libs/bgp.py
```diff
@@ -214,20 +214,19 @@ def generate_frr_conf(
...
     lines.extend([
         " address-family ipv4 unicast",
         f"  network {external_subnet_ipv4}",
     ])
     for ip in nodes_ipv4_list:
         lines.extend([
             f"  neighbor {ip} activate",
-            f"  neighbor {ip} next-hop-self",
-            f"  neighbor {ip} route-reflector-client",
+            f"  neighbor {ip} attribute-unchanged next-hop",
```
qinqon, line 224: as an alternative we can directly just remove those two lines, but this is what ovnk maintainers mentioned.

tests/network/bgp/evpn/libevpn.py
```diff
@@ -1,12 +1,17 @@
 import contextlib
 import ipaddress
+import json
```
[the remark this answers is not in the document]
servolkov: we dont do this way with std libs.

tests/network/libs/bgp.py
qinqon: set as-path exclude 64512 is no longer needed: ovn-kubernetes now generates neighbor <peer> allowas-in origin on the nodes for both unicast and l2vpn evpn, so they accept paths where 64512 is the origin AS. Removing it lets the test exercise the product's own eBGP handling instead of masking it.

- ovn-kubernetes/ovn-kubernetes#6307 — "EVPN: add allowas-in origin for BGP neighbors" (l2vpn evpn AF, merged 2026-04-28)
- ovn-kubernetes/ovn-kubernetes#6481 — "RA controller: add unicast allowas-in origin to raw FRR config" (ipv4/ipv6 unicast AF, merged 2026-06-10) — this is the one that made the exclude fully redundant

tests/network/bgp/evpn/libevpn.py
```diff
@@ -468,3 +473,14 @@ def assert_evpn_workloads_connectivity(
...
                 assert is_tcp_connection(server=l3_server, client=l3_client)
+
+
+def node_primary_ipv4_cidr(node: Node) -> str:
+    """Return the primary IPv4 CIDR of a node, skipping /32 host routes.
+
+    Parses k8s.ovn.org/host-cidrs and returns the first non-/32 IPv4
+    CIDR entry. On clusters where /32 host-route entries appear before
+    the interface CIDR, this ensures the correct address is returned.
+    """
```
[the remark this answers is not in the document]
servolkov: local helper, not necessary, the helper semantics is pretty clean.

qinqon (changes requested):

tests/network/libs/bgp.py
servolkov: yep, I decided to add for clarity.

tests/network/bgp/evpn/libevpn.py
servolkov: our nodes are dual-stack. But we filter IPs via `if "." in cidr`. So, how it can return IPv6?

>Also should we do something special for ipv6 ?

if you mean if have to setup ipv6 tunnel - it was not implemented in the product as far as I remember, we use only ipv4 VTEPs. Anyway, this is not related to the current PR.

tests/network/libs/bgp.py
servolkov: as we discussed offline you did not want to have this fix in this PR. I don't mind to have additional 1 line change, let me know.

tests/network/bgp/evpn/conftest.py
servolkov: VTEP consumes subnet string, and we get network from IP here, like `10.10.10.100/21` -> `10.10.8.0/21` (just example IP)

tests/network/bgp/evpn/libevpn.py
```diff
@@ -468,3 +473,14 @@ def assert_evpn_workloads_connectivity(
...
+    host_cidrs = json.loads(node.instance.metadata.annotations["k8s.ovn.org/host-cidrs"])
+    return next(cidr for cidr in host_cidrs if "." in cidr and ipaddress.ip_network(cidr, strict=False).prefixlen < 32)
```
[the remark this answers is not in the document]
servolkov: 1 -line generation usage, standard lib, nothing criminal

tests/network/bgp/evpn/libevpn.py
```diff
@@ -1,12 +1,17 @@
...
 import logging
 import shlex
 import uuid
 from collections.abc import Generator
 from dataclasses import dataclass
+from typing import TYPE_CHECKING
 
 from ocp_resources.pod import Pod
+
+if TYPE_CHECKING:
```
Anatw: We usually have the if TYPE_CHECKING section after the import section. Is it placed here on purpose?

tests/network/bgp/evpn/libevpn.py
Anatw: What if host-cidrs only contains /32 addresses? Then this raises a bare _StopIteration_, which is hard to debug - can we raise an explicit error with context instead?

tests/network/bgp/evpn/conftest.py
```diff
@@ -88,9 +105,7 @@ def vtep(
...
```
qinqon: this is the worker_cidr, the fact that is the vtep_cidr is something that can change in the future when for example they implement the managed vtep (loopback + new cidr) or we do that at nodes (since is the proper way of using vxlan).

tests/network/bgp/evpn/libevpn.py
```diff
@@ -468,3 +473,14 @@ def assert_evpn_workloads_connectivity(
...
+
+
+def node_primary_ipv4_cidr(node: Node) -> str:
+    """Return the primary IPv4 CIDR of a node, skipping /32 host routes.
+
+    Parses k8s.ovn.org/host-cidrs and returns the first non-/32 IPv4
+    CIDR entry. On clusters where /32 host-route entries appear before
+    the interface CIDR, this ensures the correct address is returned.
+    """
+    host_cidrs = json.loads(node.instance.metadata.annotations["k8s.ovn.org/host-cidrs"])
```
qinqon: I have start the ovnk kind cluster with evpn activated and I get:

```yaml
k8s.ovn.org/host-cidrs: '["172.18.0.5/16","172.19.0.2/16","fc00:f853:ccd:e793::5/64"]'
```

There are two IPv4 addresses while at node status

```
status:
  addresses:
  - address: 172.18.0.5
    type: InternalIP
  - address: fc00:f853:ccd:e793::5
    type: InternalIP
```

We should read this annotation instead it also has sections per family and also you will not need to filter by /32 that btw can be a proper node IP for point to point or cloud setups.
```yaml
k8s.ovn.org/node-primary-ifaddr: '{"ipv4":"172.18.0.5/16","ipv6":"fc00:f853:ccd:e793::5/64"}'
```

So maybe some code like the following it will return directly the ip address type we can use later on 
```python
    node_primary_addr = json.loads(node.instance.metadata.annotations["k8s.ovn.org/node-primary-ifaddr"])
    return ipaddress.ip_interface(node_primary_addr["ipv4"])  
```

Then we can get the ip or cidr like 
```python
iface = node_primary_ipv4_cidr(node)
iface.ip       # 172.18.0.2
iface.network  # 172.18.0.0/16```

File: tests/network/bgp/evpn/libevpn.py (hunk 3 above)
Comment (qinqon) [reply tests/network/bgp/evpn/libevpn.py]: > our nodes are dual-stack. But we filter IPs via if "." in cidr. So, how it can return IPv6?

Well that's fragile, you can have ipv4 over ipv6 with `::ffff:172.18.0.2/96` 

I have propose python code to improve the situation

File: tests/network/libs/bgp.py
Comment (qinqon) [reply tests/network/libs/bgp.py]: Let's add this change so we know if customers need to add it or not (if they need to add we will see some failures).

but implement it with some comments pointing to the ovnk PRs.

**Review by qinqon — CHANGES_REQUESTED** [review]
We can really improve the node ip function

File: tests/network/bgp/evpn/conftest.py (hunk 9 above)
Comment (servolkov) [reply tests/network/bgp/evpn/conftest.py]: what is your concern\suggestion here?

File: tests/network/bgp/evpn/libevpn.py (hunk 8 above)
Comment (servolkov) [reply tests/network/bgp/evpn/libevpn.py]: fixed, however if CI is green - I don't care.

File: tests/network/bgp/evpn/libevpn.py (hunk 3 above)
Comment (servolkov) [reply tests/network/bgp/evpn/libevpn.py]: `::ffff:172.18.0.2/96` look like a very edge case, not sure we ever use this type of IP. However, I really like your suggestion with `node-primary-ifaddr` annotation, fixed the code to use it.

File: tests/network/bgp/evpn/libevpn.py (hunk 10 above)
Comment (servolkov) [reply tests/network/bgp/evpn/libevpn.py]: `node-primary-ifaddr` - looks great, adapted the code to use it.

File: tests/network/bgp/evpn/libevpn.py (hunk 7 above)
Comment (servolkov) [reply tests/network/bgp/evpn/libevpn.py]: the helper has changed, pls take a look.

File: tests/network/libs/bgp.py
Comment (servolkov) [reply tests/network/libs/bgp.py]: fixed with dedicated commit

File: tests/network/libs/bgp.py [hunk 11]
```diff
@@ -214,20 +214,19 @@ def generate_frr_conf(
...
     lines.extend([
         " address-family ipv4 unicast",
         f"  network {external_subnet_ipv4}",
     ])
     for ip in nodes_ipv4_list:
         lines.extend([
             f"  neighbor {ip} activate",
-            f"  neighbor {ip} next-hop-self",
-            f"  neighbor {ip} route-reflector-client",
+            f"  neighbor {ip} nexthop-unchanged",
```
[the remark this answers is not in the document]
openshift-virtualization-qe-bot-3: ⚠️ This thread was automatically unresolved because it was resolved without a substantive response. Please address the review comment and explain how it was resolved before resolving this thread again.

servolkov:
[Change: ](https://github.com/RedHatQE/openshift-virtualization-tests/compare/4f6583a605817abf93c7795291b59363b8237c9e..637a9b99a9be813ee56ede2ecf7aa3d56c6ce7a7) 
1. remove `set as-path exclude` option from frr.conf
2. refactor `node_primary_ipv4_cidr` -> `node_primary_ipv4_interface`

tests/network/libs/bgp.py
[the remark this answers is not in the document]
servolkov: this is a very strange assertion. Based on this logic we should go over every single line and assert the config. Weird. Our main assertion that connectivity works over EVPN/BGP -> config is right.

qinqon (approved):
LGTM

tests/network/bgp/evpn/conftest.py
servolkov: resolving, the PR approved

servolkov:
/verified

```
$ uv run pytest -m bgp --cluster-sanity-skip-nodes-check --cluster-sanity-skip-storage-check --skip-deprecated-api-test --cluster-sanity-skip-check
...
9 tests PASSED
```

openshift-virtualization-qe-bot-3:
@coderabbitai
<details>

  Do NOT submit a blocking review event (REQUEST_CHANGES/APPROVE).
  Post a single inline PR comment on Files Changed (non-blocking COMMENT flow).

  **Analysis Requirements:**

         never imports B directly.

  **Output rules:**

  **Your deliverable:**
  Your inline informational comment will be based on the following requirements:

  **Test Execution Plan**

  Use these formats:

  **Real test commands (MANDATORY when changes affect session/runtime code):**

  When the affected code runs at session/collection time (conftest fixtures, pytest plugins,
  config hooks, session-scoped setup) or modifies runtime behavior that unit tests mock away,
  you MUST include concrete `pytest` commands the PR author must run on a real cluster
  to verify the change works end-to-end. Include:
  - A command for the **error/fix path** (the scenario the PR fixes)
  - A command for the **happy path** (regression: the normal case still works)
  - Use lightweight tests (e.g., `--collect-only` for startup failures,
    a single small test for runtime behavior)
  If the PR only changes test logic (not utilities/libs/conftest), the affected test
  paths themselves serve as the real test commands — no separate section needed.

  Example output for a session-startup fix:
  ```
  **Real tests (cluster required)**
  Error path (the fix):
  `pytest tests/storage/.../test_foo.py --storage-class-matrix=nonexistent-sc --collect-only`
  Expected: ValueError with clear message, not IndexError

  Happy path (regression):
  `pytest tests/storage/.../test_foo.py --storage-class-matrix=<valid-sc> -k test_bar`
  Expected: session starts normally
  ```

  **Guidelines:**

  1. **YOUR ONLY DELIVERABLE**: Post one non-blocking inline comment containing the test execution plan on the first changed line
  3. **FALLBACK ONLY**: If inline comment API calls fail after retrying, post as a regular PR comment

</details>

tests/network/bgp/evpn/conftest.py
```diff
@@ -6,6 +6,7 @@
 import ocp_resources.network_operator_openshift_io as openshift_no
 import pytest
 from kubernetes.dynamic import DynamicClient
+from ocp_resources.cluster_operator import ClusterOperator
```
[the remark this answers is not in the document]
servolkov: done https://github.com/RedHatQE/openshift-virtualization-tests/pull/6208#issuecomment-5540231847

Anatw (approved):
Thank you

vsibirsk (approved):

PR #6276, opened 2026-09-11
[Storage] - Guard golden image clone DV waits with restart-threshold stop status

acinko-rh:
Wires the shared `dv_stop_status_restart_threshold` stop-status guard into the
clone DataVolume waits in `tests/storage/golden_image/`.

Both `wait_for_dv_success()` calls in this module wait on DataVolumes cloned
from the Fedora golden-image DataSource, with no early-exit guard. When a clone
importer restarts excessively, the wait blocks until the full timeout before
failing, wasting CI time and hiding the real failure cause. The `cdi_clone`
tests already use `dv_stop_status_restart_threshold` for this reason; this PR
reuses the same shared helper so these clone waits fail fast on excessive
restarts.

[CNV-94467](https://redhat.atlassian.net/browse/CNV-94467)

No new function was added — the existing shared
`tests/storage/stop_status_utils.py:dv_stop_status_restart_threshold` is reused.
Changes are confined to two clone `wait_for_dv_success()` call sites in
`test_golden_image.py`. No behavior change on the success path; only adds an
early stop when a clone DV exceeds the restart threshold.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved golden-image DV test reliability by waiting for the expected stop-status restart threshold during DV readiness checks.
  * Updated namespace DV creation coverage to account for stop-status transitions.

### DIFF
--- a/tests/storage/golden_image/test_golden_image.py
+++ b/tests/storage/golden_image/test_golden_image.py
@@ -7,6 +7,7 @@ from ocp_resources.datavolume import DataVolume
 from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
 from pytest_testconfig import config as py_config
 
+from tests.storage.stop_status_utils import dv_stop_status_restart_threshold
 from utilities.constants.storage import BIND_IMMEDIATE_ANNOTATION, PVC
 from utilities.constants.timeouts import TIMEOUT_20MIN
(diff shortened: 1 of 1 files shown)


acinko-rh:
`-s -o log_cli=true --storage-class-matrix=hostpath-csi-pvc-block tests/storage/golden_image/test_golden_image.py`

ema-aka-young (approved):
/lgtm

